Issue #2145, opened 2026-05-30
Document memory-lookup protocol in AGENTS.md (skill-first, not list_memories)


rjmurillo:
## Problem

Agents repeatedly reach for `list_memories` to discover Serena memories. With 800+ atomic memory files, a bare `list_memories` call returns ~73KB (~enough to require spilling to a temp file) and burns tokens every time. Worse, `list_memories(topic=...)` filters by namespace directory prefix only (e.g. `git`), not arbitrary keywords, so guessing a topic and missing just leads to a full list anyway.

The repository already solves this, but the guidance was not captured where every agent reads it:

- The `memory` skill (`.claude/skills/memory/scripts/search_memory.py`) is a Serena-first + Forgetful router that returns ranked, token-bounded results (~2K) and points at the right `*-index`/atomic file.
- Index files (`memory-index`, `skills-*-index`) are maintained as keyword routers.
- Atomic filenames are designed as activation vocabulary so `read_memory("<name>")` can be guessed directly (a miss returns a cheap "not found", not a list).

## Fix (this issue)

Document the correct lookup protocol in `AGENTS.md` (loaded every session via `CLAUDE.md` @import), ordered skill-first:

1. Primary: `search_memory.py "<query>"` (memory skill).
2. Deep context: `context-retrieval` agent; `/memory-search` for humans.
3. Raw fallback only: guess `read_memory` -> domain `*-index` -> `memory-index`; never bare `list_memories`.

Plus: on adding a memory, update its `*-index`; do not consolidate atomic memories to cheapen listing (breaks discovery + cross-links). Rationale already in `memory-token-efficiency` and `memory-size-001-decomposition-thresholds`.

Refs #239

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #2145: Document memory-lookup protocol in AGENTS.md (skill-first, n
- **Type**: complexity_triage
- **Priority**: medium
- **Suggested Agent**: agent-explainer
- Issue labeled with low complexity.
Suggested agents: agent-explainer

### Route issue #2145 to agent-explainer
- **Type**: agent_routing
- **Priority**: low
- **Suggested Agent**: agent-explainer
- Issue content suggests routing to: agent-explainer
Complexity: low

PR #2154, opened 2026-05-30
fix(agents): unblock main CI (workspace budget + generated-file staleness)

rjmurillo:
## Summary

Unblocks `Run Python Tests` and `build_all` staleness on `main`, which were red on the default branch and therefore failing CI on every open PR.

Two independent main-branch breakages, both fixed here:

1. **Workspace-budget regression.** PR #2146 added a memory-lookup block to the root workspace context, growing it from 2982 to 4065 bytes and exceeding the 3000-byte per-file workspace budget. Because the budget test runs against the merged tree, this broke `Run Python Tests` for the whole PR queue. Fix: compress the always-on guidance to a single Retrieval-Led Reasoning bullet and relocate the full routing detail into the memory skill, where actionable detail belongs. The root context is back to 2987 bytes; both budget enforcers pass (22 tests).

2. **Generated-file staleness.** The `autofix-pr` to `pr-autofix` command rename (#2142) committed the source and bumped the claude plugin, but never regenerated the Copilot skill mirror, leaving the old mirror tracked and the new mirror absent. This tripped the generated-files staleness check on every branch. Fix: regenerate the mirror (rename) and bump both plugin manifests.

## Changes

- Compress memory-routing guidance in the root workspace context to a one-line pointer.
- Move the routing detail (index-first lookup, index-on-add, naming-as-vocabulary) into the memory skill.
- Reconcile the Copilot skill mirrors and bump the claude and copilot plugin versions.

## Validation

- Workspace-budget validator: PASS
- Budget pytest suites: 22 passed
- Generated-files staleness check: clean

Refs #2145

### DIFF
--- a/.claude/.claude-plugin/plugin.json
+++ b/.claude/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "project-toolkit",
   "description": "Complete project development toolkit: agents, slash commands, lifecycle hooks, and reusable skills for Claude Code workflows",
-  "version": "0.3.3",
+  "version": "0.3.4",
   "author": { "name": "rjmurillo" }
 }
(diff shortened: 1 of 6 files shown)


rjmurillo:
@coderabbitai review

Issue #1923, opened 2026-05-08
Enforce em/en-dash prohibition: pre-commit hook + universal.md rule (Claude + Copilot trees)


rjmurillo:
## Problem

Em-dashes (U+2014) and en-dashes (U+2013) recur in agent-authored prose (markdown, commit messages, PR descriptions, retrospectives, ADRs, rule files). Bot reviewers (Copilot, CodeRabbit) flag every occurrence. The project style at `.agents/steering/documentation.md` L228 already prohibits them, and the user's private global instructions also prohibit them, but the rule is not enforced upstream of the bots.

The same risk applies to the Copilot-side mirror tree at `.github/instructions/`: those files are generated from `.claude/rules/` via `build_all.py` and are read at runtime by Copilot CLI agents. Dashes in the source rule files propagate to the mirrors, and dashes authored directly into a mirror (when an agent treats the mirror as a hand-authored file, see PR #1897 round 6 and `feedback_generator_files_inventory.md` follow-up) survive into the next regen if not caught.

Cost shape: 1+ unresolved review thread per dash, every PR. Each thread is a round-trip (review wave, agent edit, second commit, second wave). PR #1915 round 3 is the most recent instance: 25+ em/en dashes in a single retrospective; one Copilot thread; one fix-and-recommit cycle that the rule could have prevented at write time.

### 1. Pre-commit / pre-push hook (deterministic enforcement, ship first)

Add a hook check that greps staged files for U+2014 (`—`) and U+2013 (`–`) and fails the commit when found. Five-line bash. This is the only placement that mechanically prevents the next instance.

Locations to consider:

- `.githooks/pre-commit` (per-file staged check, fastest feedback to the author).
- `scripts/validation/pre_pr.py` (cross-file final gate before push).

**Scope (must include both source and generated mirrors):**

- `**/*.md` (all markdown: docs, rules, sidecars, retros, ADRs, READMEs).
- `**/*.instructions.md` under `.github/instructions/` (Copilot-side mirrors of the `.claude/rules/` source).
- Commit message bodies (via `commit-msg` hook or pre-push commit-walk).
- Agent-authored prompts under `templates/agents/`, `templates/skills/`, and the generated copies at `src/copilot-cli/`, `src/vs-code-agents/`, `src/claude/`, `.claude/agents/`, `.claude/skills/`, `.github/agents/`.

**Excludes** (mirror `.markdownlint-cli2.yaml`):

- Vendored / third-party content.
[...]

PR #2209, opened 2026-06-02
fix(evals): replace prohibited em-dashes in eval test comments

### LINKED ISSUE #2192: Dash-prohibition violation in tests/evals/test_eval_agent_vs_baseline.py (non-exempt path)
## Summary

`tests/evals/test_eval_agent_vs_baseline.py` on `main` contains prohibited
Unicode dashes (U+2013 / U+2014) in code comments, but the file is not under
the only exempt prefix (`tests/hooks/fixtures/`). The dash-prohibition guards
should have blocked these, so either they slipped in before the guard existed
or there is a coverage gap.

## Evidence

`git show main:tests/evals/test_eval_agent_vs_baseline.py | grep -nP '[\x{2013}\x{2014}]'`:

- L1312: `# One fixture, two assertions on the same response (em dash) recall denom`
- L2535: `# per-record fraction here is well below 10% (em dash) proving the halt`

Exemption is `tests/hooks/fixtures/` only:
- `.githooks/pre-commit:342`
- `scripts/validation/pre_pr.py:324,331,433`

`tests/evals/` is NOT exempt, so these are real violations living on `main`.

## Impact

Bot reviewers (Copilot, CodeRabbit) flag every occurrence, costing a thread per
dash on any PR that touches this file. It also signals the guard is not
catching all paths.

## Suggested fix

Replace the two dashes with hyphens or restructure, and add a regression check
(or confirm the guard scans `tests/evals/`).

rjmurillo:
## Summary

`tests/evals/test_eval_agent_vs_baseline.py` carried two U+2014 (em-dash) characters in code comments. The file lives under `tests/evals/`, which is not the exempt prefix (`tests/hooks/fixtures/`), so the dashes violate `.claude/rules/universal.md` MUST NOT entry 5 (Refs #1923). Bot reviewers flag every occurrence, costing a thread per dash on every PR that touches the file.

This change replaces both em-dashes with commas. No test logic changes: the diff is two comment lines only (2 insertions, 2 deletions).

## Per-file change

`tests/evals/test_eval_agent_vs_baseline.py`:
- Line 1312 (comment in `test_recall_uses_assertion_count_not_fixture_count`): `same response - recall denom` becomes `same response, so the recall denom`. Em-dash replaced with a comma plus a connective so the sentence still reads.
- Line 2535 (comment in `test_one_error_in_one_of_five_fixtures_halts_per_fixture`): `well below 10% - proving the halt` becomes `well below 10%, proving the halt`. Em-dash replaced with a comma.

## Dash-guard coverage gap (investigation, no guard change)

The task asked whether the dash-guard SHOULD cover `tests/evals/` and whether a real coverage gap let these dashes in. There is a gap, but it is NOT a prefix problem, and I did not change the guard.

Root cause: both enforcement points scan markdown only.
- `.githooks/pre-commit` (lines 336 to 352) iterates `STAGED_MD_FILES`, which is restricted to `*.md`. Python files are never scanned by the dash block.
- `scripts/validation/pre_pr.py:validate_dash_prohibition` calls `_branch_markdown_files`, which filters `if p.endswith(".md")` (line 370). Python files are excluded from the branch-wide scan.

The exempt prefix `tests/hooks/fixtures/` in `_VENDORED_PREFIXES` (pre_pr.py line 331) and in the pre-commit `case` (line 342) is correct and does not need adjustment. `tests/evals/` is not exempted; it simply never reaches the dash check because the check filters on file extension `.md`.

Why I did not fix it here: widening the scan from `*.md` to also cover `*.py` is a multi-line behavior change across two files (`.githooks/pre-commit` and `scripts/validation/pre_pr.py`) plus their test suites. The rule text in `universal.md` MUST NOT entry 5 already binds "code comments" and "any authored text", so the intent covers `.py`, but the enforcement does not. That widening is its own work-item with its own tests, not a one-line prefix correction, so it stays out of scope per the task guardrail. Flagging it here for a follow-up issue.

Fixes #2192

## Testing

- `grep -nP '[\x{2013}\x{2014}]' tests/evals/test_eval_agent_vs_baseline.py` -> exit 1 (zero matches) after the edit. Before the edit it returned 2 lines (1312, 2535).
- `uv run pytest tests/evals/test_eval_agent_vs_baseline.py --collect-only -q` -> 145 tests collected, exit 0. The file still parses and collects.
- `git diff` confirms only two comment lines changed; no assertions, fixtures, or production code touched.

### DIFF
--- a/tests/evals/test_eval_agent_vs_baseline.py
+++ b/tests/evals/test_eval_agent_vs_baseline.py
@@ -1309,7 +1309,7 @@ class TestReportAggregatorRecall:
         assert result.error_count == 1
 
     def test_recall_uses_assertion_count_not_fixture_count(self):
-        # One fixture, two assertions on the same response — recall denom
+        # One fixture, two assertions on the same response, so the recall denom
         # MUST be the assertion count (2), not the fixture count (1).
         record = RunRecord(
             fixture_id="F001",
@@ -2532,7 +2532,7 @@ class TestPerFixtureHaltThreshold:
         assert '"fixtures_with_errors": 1' in captured.err
         assert '"executed_fixtures": 5' in captured.err
         # Per-record values are also logged so operators can correlate; the
-        # per-record fraction here is well below 10% — proving the halt
+        # per-record fraction here is well below 10%, proving the halt
         # fired on the per-fixture rule, not the per-record one.
         assert '"error_count": 1' in captured.err
         assert '"executed_records": 30' in captured.err


rjmurillo:
Closing as superseded. origin/main already removed both prohibited dashes in tests/evals/test_eval_agent_vs_baseline.py (a later merge replaced them with a colon and comma); a fresh `grep -nP '[\x{2013}\x{2014}]'` against current main returns zero matches. This PR's two-line change is now redundant. The separate coverage-gap finding (dash guard scans .md only, not .py) is captured on issue #2192 for follow-up.

PR #2212, opened 2026-06-02
fix(ci): bypass invalid cached verdicts in agent-review action

### LINKED ISSUE #2195: agent-review caches malformed/truncated verdicts; reruns replay bad result until SHA changes
## Summary

The `agent-review` composite action caches AI verdicts keyed by commit SHA.
When a review produces a truncated or malformed verdict, that bad result is
cached and replayed on every `gh run rerun` for the same SHA. The only escape
is pushing a new commit (new SHA) or setting `bypass-cache`.

## Evidence

- `.github/actions/agent-review/action.yml` cache key:
  `ai-review-${agent}-${commit-sha}-${hashFiles(prompt-file)}`
- Verdict output is served from cache on hit:
  `value: ${{ steps.cache.outputs.cache-hit == 'true' && steps.cached-results.outputs.verdict || steps.review.outputs.verdict }}`
- A `bypass-cache` input exists but is not auto-triggered on invalid output.

Observed in this run: an identical 181-byte truncated verdict replayed across
three reruns until an empty commit busted the cache.

## Impact

A transient truncation (network blip, token limit) becomes a sticky failing
check that cannot be cleared by rerun, forcing a no-op commit to recover.

## Suggested fix

Do not cache verdicts that fail structural validation (missing `VERDICT:`,
truncated body, infrastructure_failure=true). Only cache well-formed results,
so reruns re-evaluate bad ones.

rjmurillo:
## Summary

The agent-review composite action caches AI verdicts keyed by commit SHA so a rerun on the same commit skips the Copilot API call. The cache replays on every rerun until the SHA changes. A truncated or malformed verdict cached once was therefore served forever for that commit, and the review never re-ran.

This change validates the verdict shape at both the write side and the read side, keeping the existing fast path for valid cached verdicts. Validation logic lives in Python per ADR-006; YAML holds only trivial expressions.

### Failure mode

1. `Populate cache directory` previously only skipped caching on an infrastructure failure. It never checked the verdict shape, so an empty, truncated (for example `PAS`), lowercase, multi-line, or otherwise malformed verdict was written to `verdict.txt`.
2. The cache key is `ai-review-<agent>-<commit-sha>-<hashFiles(prompt-file)>`. On the next rerun for the same commit, `Load cached results` read the bad `verdict.txt` and served it directly.
3. With no shape check on the read side either, the poisoned verdict was replayed on every rerun until the commit SHA changed. The fresh review never ran.

### Fix

- New step `Validate cached verdict` runs on a cache hit and calls `scripts/validate_cached_verdict.py`. It sets `cache-valid=true` only when the cached verdict is well-formed. An invalid entry emits a warning and the action falls through to a fresh review.
- Every cached-vs-fresh selection (action `outputs`, `Save review results`, `Generate step summary`, `Check verdict and fail if needed`) now keys off `steps.cache-validate.outputs.cache-valid` instead of the raw `cache-hit`, so an invalid hit uses the fresh review output.
- The review step runs on a true cache miss OR an invalid-cache bypass.
- `Populate cache directory` validates the fresh verdict before writing and clears any stale restored files first, so a malformed verdict cannot poison the per-commit cache. `Save cache` runs only when `verdict.txt` was actually written.

A valid cached verdict still produces a cache hit and skips the API call, exactly as before.

## Per-file change

- `scripts/ai_review_common/verdict.py`: add `is_valid_cached_verdict(verdict)` and the `_CACHEABLE_VERDICT_TOKENS` set (the known token set minus `UNKNOWN`). A cacheable verdict is a single uppercase outcome token with no surrounding or embedded whitespace. `UNKNOWN` is excluded because it means the axis could not be evaluated, not a real outcome; caching it would lock a non-result into the per-commit cache. The token set lives in one place to avoid the drift documented in the existing `merge_verdicts` comments.
- `scripts/validate_cached_verdict.py`: new thin CLI wrapper the action calls. Reads the verdict from `--file` (cache file) or `--verdict` (literal). Exit codes per ADR-035: 0 valid, 1 invalid or missing, 2 usage. A missing or unreadable cache file is treated as invalid (no crash).
- `scripts/ai_review_common/quality_gate.py`, `scripts/ai_review_common/__init__.py`: re-export `is_valid_cached_verdict` through the package facade.
- `.github/actions/agent-review/action.yml`: add the validation step, route all cached-vs-fresh selection off `cache-valid`, gate the cache write on verdict validity, and clear stale restored files before a skipped save. Third-party actions stay pinned to commit SHAs; no new action references added.
- `tests/test_ai_review.py`: add 33 tests (positive, negative, edge) for `is_valid_cached_verdict` and the CLI, including the truncated-token and multi-line-blob failure shapes from this issue.

## Testing

All commands run from the worktree root.

- `uv run pytest tests/test_ai_review.py -q`: 264 passed (33 new cached-verdict tests plus the existing 231).
- `uvx ruff@0.15.14 check scripts/validate_cached_verdict.py tests/test_ai_review.py`: All checks passed.
- `uvx ruff@0.15.14 format --check scripts/validate_cached_verdict.py`: already formatted.
- `uvx mypy@2.1.0 scripts/validate_cached_verdict.py scripts/ai_review_common/verdict.py`: no issues found.
- `python .claude/skills/security-scan/scripts/scan_vulnerabilities.py scripts/validate_cached_verdict.py`: no vulnerabilities.
- CLI smoke: `--verdict PASS` exits 0; `--verdict PAS` exits 1; `--verdict ""` exits 1.
- `python -c "import yaml; yaml.safe_load(open('.github/actions/agent-review/action.yml'))"`: valid YAML.

Note: the repo does not run `ruff check` or `ruff format --check` as a blocking CI gate (the pytest workflow runs pytest, pip-audit, bandit only). Pre-existing import-order and format drift on `scripts/ai_review_common/__init__.py` and `verdict.py` exists on the default branch and was left untouched to keep the diff minimal; my net-new files are fully ruff-clean and formatted.


rjmurillo:
Closing as superseded. Issue #2195 was already resolved by merged PR #2203, which added `scripts/ai_review_common/cache_guard.py` and wired it into the agent-review action (`main` line 141) to skip caching malformed/empty verdicts. This PR is a competing implementation (`validate_cached_verdict.py` + changes to the same `action.yml`, `quality_gate.py`, `verdict.py`) that would duplicate the cache-validation path and conflict with the merged approach. If the additional standalone validator adds value beyond cache_guard, please rework it against current `main` and reopen.

Issue #2993, opened 2026-07-09
Repo-wide ruff cleanup: 331 pre-existing violations block pushes when touching shared files (12 remain via diff-line gate)


rjmurillo:
## Current decision

The original 627-finding measurement came from a filesystem walk and is stale. The repository explicitly accepted option C from the original issue: keep existing debt behind a monotonic whole-tree count ratchet and reject only new authored debt on changed lines. A zero-error bulk cleanup is no longer the completion criterion.

The 2026-07-30 closing comment was only partly correct. It correctly rejected the 627-file cleanup and preserved the count ratchet. It incorrectly concluded no standalone defect remained. The changed-file Ruff gate still linted whole touched files and blocked pre-existing findings. Branch `fix/portability-scanners-p2` reproduces 12 such findings, all outside changed lines, and changes the gate to a shared diff-line scope.

## Accepted contract

- The whole-tree scanner reads Git-tracked files, not a filesystem walk.
- `scripts/ci/ruff_count_baseline.txt` may only decrease. The current branch records the measured count of 306, down from main's stale 308.
- The per-change Ruff gate blocks a finding when any row in its reported range intersects an added or modified line.
- Pre-existing findings outside changed lines remain visible but do not block unrelated work.
- Git failures, malformed Ruff output, and unresolved diff bases fail closed.

## Revised acceptance criteria

- [x] Whole-tree Ruff count ratchet is wired locally and in CI.
- [x] The scanner scope is Git-tracked files and excludes local ignored state.
- [x] The recorded baseline matches the measured branch count of 306.
- [x] Changed-line Ruff tests cover clean, regression, pre-existing, multiline, rename, quoted-path, and Git-failure cases.
- [ ] Merge the diff-line implementation and baseline reduction to `main` before closing this issue again.

## Historical note

The original zero-error criterion is retained in issue history as the initial proposal, not as current policy. The merged history includes incremental cleanup PRs plus the accepted count-ratchet implementation in #3099 and scope correction in #3333.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #2993: Repo-wide ruff cleanup: 627 pre-existing violations block pu
- **Type**: complexity_triage
- **Priority**: medium
- **Suggested Agent**: agent-implementer
- Issue labeled with low complexity.
Suggested agents: agent-implementer, agent-qa

### Route issue #2993 to agent-implementer, agent-qa
- **Type**: agent_routing
[...]

Issue #2876, opened 2026-07-04
pre-push per-file mypy trips no-any-return on all checks_* wrappers (missing scripts/validation on MYPYPATH)


rjmurillo:
## Problem

The pre-push mypy gate (`.githooks/pre-push`, Phase 7, per-file invocation `mypy "${PY_FILES_UNIQUE[@]}"`) invokes mypy on individual changed files without `scripts/validation` on the module search path. Under the repo's `ignore_missing_imports = true`, a bare intra-package import like `from checks_common import _run_subprocess` resolves to `Any`. Combined with `warn_return_any = true`, every wrapper that ends in `return exit_code == 0` then trips `no-any-return`.

This is latent: the gate only mypy-checks CHANGED files, so it stays green until someone touches a `checks_*` module. The next editor of any `checks_*` file gets 9+ spurious BLOCKING errors on code they did not write, with no obvious cause.

## Reproduction (live on `main`)

```bash
tmp=$(mktemp -d)
git show origin/main:scripts/validation/checks_spec.py > "$tmp/checks_spec.py"
uv run mypy "$tmp/checks_spec.py"
# -> Found 9 errors in 1 file:
#    checks_spec.py:50: error: Returning Any from function declared to return "bool"  [no-any-return]
#    ... (8 more at lines 72, 95, 120, 147, 170, 199, 230, 254)
rm -rf "$tmp"
```

Setting `MYPYPATH=scripts/validation` makes the same command report 0 errors, confirming the root cause is the missing module search path, not a genuine typing defect. `_run_subprocess` in `checks_common.py` is correctly typed `-> tuple[int, str, str]`.

## Affected files

All `checks_*` wrapper modules in `scripts/validation/`:
`checks_spec.py`, `checks_plugin.py`, `checks_tooling.py`, `checks_coverage.py`, `checks_dash.py`.

## Proposed fix

Add `scripts/validation` to the module search path for the pre-push per-file mypy invocation (`MYPYPATH=scripts/validation`), or run those modules as a package. Guard against duplicate-module collisions the same way the existing pre-push logic already does (`.githooks/pre-push` duplicate-module guard).

## Workaround (used in PR #2874)

[...]

PR #3022, opened 2026-07-10
style(tests): ruff cleanup burndown slice (9 test files)

rjmurillo:
Burndown slice for the repo-wide ruff cleanup epic.

## Summary

Cleans 9 test files under \`tests/\` to zero ruff violations via safe autofix (I001 unsorted-imports, F401 unused-import, UP-series annotation modernization). No manual E501 rewrites, no logic changes: import reordering and unused-import removal only.

- 10 insertions, 27 deletions across 9 files.
- All touched test files pass (164 tests in the affected files, verified locally).
- Scoped to non-mirror paths so no plugin-version-bump or mirror regeneration is triggered.

## Why these files

The CI ruff ratchet (\`scripts/ci/ruff_ratchet.py\`) and pre-push per-file ruff gate check only changed files. Cleaning currently-dirty files to zero removes them from the backlog and unblocks future edits to them. This slice targets files that reach 100 percent clean with safe autofix alone.

## Scope note

\`tests/test_detect_scope_explosion.py\` was intentionally excluded. Touching it for a ruff autofix trips the pre-push per-file mypy gate: line 68 deliberately does \`result.file_count = 10\` inside \`pytest.raises(AttributeError)\` to assert the frozen \`ScopeResult\` dataclass rejects mutation, and mypy flags the deliberate assignment as \`[misc] read-only\`. Same two-gate interaction as #2876. Left for a separate mypy-targeted pass to keep this slice mechanical.

Refs #2993

## References

- scripts/ci/ruff_ratchet.py
- tests/test_detect_scope_explosion.py

### DIFF
--- a/tests/hooks/test_invoke_lsp_pre_delegation_guard.py
+++ b/tests/hooks/test_invoke_lsp_pre_delegation_guard.py
@@ -14,12 +14,8 @@ from __future__ import annotations
 import json
 import sys
 from pathlib import Path
-from typing import TYPE_CHECKING
 
 import pytest
-
-if TYPE_CHECKING:
-    pass
 
 REPO_ROOT = Path(__file__).resolve().parents[2]
 HOOK_DIR = str(REPO_ROOT / ".claude" / "hooks" / "PreToolUse")
--- a/tests/hooks/test_invoke_lsp_pre_delegation_guard.py
+++ b/tests/hooks/test_invoke_lsp_pre_delegation_guard.py
@@ -15,12 +15,11 @@ import json
 import sys
 from pathlib import Path
 from typing import TYPE_CHECKING
-from unittest.mock import MagicMock
 
 import pytest
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
+    pass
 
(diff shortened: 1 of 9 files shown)


rjmurillo:
Addressed all 8 review threads in 742f9cdb5.

**test_qa_agent_validator.py (7 threads):** ruff UP037 unquoted the \`Callable[...]\` parameter annotations, but \`Callable\` is imported only under \`TYPE_CHECKING\`. The tests pass on the repo floor (Python >=3.14, where PEP 649 defers annotation evaluation), so there was no runtime break, but the bots correctly flagged the fragile shape. Added \`from __future__ import annotations\` so the deferred-string form is explicit and matches the sibling hook test. Verified: ruff clean, mypy clean, 23 tests pass.

**test_invoke_lsp_pre_delegation_guard.py (1 thread):** ruff emptied the \`if TYPE_CHECKING:\` block to a bare \`pass\`. Removed the now-dead \`from typing import TYPE_CHECKING\` import and the empty block. Verified: ruff clean, 32 tests pass.

Zero behavior change; the fix is import-header hygiene only.

Issue #3034, opened 2026-07-11
chore(memory): single-episode edit drags full global causal-graph regen into the commit (ADR-038 pre-commit auto-fix)


rjmurillo:
## Summary

Editing a single episode file drags a full regeneration of the global causal graph (`.agents/memory/causality/causal-graph.json`, ~1,433 nodes across all sessions) into the current commit. When the graph is stale relative to episodes accumulated on other branches, this bundles a large, unrelated cross-session memory diff into whatever commit happens to touch one episode. There is no clean escape: the `SKIP_AUTOFIX=1` bypass avoids the regen but then ships the episode change without its derived graph, which the generated-artifacts rule (regenerate-and-commit-together) forbids.

## Where

`.githooks/pre-commit`, "Causal Graph Update (Auto-fix)" block (around lines 1439 to 1510):

- Line 1452: triggers when any `.agents/memory/episodes/episode-*.json` is staged.
- Line 1467: runs `update_causal_graph.py --episode-path <entire episodes dir>`, deliberately regenerating from ALL episodes ("to ensure all relationships are captured, not just the new episode").
- Lines 1478 to 1479: if the graph changed at all, `git add` stages it into the commit.

Governed by ADR-038 (Reflexion Memory Schema).

## Observed impact

While clearing a bot-review nit on PR #3030 (dropping one duplicate milestone from one episode), the pre-commit hook regenerated and staged the global causal graph, turning a 4-line episode edit into a 3-file, +515/-66-line commit. 555 of those lines were the global graph catching up drift from unrelated sessions (for example memory from PRs #2969, #2960). That churn is inappropriate on an owner-decision ADR PR, so the commit was reverted and the nit left as benign session-history.

## Why it happens

The graph is regenerated opportunistically (only when an episode is staged), so it drifts whenever episodes land through paths that do not run this hook (other branches, squash merges). The next episode-touching commit "catches up" the accumulated global diff, coupling unrelated memory state to an unrelated change.

## Options to consider

1. Scope the regen to the staged episode's own subgraph (nodes/edges for that episode/session), not the entire directory, so the staged diff stays proportional to the change.
2. Keep the full regen but move it out of per-commit pre-commit and into a scheduled/CI job that maintains the graph on `main`, so feature commits are not the catch-up vehicle.
3. Make the graph a build artifact regenerated deterministically in CI and excluded from feature diffs, with a drift check rather than in-commit staging.
[...]

PR #3058, opened 2026-07-15
fix(memory): retract stale causal-graph contributions on episode edit/delete (#3039)

### LINKED ISSUE #3039: Causal graph: episode edits and deletions leave orphaned nodes/edges/patterns
## Problem

The causal-graph generator (`update_causal_graph.py`) merges each episode's
contributions additively and never removes them. After #3034 made edges and
patterns episode-aware (an `episodes` list makes a same-episode reprocess a
no-op), one residual remains: editing an episode to REMOVE an event, decision,
or chain, or deleting the episode file entirely, leaves the prior nodes, edges,
and patterns in the graph forever.

Two concrete cases:

1. **Edit-shrinks-episode.** An episode is edited so a chain that used to exist
   is gone. On reprocess the episode is already in each affected edge's
   `episodes` list, so the merge is a no-op and the now-obsolete edge stays.
   Worse, an edge whose weight should change is also frozen (the episode already
   contributed).
2. **Episode deletion.** The pre-commit selects staged episodes with
   `--diff-filter=ACMR`, which excludes deletions (D). A deleted episode's nodes,
   edges, and patterns are never removed from the graph.

## Why #3034 does not cover this

#3034 fixed the churn direction (whole-directory regeneration dragging unrelated
drift into a commit) and made reprocessing idempotent. Idempotency is
"reprocessing the same content changes nothing." This issue is the opposite:
reprocessing CHANGED or DELETED content should REMOVE the stale contributions,
which additive merge cannot do.

## Reproduce

1. Commit an episode with a decision and an error/recovery chain; observe the
   edge and pattern appear in `.agents/memory/causality/causal-graph.json`.
2. Edit the episode to drop the chain and re-commit.
3. The edge and pattern remain in the graph, now unsupported by any episode
   content.

## Proposed direction

Give the generator replace-episode-contributions semantics: before applying an
episode, remove every node/edge/pattern occurrence attributable to that
`episode_id` (the `episodes` provenance added in #3034 makes this possible for
new contributions), then re-add from current content. Handle deletion by
passing deleted episode ids to a prune path. This is a larger change than
#3034's scope, hence a separate issue.

## Source

Surfaced by the GPT-5.6 Sol adversarial re-review of the #3034 fix. Confirmed
against `update_causal_graph.py` (additive merge, no removal path) and the
pre-commit `--diff-filter=ACMR` episode selection.

rjmurillo:
### What

`update_causal_graph.py` merged each episode additively and never removed anything. Editing an episode to drop a chain left the stale edge/pattern frozen (the `episodes`-guard made the re-add a no-op), and deleting an episode file left its nodes/edges/patterns in the graph forever (the pre-commit's `--diff-filter=ACMR` excludes deletions).

### Why

The causal graph accreted contributions with no retraction path, so its edges and patterns drifted away from what the episodes actually say. #3034 made reprocessing idempotent but explicitly could not un-merge changed or deleted content; this is that inverse.

## Specification References

| **Issue** | Fixes #3039 | episode edits and deletions leave orphaned nodes/edges/patterns |

## Design

- **Per-episode `contributions` map instead of a running average.** Each edge/pattern stores `{episode_id: value}`; the derived `weight`/`success_rate` is the mean. This makes every op exact and reversible: removing one episode recomputes the mean over the survivors (a running average could not be un-merged).
- **Reconcile, not blind replace.** For each processed episode: snapshot the nodes/edges/patterns it currently supports, re-add current content (a no-op when unchanged, preserving #3034 byte-idempotency and `created` timestamps), then retract only the items it no longer supports. An unchanged reprocess changes nothing; an edited episode loses exactly the dropped chain.
- **Deletion path.** New `--prune-episode-ids`; the pre-commit now detects staged episode deletions (`--diff-filter=D`) and prunes them (episode id == filename stem, verified).
- **Legacy handled.** #3034-era items (episodes list, no per-episode values) seed each contribution with the recorded average; pre-#3034 items (bare aggregate count, no provenance) carry their evidence as anonymous sentinel-keyed contributions that survive any episode-id prune.

## Testing

- `tests/skills/memory/` plus the existing pre-commit scope suite, as in `tests/hooks/test_pre_commit_causal_graph_scope.py`: **258 pass** (14 new across edit-shrink retraction, exact shared-edge recompute, deletion prune, dry-run, legacy-evidence preservation; the two #3034 reprocess-idempotency tests still pass).
- ruff + mypy clean on the script; copilot mirror regenerated; `bash -n` on the pre-commit; plugin bumped 0.6.34 -> 0.6.35.

## Notes for Reviewers

The reconcile approach (second commit) fixes a regression the first commit's unconditional pre-remove introduced against #3034 idempotency; both are folded here as the complete #3039 fix. Filed during an open-issue triage sweep.


rjmurillo:
## Status of the red `Validate PR` check: stale run + merge conflict

Root cause (log-grounded): the only `Validate PR` run on head SHA `79f8cfc` (2026-07-16 05:49Z) failed on one CRITICAL, "File mentioned but not in diff: tests/hooks/test_pre_commit_causal_graph_scope.py". The description was **already corrected** afterward (the path now reads "...as in `tests/hooks/test_pre_commit_causal_graph_scope.py`", which the citation pattern masks). Verified against the CI copy of `pr_description.py`: `extract_mentioned_files()` now returns only `update_causal_graph.py` (in diff), 0 CRITICAL.

So the body needs no further change. Two things keep it red:
1. **Stale run.** GitHub does not re-trigger `Validate PR` for a token-authored description edit, so the 05:49Z failure still shows. A new push (or a human/non-token edit, or a manual workflow re-run) clears it.
2. **Merge conflict.** `mergeable_state=dirty`: this branch conflicts with main and needs a rebase/merge before it can land.

Both need a branch push, which the auto-mode classifier gates for a pre-existing PR. Authorize a push (or rebase locally) and I will resolve the conflict and re-trigger the check.

Issue #3045, opened 2026-07-14
Capability-floor detection: make 'auto resolved below the calibrated tier' observable, not silent


rjmurillo:
## Problem

Spun out of #3041 (acceptance criterion 4: decide whether capability-floor
detection is in scope; decision: out of scope for #3041, tracked here).

#3041 documents the frontier-model operating assumption: this project's
guardrails are written to constrain a capable model down, not to scaffold a
weaker model up. ADR-080 defaults every unit of work to the harness-inherited
model (`auto`). That default is safe only while the harness default is
frontier-tier.

Today, "auto resolved to a model below the calibrated envelope" is a **silent**
condition. There is no signal. Quality degrades in the specific way #3041
describes (constrain-down instructions pointed at a model that needed
scaffold-up), and nobody sees it until output quality drops.

## Ask

Decide whether the harness can assert or detect a minimum model tier, so that
running below the calibrated envelope is a **detectable** condition rather than
a silent quality regression.

Open questions:

- Can the harness expose the resolved model id/tier at runtime (Copilot CLI,
  Claude Code, others) in a portable way?
- If detectable, what is the response: a warning, a session-log field, a hook
  that blocks or annotates, or documentation only?
- What defines "the calibrated tier"? #3042 (tiered eval panel) is the natural
  source of a defensible floor: the lowest tier at which evals still pass.

## Relationship

- #3041 documents the assumption this issue makes observable.
- #3042 (tiered eval sweep) can produce the empirical floor this detection would
  compare against.
- ADR-080 is the decision whose `auto` default this detection would guard.

## Out of scope for now

No implementation is proposed here. This is a scoping/decision issue: agree
whether capability-floor detection is worth building, and if so, at what layer
(harness signal, hook, eval-derived floor, or docs only).

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #3045: Capability-floor detection: make 'auto resolved below the ca
- **Type**: complexity_triage
- **Priority**: medium
- **Suggested Agent**: agent-architect
- Issue labeled with low complexity.
Suggested agents: agent-architect, agent-implementer, agent-qa

### Route issue #3045 to agent-architect, agent-implementer, agent-qa
- **Type**: agent_routing
- **Priority**: low
- **Suggested Agent**: agent-architect
- Issue content suggests routing to: agent-architect, agent-implementer, agent-qa
Complexity: low

rjmurillo:
[...]

Issue #3042, opened 2026-07-12
Eval harness must sweep a tiered model panel (Opus + Sol reference, Sonnet + Terra degradation probes), not validate against Sonnet alone


rjmurillo:
## Problem

The eval harness validates agents/skills against **Sonnet** as the reference model. That is the wrong reference tier given the project's actual operating assumption (see #3041): the project is written for, and should be validated against, a **top-tier thinking frontier model** whose failure mode is over-thinking, not incapability.

Two consequences:

1. **We validate at the wrong ceiling.** If evals pass on Sonnet, that tells us the instructions work for a mid-tier instruction-follower — not that they work (or are correctly calibrated) for the frontier model the project assumes and #2840 defaults to via `auto`.
2. **We cannot observe degradation.** The whole point of #3041's constrain-down-vs-scaffold-up asymmetry is that quality *degrades in a specific, measurable way* as you drop below the calibrated tier. A single-model (Sonnet) eval has no reference frame to detect that. We need to *see* the drop-off, not assume it.

## Proposal — multi-tier reference sweep

Restructure the eval to run each unit of work across a **tiered model panel**, with the frontier tier as the pass/fail reference and the lower tiers as **degradation observation points**:

| Role | Model | Purpose |
|------|-------|---------|
| **Reference / default** | **Opus (4.8)** | The tier the project is calibrated for. Pass/fail is judged here. |
| **Reference / cross-family** | **Sol (GPT-5.6 top tier)** | Confirms the calibration isn't Anthropic-specific; frontier-tier peer. |
| **Degradation probe** | **Sonnet (5)** | Strong instruction-follower one tier down. Observe *where* and *how much* quality drops. |
| **Degradation probe** | **Terra (GPT-5.6 mid tier)** | Cross-family mid-tier degradation point. |

The two frontier models (Opus, Sol) establish the reference band; the two lower tiers (Sonnet, Terra) are run **to observe degradation**, not to gate. The report should surface **effect size per tier per unit**, so we can see:

- Which skills/agents are robust across tiers (candidate for `auto`, cheap to run anywhere).
- Which degrade sharply below frontier (the ones whose constrain-down instructions actively mislead a lesser model — the #3041 failure mode, now measured instead of asserted).

## Why this pairs with #2840 and #3041

[...]

Issue #3041, opened 2026-07-12
Document the frontier-model operating assumption; #2840's pin-stripping makes it load-bearing and silent


rjmurillo:
## Problem

The project's prescriptive instructions (skills, agents, commands, ADRs) are written **assuming the executing model is a top-tier thinking frontier model** — one whose failure mode is over-thinking / over-complication, not incapability. Our guardrails are therefore predominantly **constrain-down** guardrails (rein in an over-capable model), not **scaffold-up** guardrails (expand a weaker model's reach).

This assumption is **nowhere documented**. It is an implicit operating model, and #2840 is about to make it load-bearing and silent.

## Why #2840 makes this urgent

[#2840](https://github.com/rjmurillo/ai-agents/issues/2840) strips `model:` frontmatter pins and defaults every unit of work to the **harness-inherited model (`auto`)**. That is the right call for the reasons #2840 gives — but it silently transfers a critical assumption from the frontmatter to the harness: **whatever the harness hands us is capable enough that our constrain-down instructions are the correct polarity.**

If a consumer runs this project on a lesser/cheaper model (or the harness default drifts down), the polarity inverts:

- Our instructions assume: *model is capable, needs constraining down* → minimal, trust-the-model guardrails.
- A lesser model actually needs: *scaffolding up* → heavy, explicit, decompositional guardrails.

Point the constrain-down instruction set at a model that needs scaffold-up and you don't get graceful degradation — you get a model that was never given the scaffolding it needed, executing instructions tuned to suppress a capability it doesn't have.

## The asymmetry this rests on (and why direction matters)

This is not symmetric. The two guardrail directions have different cost/feasibility profiles:

- **Constrain-down** a frontier model: bounded, cheap. Pruning over-thinking is a small edit; the capability is present, you're just reining it in.
- **Scaffold-up** a lesser model: unbounded, often infeasible. You're trying to manufacture a capability that isn't there.

Empirical grounding (external, not ours):
[...]

Issue #2840, opened 2026-07-03
Remove model pins from skills/agents/commands unless eval evidence justifies them


rjmurillo:
## Problem

Skills, agents, and commands pin a specific model and version in frontmatter, but we have no eval evidence that any pinned model outperforms the harness default for that unit of work. The pins are unmanaged, drift over time, and just broke CI when a pinned model ID was retired (the motivation for #2839).

### Current scope (grounded)

- **71** `.claude/skills/*/SKILL.md` files carry a `model:` frontmatter pin.
- Distinct pinned values today:
  - `claude-sonnet-4-6` x50
  - `claude-opus-4-6` x15
  - `claude-haiku-4-5` x6
  - `claude-opus-4-8` x1
  - `haiku` x1 (bare alias, no version)
- Agents pin too: `.github/agents/*.agent.md` sourced from `templates/agents/*.shared.md` (18 bumped in #2839).
- **1** `.claude/commands/*.md` pins a model.

### Evidence the pins are unmanaged

- **Format drift**: skills use `claude-opus-4-6` (hyphen), agents use `claude-opus-4.6` (dot). Same intended model, two spellings, no single source of truth.
- **Version drift**: one skill sits on `claude-opus-4-8` while 15 others sit on `claude-opus-4-6`. No policy decided the split; it happened by hand-editing.
- **Bare alias**: one skill pins `haiku` with no version at all.
- **CI breakage**: a retired pinned model ID caused the failure #2839 works around. Every pin is a latent CI break waiting for the next retirement.

We do not have proof that any of these pins earns its keep. The pins encode a guess ("opus for hard reasoning, sonnet for instruction-following, haiku for cheap lookups"), not a measured result.

## Proposal

Default every skill, agent, and command to the harness-inherited model (`auto` / no pin). Keep a pin **only** when there is eval evidence that a specific model and version measurably beats the default for that unit of work.

Example of the kind of evidence that WOULD justify a pin: an instruction-following task where `sonnet-5` beats `opus` because opus overthinks the constraints. If we can show that with the eval harness, that skill keeps its `sonnet-5` pin and cites the eval run. Everything without such evidence drops the pin.

## Why this is larger than a find-and-replace

This needs eval-harness support and a new class of check we have not tracked before:

[...]

PR #3061, opened 2026-07-15
feat(eval): tiered model-panel sweep harness (#3042)

rjmurillo:
### What

Adds the tiered model-panel sweep harness from #3042. Instead of validating a unit (agent/skill) against one model, it sweeps each unit across a tiered panel and reports effect size per tier so degradation below frontier is measured, not assumed.

- Two frontier tiers (Opus, Sol) form the pass/fail **reference band**; two lower tiers (Sonnet, Terra) are **degradation probes**, never gating.
- Per (unit, tier) it reuses the existing per-unit harness (see `eval-agent-vs-baseline.py`) via the `--provider`/`--model` transport from #2710, reads `recall_delta`, and classifies each unit **robust-across-tiers** or **degrades-at-\<tier\>** when a probe falls more than the drop threshold below the reference band.

### Why

The harness validated at the wrong ceiling (Sonnet) and had no reference frame to *see* the constrain-down-vs-scaffold-up degradation #3041 describes. This measures it.

## Specification References

| **Issue** | Refs #3042 | Eval harness must sweep a tiered model panel |

## Design

- `_model_panel_core.py` is pure: panel config, effect-size aggregation (reference band = mean `recall_delta` over frontier tiers), degradation classification, JSON/human report. Fully unit-testable, no spend.
- `eval-model-panel.py` orchestrates the sweep behind a single **runner seam** (`Runner`); the default runner shells out to the existing harness, tests inject a fake.
- `--dry-run` validates the panel and prints the plan with **zero spend**.
- The panel is `--panel-config`-overridable: the GPT-5.6 (Sol/Terra) model ids are deployment-specific, so they are data, not hardcoded. `parse_panel` rejects unknown providers, missing fields, and an empty reference band.

## What this unblocks

- The **paid sweep** (units x tiers x n-runs x 2 vendors) is the only owner-gated step: confirm the panel model ids + run budget, then run.
- **#3045** (capability-floor detection) consumes this panel's per-unit floor.
- **#2840** (strip model pins unless eval-justified) consumes the per-pin evidence this produces.

## Testing

- `uv run pytest tests/eval/` -> **293 pass** (24 new): panel parsing, band math, robust/degraded/incomplete classification, both renderers, and the CLI dry-run + exit-code contract with a mocked runner.
- ruff + mypy clean; no security-suppression comments; `--dry-run` verified against a real 2-unit x 4-tier plan.

Filed during an open-issue triage sweep.

### DIFF
--- a/.agents/memory/episodes/episode-2026-07-15-session-2604-build-tiered-model-panel-sweep-harness.json
+++ b/.agents/memory/episodes/episode-2026-07-15-session-2604-build-tiered-model-panel-sweep-harness.json
@@ -1,0 +1,35 @@
+{
+  "id": "episode-2026-07-15-session-2604-build-tiered-model-panel-sweep-harness",
+  "session": "2026-07-15-session-2604-build-tiered-model-panel-sweep-harness",
+  "timestamp": "2026-07-15T00:00:00+00:00",
+  "outcome": "success",
+  "task": "Build tiered model-panel sweep harness (#3042): panel config, per-tier effect-size, degradation classification, dry-run, tests",
+  "decisions": [],
+  "events": [
+    {
+      "id": "e001",
+      "timestamp": "2026-07-15T00:00:00+00:00",
+      "type": "test",
+      "content": "scripts/eval/_model_panel_core.py + eval-model-panel.py + 24 tests. Reuses eval-agent-vs-baseline via #2710 --provider/--model; effect-size=recall_delta, degradation classification vs frontier band; --dry-run zero-spend; panel --panel-config-overridable. uv run pytest tests/eval/ -> 293 passed; ruff+mypy clean; no suppressions. Paid sweep owner-gated; unblocks #3045 #2840.",
+      "caused_by": [],
+      "leads_to": []
+    },
+    {
+      "id": "e002",
+      "timestamp": "2026-07-15T00:00:00+00:00",
+      "type": "commit",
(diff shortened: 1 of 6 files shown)


rjmurillo:
## Ready fix for the red `Validate PR` check

Root cause (log-grounded): `scripts/validation/pr_description.py` raised one CRITICAL, "File mentioned but not in diff: eval-agent-vs-baseline.py". In the body's `### What` section the pre-existing harness is cited as `` `eval-agent-vs-baseline.py` ``, but the "Per" cue is separated from the backtick path by "(unit, tier) it reuses", so `_INLINE_CITATION_PATTERN` does not strip it and it is scored as a change-claim (the file is reused, not changed).

Exact fix (PR-description edit, one line): change

> Per (unit, tier) it reuses `eval-agent-vs-baseline.py` via the ...

to

> Per (unit, tier) it reuses the existing per-unit harness (see `eval-agent-vs-baseline.py`) via the ...

The `(see \`...\`)` span matches the citation pattern and is masked before file extraction. Verified against the CI copy of `pr_description.py`: `extract_mentioned_files()` drops `eval-agent-vs-baseline.py` and keeps the two in-diff files. No code change, no dashes.

I could not apply this: `gh pr edit` on a pre-existing PR is gated by the auto-mode classifier. Apply the one-line edit, or grant a `gh pr edit` Bash permission and I will.

Issue #3064, opened 2026-07-15
pre-push workflow gate has no offline-degradation path when gh act/Docker absent (blocks #2967 ADR-006 burn-down)


rjmurillo:
## Problem

The pre-push workflow local-run gate (`.githooks/pre-push:858`, "Workflow local-run gate (actionlint + gh act dry-run + full run) - BLOCKING") hard-blocks any changed `.github/workflows/*.yml` when `gh act` / actionlint / Docker are unavailable:

> the tooling (actionlint, gh act, Docker) is unavailable; that blocks too

There is no offline-degradation path for the "act/Docker absent" case. This is inconsistent with the same gate's existing handling of the absent-**secrets** case, which degrades gracefully via `record_warn` (`.githooks/pre-push:923`, `:953`) and lets the push through, relying on CI to run the workflow with real secrets.

## Impact

- **Blocks ADR-006 burn-down (#2967).** Every one of the 93 genuine violations in #2967 is a `.github/workflows` or `.github/actions` edit. A contributor or agent without a full Docker + `gh act` environment cannot push any of them, so #2967 is uncompletable for that class of environment.
- **Incentivizes `--no-verify`**, which `AGENTS.md` forbids absolutely. A hard block with no degrade path is exactly the friction that pushes contributors toward bypassing hooks.
- **Asymmetric with the secrets carve-out.** The gate already accepts "cannot validate locally, CI will" for absent secrets. The same reasoning applies when `act`/Docker is absent: CI runs the workflow regardless.

## Proposal

Extend the existing `record_warn` degrade pattern to the "act/Docker/actionlint unavailable" case:

- When `gh act` or Docker is provably unavailable, emit a loud `record_warn` (not a block) naming exactly what was skipped, and allow the push.
- Keep `validate_workflows.py` (the offline schema/structure validator that covers both `.github/workflows/` and `.github/actions/`, `.githooks/pre-push:979`) as the BLOCKING floor, since it runs offline.
- Rely on the PR's own GitHub Actions CI as the real workflow-execution gate, exactly as the absent-secrets path already does. A wrong workflow fails the PR's CI, not `main`.

This preserves the safety intent (no unvalidated workflow reaches `main`) while removing the offline dead end.

## Acceptance criteria

- [ ] Pre-push allows a workflow/action change to push when `act`/Docker/actionlint are unavailable, with a `record_warn` naming the skipped validation.
[...]

Issue #2967, opened 2026-07-08
chore(ci): burn down remaining ADR-006 workflow run blocks


rjmurillo:
Tracks the remaining ADR-006 workflow YAML extraction work after issue #2814 validation and the first ai-review context extraction.

## Current inventory

- Total run blocks scanned under .github/workflows and .github/actions: 297
- Genuine ADR-006 violations remaining: 93
- Borderline small inline logic blocks: 62
- False positives, mostly single commands or non-logic calls: 142

The original 91 count is not a scanner false positive problem. The audit used a high-signal subset: run blocks over 10 code lines with conditionals, loops, parsing, computed variables, or output assembly. Revalidation on this branch found 94 before the first extraction and 93 after extracting the ai-review context builder.

## Remaining genuine blocks

- .github/actions/ai-review/action.yml:440, 150 code lines
- .github/actions/ai-review/action.yml:226, 117 code lines
- .github/workflows/ai-issue-triage.yml:331, 75 code lines
- .github/workflows/pr-maintenance.yml:280, 68 code lines
- .github/workflows/pr-validation.yml:211, 68 code lines
- .github/actions/check-agent-infrastructure/action.yml:43, 66 code lines
- .github/workflows/ai-issue-triage.yml:108, 64 code lines
- .github/workflows/codeql-analysis.yml:239, 58 code lines
- .github/workflows/velocity-accelerator.yml:133, 56 code lines
- .github/actions/agent-review/action.yml:300, 51 code lines
- .github/workflows/agent-drift-detection.yml:181, 49 code lines
- .github/workflows/ai-session-protocol.yml:185, 49 code lines
- .github/actions/agent-review/action.yml:232, 48 code lines
- .github/workflows/artifact-insight-scanner.yml:152, 45 code lines
- .github/workflows/post-pr-retrospective.yml:168, 45 code lines
- .github/workflows/ai-session-protocol.yml:90, 44 code lines
- .github/actions/ai-review/action.yml:680, 43 code lines
- .github/workflows/test-codeql-integration.yml:206, 40 code lines
- .github/actions/setup-code-env/action.yml:178, 38 code lines
- .github/workflows/post-pr-retrospective.yml:83, 38 code lines
- .github/workflows/ai-issue-triage.yml:253, 37 code lines
- .github/actions/workflow-debounce/action.yml:38, 36 code lines
- .github/workflows/ai-spec-validation.yml:125, 34 code lines
[...]

PR #3117, opened 2026-07-16
test(validation): make workflow-gate tests hermetic against SKIP_WORKFLOW_LOCAL_TEST (#3115)

### LINKED ISSUE #3115: bug(ci): test_run_workflow_local_test.py is not hermetic against SKIP_WORKFLOW_LOCAL_TEST, making the workflow-gate bypass unusable during any workflow-touching push
## Problem

The pre-push workflow-local-run gate (`.githooks/pre-push` section 8) delegates to `scripts/validation/run_workflow_local_test.py`, which honors the documented `SKIP_WORKFLOW_LOCAL_TEST=true` bypass for a workflow that genuinely cannot run under `gh act`. But `tests/validation/test_run_workflow_local_test.py` reads that same env var at runtime and is NOT hermetic against it: when a push sets `SKIP_WORKFLOW_LOCAL_TEST=true` to get past the act-unrunnable gate, the var leaks into the pre-push pytest subprocess and breaks ~5 tests in that file (they assert non-bypass behavior).

Result: a catch-22 for any push that edits a workflow which is act-unrunnable AND has no absent-secret to trigger the graceful exit-4 skip (e.g. `pr-validation.yml`, whose only secret is the act-builtin GITHUB_TOKEN):
- Without the bypass: the workflow-local-run gate returns exit 1 (act full-run fails), blocking the push.
- With the bypass: the leaked env var breaks the pytest phase, blocking the push.

## Evidence

- Observed while extracting `pr-validation.yml:211` for #2967 (ADR-006 burn-down). The extracted branch (case-insensitive-compare fix, 26 tests green in isolation) cannot be pushed.
- Confirmed pre-existing and unrelated to the diff: swapping in origin/main's `pr-validation.yml` and running the same gate fails identically (act full-run FAIL, truncating at the first job step). So this is an act limitation of the workflow, not the change.
- Proven leak: `tests/validation/test_run_workflow_local_test.py` passes 92 without the var and fails exactly ~5 with `SKIP_WORKFLOW_LOCAL_TEST=true` set.

## Impact

Workflow-file ADR-006 extractions (#2967) that touch an act-unrunnable, secret-only-GITHUB_TOKEN workflow are unpushable. Composite-action extractions (e.g. #3103) and absent-secret workflows (e.g. #3106 issue-triage, which degrades via the exit-4 secret-absent path) are unaffected.

## Proposed fix

Make the test hermetic: add a `tests/validation/conftest.py` autouse fixture that clears `SKIP_WORKFLOW_LOCAL_TEST` (and any other run_workflow_local_test env vars) via `monkeypatch.delenv(..., raising=False)`, so the documented bypass can be used during a workflow-touching push without breaking the unit tests. Overlaps #3064 (pre-push offline-degradation), which owns the same gate.

## Acceptance criteria

- [ ] `tests/validation/test_run_workflow_local_test.py` passes with `SKIP_WORKFLOW_LOCAL_TEST=true` set in the environment.
- [ ] A workflow-touching push that legitimately needs the bypass (act-unrunnable workflow) is not blocked by a false pytest failure.
- [ ] The branch `feat/adr006-extract-pr-validation-2967` (the blocked #2967 slice) can then push.

rjmurillo:
Fixes #3115. run_workflow_local_test.py reads SKIP_WORKFLOW_LOCAL_TEST plus the remote-container markers (CLAUDECODE, CODESPACES) and CI to decide bypass/degrade. Its tests assert the non-bypassed behavior, so exporting SKIP_WORKFLOW_LOCAL_TEST=true (the gate's own documented bypass for an act-unrunnable workflow) leaked into the pre-push pytest subprocess and broke 5 tests, a catch-22 that made any workflow-touching push to an act-unrunnable, secret-only-GITHUB_TOKEN workflow (e.g. pr-validation.yml) unpushable. Adds tests/validation/conftest.py with an autouse fixture that clears those vars via monkeypatch.delenv(raising=False) so the suite runs from a known-empty environment; tests that want a value set it explicitly. Verified: 92 pass with SKIP_WORKFLOW_LOCAL_TEST=true set (was 5 failing) and 92 pass without it, ruff clean. Unblocks the blocked #2967 workflow-file extraction slice (feat/adr006-extract-pr-validation-2967). Overlaps #3064 (same pre-push gate).

### DIFF
--- a/tests/validation/conftest.py
+++ b/tests/validation/conftest.py
@@ -1,0 +1,32 @@
+"""Hermetic env for the workflow-local-run gate tests (#3115).
+
+`scripts/validation/run_workflow_local_test.py` reads process env vars
+(`SKIP_WORKFLOW_LOCAL_TEST`, the remote-container markers, and `CI`) to decide
+whether to bypass or degrade. `tests/validation/test_run_workflow_local_test.py`
+asserts the non-bypassed behavior, so if a real push exports
+`SKIP_WORKFLOW_LOCAL_TEST=true` to get past an act-unrunnable workflow, that var
+leaks into the pre-push pytest subprocess and breaks these tests. That made the
+gate's own documented bypass unusable during any workflow-touching push.
+
+Clear those vars before every test in this directory so the suite pins the
+script's behavior from a known-empty environment. A test that wants a specific
+value sets it explicitly via monkeypatch.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+_GATE_ENV_VARS = (
+    "SKIP_WORKFLOW_LOCAL_TEST",
+    "CLAUDECODE",
+    "CODESPACES",
+    "CI",
+)
+
+
+@pytest.fixture(autouse=True)
+def _clear_workflow_gate_env(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Remove ambient workflow-gate env vars so tests run hermetically."""
+    for name in _GATE_ENV_VARS:
+        monkeypatch.delenv(name, raising=False)


rjmurillo:
owner: local:pr-autofix
session: copilot-1e6b5500
acquired_at: 2026-07-16T19:47:02Z
expires_at: 2026-07-16T20:02:02Z
base_sha: 3dc2a9f2a64e6c46993c5f6f2452dd40094a248d

rjmurillo:
owner: none
session: copilot-1e6b5500
acquired_at: 2026-07-16T19:32:36Z
expires_at: 2026-07-16T19:32:36Z

PR #3155, opened 2026-07-17
fix(validation): detect uncommitted workflow changes in pre_pr (#3134)

### LINKED ISSUE #3134: pre_pr.py misses staged, unstaged, and untracked workflow changes
## Problem

`validate_workflow_local_run()` only inspects `git diff --name-only <base>...HEAD`. It sees committed branch changes, but it ignores staged, unstaged, and untracked workflow files. Running `pre_pr.py` before committing a workflow edit can therefore report success without executing actionlint or the act dry run.

## Repro

An isolated repository was created with one committed workflow and a sentinel `scripts/validation/run_workflow_local_test.py` that exits 1 if invoked. After appending an unstaged workflow edit:

```text
$ git status --short
 M .github/workflows/repro.yml
$ git diff --name-only
.github/workflows/repro.yml
```

Calling the repository function against that checkout produced:

```text
No changed workflow files; nothing to run locally.
result=True
```

The sentinel runner was not invoked. This reproduces the behavior recorded while preparing PR #3125.

## Evidence

- `scripts/validation/checks_plugin.py:252-255` runs only `git diff --name-only <base>...HEAD`.
- The changed-file list therefore excludes the index, worktree, and untracked files.
- Issue #2571 fixed the branch-upstream comparison, but its committed-diff solution still omits local edits.

## Impact

`pre_pr.py` is the required shift-left gate. A developer can run it after editing a workflow, see a pass, and commit a workflow that never received the advertised local validation. The failure then moves to pre-push or CI.

## Expected

Workflow change detection must include every local file state that could enter the next commit.

## Suggested fix

Build a deduplicated union of:

1. Committed branch changes from `<base>...HEAD`.
2. Staged changes from `git diff --cached --name-only`.
3. Unstaged changes from `git diff --name-only`.
4. Untracked workflows from `git ls-files --others --exclude-standard`.

Filter deleted paths before invoking the runner. Keep path containment checks inside `run_workflow_local_test.py`.

## Acceptance

- A committed workflow change runs the local workflow gate.
- A staged workflow change runs it before commit.
- An unstaged workflow change runs it before staging.
- An untracked workflow runs it.
- A deleted workflow does not pass a missing path to the runner.
- Duplicate paths across committed, staged, and unstaged sets run once.
- A repo with no workflow changes retains the current fast skip.

rjmurillo:
## Summary

`pre_pr`'s workflow local-run gate only diffed the committed range
(`base...HEAD`). A workflow change that was staged, unstaged, or untracked was
reported as "No changed workflow files" and skipped, so `pre-push` later flagged
it. That defeats the shift-left intent: the same edit should be caught at
`pre_pr` time, not at push time.

## Fix

Extract `_collect_changed_workflow_files(repo_root, base_ref)`, which unions:

1. the committed branch range (`base...HEAD`),
2. staged files (`git diff --cached`),
3. unstaged files (`git diff`), and
4. untracked files (`git ls-files --others --exclude-standard`).

It dedupes on first occurrence, then filters to existing
`.github/workflows/*.{yml,yaml}`. Deleted paths drop out via the `is_file`
check, so a removed workflow never reaches the runner.

Fail-open behavior is preserved: the committed-range diff still returns `None`
on failure (the caller warns and skips). The three local-state sources are
additive, so a non-zero exit in any one contributes nothing instead of masking
real changes from the others.

## Tests

Nine regression tests over a real isolated git repo (hermetic identity, no
signing) cover: committed, staged, unstaged, untracked, deleted-dropped,
dedup-across-states, non-workflow-ignored, no-change, and
committed-diff-failure fail-open. The pre-existing wrapper test still passes.

Verified locally: 10 passed, `ruff` clean, `mypy` clean.

Fixes #3134

### DIFF
--- a/scripts/validation/checks_plugin.py
+++ b/scripts/validation/checks_plugin.py
@@ -222,6 +222,53 @@ def _is_linked_worktree(repo_root: Path) -> bool:
     return git_dir_path.resolve() != git_common_dir_path.resolve()
 
 
+def _collect_changed_workflow_files(repo_root: Path, base_ref: str) -> list[str] | None:
+    """Deduplicated workflow files changed across every local git state.
+
+    Unions the committed branch range (``base_ref...HEAD``) with staged,
+    unstaged, and untracked files so ``pre_pr`` validates a workflow edit that
+    has not been committed yet (issue #3134). Deleted paths are dropped by the
+    final ``is_file`` check, which keeps a missing path from reaching the
+    runner (the runner rejects paths that escape or do not exist).
+
+    Returns ``None`` when the committed-range diff cannot run, preserving the
+    fail-open skip the caller documented. The three local-state sources are
+    additive: a non-zero exit in any one contributes nothing rather than
+    masking real changes from the others.
+    """
+    committed_code, committed_out, _ = _run_subprocess(
+        ["git", "-C", str(repo_root), "diff", "--name-only", f"{base_ref}...HEAD"]
+    )
+    if committed_code != 0:
+        return None
+
+    blocks = [committed_out]
+    for extra_args in (
+        ["diff", "--cached", "--name-only"],
+        ["diff", "--name-only"],
+        ["ls-files", "--others", "--exclude-standard"],
+    ):
+        code, out, _ = _run_subprocess(["git", "-C", str(repo_root), *extra_args])
+        if code == 0:
+            blocks.append(out)
+
+    seen: set[str] = set()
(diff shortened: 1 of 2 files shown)


rjmurillo:
Superseded by #3158, the later and more complete fix for issue #3134. Both PRs are green and target the same issue; closing this one to leave a single PR per issue. This is a duplicate created by a redundant automated rerun. Reopen if you prefer this approach over #3158.

Issue #2454, opened 2026-06-05
lsp-read-guard blocks reading files with active merge-conflict markers


rjmurillo:
## Problem

The LSP-first read guard (`.claude/hooks/PreToolUse/invoke_lsp_read_guard.py`, ADR-062) blocks `Read` on a configured code file until LSP navigation warmup is done. During a **merge/rebase conflict resolution**, the file on disk contains conflict markers (`<<<<<<<`, `=======`, `>>>>>>>`) and is not valid syntax, so no LSP can parse or navigate it. The guard still blocks the `Read`, forcing a `sed`/`awk` workaround to inspect the conflict.

## Evidence (pr-autofix session 2026-06-05)

Resolving PR #2412's conflict in `scripts/validation/run_workflow_local_test.py`:

```
PreToolUse:Read hook error: lsp-read-guard: hard-block read 3 (nav=1): .../run_workflow_local_test.py
LSP-FIRST BLOCK (Surgical mode required)
Read 3 requires 2 LSP navigation calls; you have 1.
```

The file had active `<<<<<<< HEAD` / `>>>>>>> origin/main` markers; Serena/LSP cannot navigate a file that does not parse, so the warmup requirement is unsatisfiable for that read.

## Impact

Conflict resolution (a core pr-autofix / merge-resolver workflow) is forced onto raw `sed`/`grep`, which the model-patches rule otherwise discourages. Friction on every multi-conflict merge.

## Proposed fix

Exempt the guard when a merge/rebase is in progress or the target file has conflict markers:
- Skip when `.git/MERGE_HEAD`, `.git/rebase-merge`, or `.git/rebase-apply` exists, or
- Skip when the file contains a `^<<<<<<< ` / `^>>>>>>> ` line.

Fail-open in these cases (the guard already fails open when no LSP is reachable).

## References

- `.claude/hooks/PreToolUse/invoke_lsp_read_guard.py`
- `.claude/rules/lsp-first.md`
- ADR-062

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #2454: lsp-read-guard blocks reading files with active merge-confli
- **Type**: complexity_triage
- **Priority**: medium
- **Suggested Agent**: agent-architect
- Issue labeled with medium complexity.
Suggested agents: agent-architect, agent-devops, agent-implementer

### Route issue #2454 to agent-architect, agent-devops, agent-implementer
- **Type**: agent_routing
[...]

PR #3156, opened 2026-07-17
test(hooks): isolate LSP read-guard tests from outer merge state (#3137)

### LINKED ISSUE #3137: bug(test): LSP read-guard tests false-fail during an in-progress merge
## Summary

The LSP read-guard tests depend on the live checkout merge state. During the normal merge-resolver workflow, `MERGE_HEAD` exists until the merge commit. The production guard correctly bypasses source reads during that state per #2454, but 15 tests assume gating remains active and fail.

This prevents a clean full-suite run before the merge commit, which is exactly when merge-resolver asks for validation.

## Repro

1. Create or enter a worktree with a merge in progress:

```bash
git merge --no-commit origin/main
# Resolve and stage conflicts, but do not commit yet.
```

2. Run the focused suite:

```bash
uv run pytest \
  tests/hooks/test_invoke_lsp_read_guard.py \
  tests/hooks/test_invoke_lsp_read_guard_entrypoints.py -q
```

Observed on current `origin/main` content while resolving PR #3097:

```text
15 failed, 53 passed in 0.95s
lsp-read-guard: non-gated target, allow: .../scripts/sample_module.py
```

Representative failures:

```text
TestIsGatedTarget.test_in_repo_code_file_is_gated
TestEvaluateTiers.test_warmup_tier_blocks
TestMain.test_blocks_on_warmup_via_main
TestModuleAsScript.test_block_via_subprocess
```

Control run from detached current `origin/main` with no merge in progress:

```text
68 passed in 0.81s
```

## Root cause

`is_gated_target()` calls `_merge_in_progress(str(root))` against the real repository root. The tests use the live repository as `REPO_ROOT`, so the test outcome changes with `.git/MERGE_HEAD`, `rebase-merge`, or `rebase-apply` outside the test fixture.

The production behavior is correct. The tests do not isolate the external git-state dependency.

## Impact

- Full-suite validation during merge resolution reports 15 false failures.
- The failed full run took 3 minutes 24 seconds before diagnosis.
- Real regressions can be hidden among environment-shaped failures.
- Merge-resolver cannot meet its pre-commit validation goal without knowing this exception.

## Proposed fix

- Unit tests that exercise tier behavior should monkeypatch `_merge_in_progress` to return `False`.
- Merge-bypass tests should create an isolated temporary git repository and assert the bypass there.
- Entrypoint subprocess tests should point `project_dir` at an isolated repository, not the live checkout.
- Add a regression test that runs the focused files while the outer checkout has `MERGE_HEAD` and proves only the explicit merge-bypass cases observe it.

## Acceptance criteria

- Both test files pass while the containing ai-agents checkout has an in-progress merge.
- Explicit tests still prove merge, rebase, linked-worktree, and conflict-marker bypass behavior.
- No production guard behavior changes.

rjmurillo:
## Summary

The LSP read-guard tests pass the live `REPO_ROOT` as `project_dir`. Production
`is_gated_target` consults `_merge_in_progress` against that directory, so when
the containing ai-agents checkout has `.git/MERGE_HEAD` (the normal
merge-resolver window, and exactly when full-suite validation runs) every gated
target flips to bypass and 15 tests false-fail. The production guard is correct
per #2454; only the tests leak an external git-state dependency.

## Fix

- Add a scoped autouse neutralizer to both read-guard test files. It forces the
  `_merge_in_progress` probe to report no merge for the live `REPO_ROOT` only,
  via `monkeypatch` on the shared gate-state module (late binding, so the
  running `is_gated_target` observes it). Isolated temp repositories keep the
  real detection, so the explicit merge, rebase, linked-worktree, and conflict
  bypass tests still exercise the production signal.
- Point the subprocess block test at an isolated project directory. A child
  interpreter cannot be monkeypatched, so isolation comes from `CLAUDE_PROJECT_DIR`
  plus `cwd`. `native_lsp` availability keys on the `.py` extension alone, so the
  warmup tier still blocks outside the real repo.

## Regression coverage

- Per-project_dir merge scoping in-process (neutralizer holds `REPO_ROOT`,
  isolated repo still detects its own `MERGE_HEAD`).
- A subprocess that bypasses its own isolated mid-merge repo, proving the child
  reads the merge signal from its project dir, not the outer checkout.
- Proof the live-root code target stays gated under the neutralizer.

## Verification

Simulated `.git/MERGE_HEAD` in the checkout, ran both files:

```text
before fix: 15 failed, 53 passed
after fix:  73 passed
```

Marker removed on cleanup. No production guard behavior changed.

Fixes #3137

### DIFF
--- a/tests/hooks/test_invoke_lsp_read_guard.py
+++ b/tests/hooks/test_invoke_lsp_read_guard.py
@@ -29,6 +29,7 @@ HOOK_DIR = REPO_ROOT / ".claude" / "hooks" / "PreToolUse"
 sys.path.insert(0, str(HOOK_DIR))
 
 import invoke_lsp_read_guard as guard  # noqa: E402
+from hook_utilities import lsp_gate_state as _gate_state  # noqa: E402
 
 # A real in-repo file path whose extension is overview-capable (python). The
 # file need not exist; detect_providers keys on extension + config, and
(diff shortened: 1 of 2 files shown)


rjmurillo:
Superseded by #3164, the later and more complete fix for issue #3137. Both PRs are green and target the same issue; closing this one to leave a single PR per issue. This is a duplicate created by a redundant automated rerun. Reopen if you prefer this approach over #3164.

PR #3157, opened 2026-07-17
test(eval): isolate suite dry-run from live git state (Fixes #3138)

### LINKED ISSUE #3138: bug(test): eval-suite dry-run test is not isolated from staged merge changes
## Summary

`TestEvalSuiteAgentsDryRun.test_suite_agents_dry_run_does_not_fail_on_zero_scores` claims that `--base-ref HEAD` guarantees an empty diff regardless of repository state. It does not. During an in-progress merge, staged and working-tree changes remain visible to the suite, so the test classifies live PR files instead of exercising an empty diff.

## Repro

1. Enter a repository with a merge in progress and staged resolutions:

```bash
git merge --no-commit origin/main
# Resolve and stage conflicts, but do not commit yet.
```

2. Run the focused test:

```bash
uv run pytest \
  tests/eval/test_eval_agents_dry_run.py::TestEvalSuiteAgentsDryRun::test_suite_agents_dry_run_does_not_fail_on_zero_scores \
  -q
```

Observed while resolving PR #3097:

```text
1 failed in 0.39s
Suite dry-run with empty diff must exit 0; got 1
WARNING: Failed to parse JSON from eval-agents for SKILL
Category agents: FAIL
```

The in-progress merge brought multiple `SKILL.md` files into the live index. The classifier treated those live changes as agent inputs named `SKILL`.

Controls:

```text
Detached current origin/main with a clean index: 1 passed in 0.58s
Same PR worktree after the merge commit: full suite passed this test
```

## Root cause

The test launches `eval-suite.py` with `cwd=REPO_ROOT` and `--base-ref HEAD`. A diff against `HEAD` includes staged and unstaged changes. It does not mean `HEAD..HEAD` and cannot guarantee an empty diff.

The test therefore depends on external git state that its fixture does not own.

## Impact

- Full-suite validation during merge resolution reports a false eval failure.
- The failure joins 15 LSP guard false failures and makes pre-commit full-suite evidence noisy.
- The error text points at eval JSON parsing, not the live-index dependency, which lengthens diagnosis.

## Proposed fix

- Run this end-to-end case in an isolated temporary git repository with one committed tree and a clean index.
- Alternatively, add an explicit changed-files or diff-file input to `eval-suite.py` and drive the empty-diff case without reading the live checkout.
- Keep a separate test that proves staged files are intentionally included when that behavior is desired.
- Replace the inaccurate comment that `--base-ref HEAD` guarantees an empty diff regardless of repo state.

## Acceptance criteria

- The focused test passes when the outer ai-agents checkout has staged changes or an in-progress merge.
- The empty-diff case remains an end-to-end suite invocation.
- A separate test covers real staged-change classification without depending on the developer checkout.

rjmurillo:
## Summary

`eval-suite.py --scope agents --dry-run --base-ref HEAD` did not guarantee an empty diff. `git diff --name-only HEAD` includes staged and unstaged edits, so during an in-progress merge live agent/skill files were classified as agent inputs and the dry-run false-failed with an all-zero weak-spot FAIL.

## Root cause

`detect_changed_files` runs `git diff` with `cwd=REPO_ROOT` derived from the script location, so a temp working directory cannot isolate it. The test comment claimed `--base-ref HEAD` yields an empty diff regardless of repo state; that is false whenever the index or working tree diverges from HEAD (staged merge resolutions, unstaged edits).

## Fix

- Add an explicit `--changed-files-from <path>` input that reads the changed-file list from a file instead of running git diff. Empty file means no changes.
- Drive the empty-diff acceptance case through an empty file, reproducing the original intent without reading the live checkout.
- Add a classifier test proving explicitly provided agent files are still included (the empty case is empty because the list is empty, not because classification broke).
- Replace the inaccurate comment.

## Verification

- 11 tests pass (was 7; +4 for the explicit-input path and classification).
- ruff and mypy clean on both files.
- Empirical: in an isolated temp repo with a simulated in-progress merge and a staged agent edit, the old `--base-ref HEAD` path sees the agent file (non-empty, would false-fail) while the new empty `--changed-files-from` file sees nothing (exit 0).

Fixes #3138

### DIFF
--- a/scripts/eval/eval-suite.py
+++ b/scripts/eval/eval-suite.py
@@ -84,6 +84,20 @@ def detect_changed_files(base_ref: str) -> list[str]:
     return sorted(set(committed + staged))
 
 
+def read_changed_files(source: str) -> list[str]:
+    """Read a newline-delimited changed-file list from ``source``.
+
+    Explicit alternative to :func:`detect_changed_files` for callers that must
+    not depend on the live checkout's git state (issue #3138). Blank lines are
+    dropped; an empty or whitespace-only file yields an empty list, which the
+    classifier treats as "nothing changed". Paths are returned deduplicated and
+    sorted to match :func:`detect_changed_files`.
+    """
+    text = Path(source).read_text(encoding="utf-8")
+    files = [line.strip() for line in text.splitlines()]
+    return sorted({path for path in files if path})
+
+
 # ---------------------------------------------------------------------------
 # Change classification
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


rjmurillo:
Superseded by #3163, the later and more complete fix for issue #3138. Both PRs are green and target the same issue; closing this one to leave a single PR per issue. This is a duplicate created by a redundant automated rerun. Reopen if you prefer this approach over #3163.

PR #3161, opened 2026-07-17
fix(hook): treat -F/--body-file - as stdin in false-completion gate (#3089)

### LINKED ISSUE #3089: false-completion-gate blocks 'git commit -F -' (stdin message) as an unreadable message file
## Summary

The `false-completion-gate` PreToolUse hook
(`.claude/hooks/PreToolUse/invoke_false_completion_gate.py`) blocks
`git commit -F -`, where the commit message is read from stdin (a standard git
idiom, e.g. a heredoc). The gate parses `-F -` as `--file <path="-">`, cannot
read a file literally named `-`, and treats that unreadable file as an inferred
completion claim (`body_unreadable=True`). That sets `body_inferred=True`, which
disables the fail-open branch, so the commit is blocked even when the real
message contains no completion words and there is no session-log evidence.

## Impact

Any agent or user who commits with `git commit -F -` is blocked regardless of
message content. That is a common way to pass a multi-line message via a heredoc
without a temp file. The block reason ("claimed completion ... no verification
evidence") misdescribes the cause, so the fix is non-obvious. The workaround is
to write the message to a real file under a trusted temp root and use
`-F <file>`.

## Root cause (code trace)

1. `_extract_commit_message_file` (regex near L271) matches `-F -` and returns
   the filename `-`.
2. `_read_commit_message_file("-")` (near L325) resolves `-` to a path, finds no
   such file, and returns `None`.
3. `_is_completion_claim_in_message_file` (near L357) maps `None` content to
   `(has_claim=True, body_unreadable=True)`.
4. In `main` (near L677 and L690), `body_inferred = msg_unreadable` becomes
   `True`.
5. The fail-open branch (near L714) `if not session_logs and not body_inferred:`
   is skipped because `body_inferred` is `True`. A commit with no completion
   words and no verification evidence is then blocked.

## Reproduction (deterministic)

Feed two crafted hook inputs to the gate. The ONLY difference is `-F <file>` vs
`-F -`; both messages have identical content with no completion words.

```bash
S=$(mktemp -d)
printf 'feat(x): add thing\n\nno completion words here\n' > "$S/msg.txt"
printf '{"tool_input":{"command":"git commit -F %s/msg.txt"}}\n' "$S" > "$S/in_file.json"
printf '{"tool_input":{"command":"git commit -F -"}}\n'          > "$S/in_stdin.json"

H=.claude/hooks/PreToolUse/invoke_false_completion_gate.py
python3 "$H" < "$S/in_file.json";  echo "file  exit=$?"   # expect 0 (ALLOW)
python3 "$H" < "$S/in_stdin.json"; echo "stdin exit=$?"   # actual 2 (BLOCK)
```

Observed (rjmurillo/ai-agents, `.agents/` present, one today session log without
Pester evidence):

```
file  exit=0
stdin exit=2   {"decision":"block","reason":"... FALSE COMPLETION GATE ..."}
```

## Expected

`-F -` denotes stdin, not a file path. The gate should treat `-F -` as "no
message file to inspect" and fall through to the normal command-string
completion check. A `git commit -F -` whose message has no completion words
should be allowed.

## Suggested fix

Return `None` from `_extract_commit_message_file` (and the parallel
`_extract_pr_body_file`) when the extracted token is exactly `-`. That routes the
command to the command-string claim check instead of inferring a claim from an
unreadable "file". Add regression tests: (a) `git commit -F -` with a no-claim
message is allowed; (b) the existing `-F <unreadable path>` fail-closed behavior
is unchanged for real (non-`-`) paths.

## Related, lower priority

The command matcher keys on the substring `git commit` anywhere in the command,
so a diagnostic such as `echo '{"...":"git commit -F -"}' | python3 hook.py` is
itself blocked because the quoted argument contains `git commit`. Worth noting:
the matcher does not distinguish a real `git commit` invocation from `git commit`
appearing inside a quoted argument.

## Environment

- Repo: rjmurillo/ai-agents
- Hook: `.claude/hooks/PreToolUse/invoke_false_completion_gate.py`
- Found by the implementer agent while committing PR work in a linked worktree
  using `git commit -F -`.

rjmurillo:
## Summary

Fixes #3089. The false-completion gate blocked `git commit -F -` (message
piped on stdin, e.g. a heredoc) while allowing the identical message from a
real file. The stdin form was parsed as a file literally named `-`, which
read as unreadable, inferred a completion claim, and blocked the commit.

## Root cause

`_extract_commit_message_file` and `_extract_pr_body_file` returned the token
`-` as if it were a path. The reader found no such file and returned None
content. The claim-inference helper maps unreadable content to
`(has_claim=True, body_unreadable=True)`, which set `body_inferred=True` in
`main` and disabled the fail-open branch, so the commit was blocked even
though its real stdin message contained no completion words.

## Fix

When the extracted token is exactly `-`, return None. A `-` denotes stdin,
not a file, so the command now routes to the command-string completion check
(the same path used when no message file is present). This is one guard in
each extractor.

The carve-out is scoped to the exact `-` token. Real paths, including
unreadable ones and filenames that merely start with a dash, are still
extracted and keep the fail-closed contract.

## Tests

New `TestStdinMessageFile` class:

- `-F -`, `--file -`, `--file=-`, `-F '-'` (and the `gh pr create` body-file
  equivalents) extract to None.
- The claim-inference helpers return `(False, False)` for the stdin forms.
- A real filename, including one that starts with a dash, is still extracted.
- A real unreadable path still returns `(True, True)` (fail-closed intact).

Full suite: 74 passed. ruff clean on the hook and the test file.

## Generated artifacts and plugin bump

Regenerated the Copilot hook mirror and bumped both plugin manifests to
0.6.46 in the same change. Parity and version-bump gates are green. The
diff exceeds the atomic five-file guideline only because the generated
mirror and the two manifests ship with the source edit, per the
generated-artifacts rule.

## Out of scope

The hook carries pre-existing mypy findings (missing `dict` type args and a
few `Any` returns) in its stdin-JSON parsing helper and one pre-existing
long line elsewhere. The long line was wrapped because the CI ruff ratchet
runs on the whole changed file; the mypy findings are untouched (CI runs no
mypy gate) to keep the blast radius on the reported bug.


rjmurillo:
## Independent verification (out-of-loop + adversarial)

Verified this security-gate fix outside the in-PR test suite. A standalone driver fed crafted PreToolUse payloads through the real `main()` on this branch's `invoke_false_completion_gate.py`, with a today session log that lacks verification evidence (the incident environment):

| Case | Expected | Got |
|------|----------|-----|
| claim-free `-F -` heredoc | allow (exit 0) | 0 |
| claim `-F -` heredoc | block (exit 2) | 2 |
| `printf 'fixed the bug' \| git commit -F -` | block (exit 2) | 2 |
| bare `git commit -F -` (issue repro) | allow (exit 0) | 0 |

The fix works and there is no bypass on normal claim payloads. CI is green (90/90).

## Adversarial review (Codex / GPT-5.x)

One ceiling surfaced: a hex-escaped `printf '\x66\x69\x78\x65\x64...' | git commit -F -` delivers "fixed the bug" but passes, because the gate inspects the literal command string, not the runtime-expanded bytes.

This is not a regression and not a blocker:

1. Pre-existing, not introduced here. The identical bypass already works on `git commit -m "$(printf '\x66\x69\x78\x65\x64')"` today. Static command-string inspection cannot see runtime expansion on any path.
2. Threat-model-irrelevant. This gate is a self-honesty nudge for the agent, with a documented `SKIP_COMPLETION_GATE=true` first-class bypass. The hex payload is strictly harder to reach than the env-var, so it adds no attack surface.
3. The obvious remediation is invalid. "Fail closed on unverifiable stdin" would re-block the bare `git commit -F -` that this PR fixes, reintroducing the bug.

Recommendation: merge as-is. The command-string-inspection ceiling is worth a one-line note in the docstring if not already there, but it is not a defect this PR should carry.

Issue #3329, opened 2026-07-25
Two CI ratchet scripts shipped with tests but were wired into nothing


rjmurillo:
## Summary

Two CI gate scripts shipped with full test coverage but were never invoked by any workflow, hook, or script. Both were written to satisfy an explicit acceptance criterion on an open issue, and both criteria read as met because the script existed and its tests passed.

- `scripts/ci/ruff_count_ratchet.py`, written for the #2993 regression-guard criterion. Tests in `tests/ci/test_ruff_count_ratchet.py`.
- `scripts/ci/adr006_run_block_scanner.py` in `--max` ratchet mode, written for the #2967 burn-down. Tests in `tests/ci/test_adr006_run_block_scanner.py`.

Verified by grepping every call site across `.github/`, `lefthook.yml`, `.githooks/`, and `scripts/`. Zero hits for either name outside its own test file. Fixed in the PR that closes this, but the class of defect is what matters.

## Why this is worse than an unwritten gate

An unwritten gate is visible in the backlog. A written, tested, unwired gate is invisible: the tests are green, the script is in the tree, and the acceptance criterion looks satisfied. The ADR-006 count could have climbed from 87 back to 150 with nothing failing.

A second symptom fell out of the same gap. Because nothing ran the count ratchet, nobody noticed it invoked `ruff check .` with a single hardcoded exclude, so it walked untracked nested worktrees and reported 767 violations against a real tracked count of 361. A gate that has never run in anger has never been calibrated.

## Proposal

Add a check that fails when a script under `scripts/ci/` has no invocation outside its own tests. Rough shape:

1. Enumerate `scripts/ci/*.py`.
2. For each, grep the repo for its basename, excluding `tests/**` and the file itself.
3. Fail with the list of scripts that have zero call sites.

Escape hatch for genuinely library-only modules: an allowlist file, or a module-level marker such as `CI_ENTRYPOINT = False`. Prefer the marker, since an allowlist drifts.

## Alternative considered

Requiring every new `scripts/ci/` script to arrive in the same PR as its workflow wiring is a review-discipline fix, not a mechanical one, and review discipline is exactly what missed this twice.

## Notes

Related: #2993 (ruff debt), #2967 (ADR-006 burn-down).

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #3329: Two CI ratchet scripts shipped with tests but were wired int
- **Type**: complexity_triage
[...]

PR #3333, opened 2026-07-25
ci: wire two orphaned ratchets and fix the act paths-filter payload gap

### LINKED ISSUE #3331: workflow-local gate blocks on an act-only paths-filter payload gap in normal checkouts
## Summary

The pre-push `workflow-local` gate runs `act` against changed workflows and downgrades known act-only failure signatures to a `[WARN]`. One signature is missing, and it blocks every push that touches a workflow using `dorny/paths-filter`.

Verbatim failure:

```
[Python Tests/Check Changed Paths]   ::error::This action requires 'base' input to be configured or 'repository.default_branch' to be set in the event payload
[Python Tests/Check Changed Paths] exitcode '1': failure
Error: Job 'Check Changed Paths' failed
```

`dorny/paths-filter` resolves its comparison base from the event payload. GitHub always populates `repository.default_branch`, so the branch is unreachable in CI. `act` builds a synthetic payload that omits it.

## Why the existing downgrade missed it

The list already carries `fatal: not a git repository` for the case where the act container cannot reach a `.git`. Which of the two fires depends on the checkout shape:

- Linked worktree: `.git` is a file pointing outside the mount, the action fails early on the missing repo, and the existing pattern downgrades it to `[WARN]`.
- Normal checkout: the container reaches `.git`, the action gets further, and it aborts on the missing payload key instead. No pattern, hard fail.

So the gate passes in a worktree and blocks in a normal clone for the same workflow and the same underlying limitation. Reproduced both ways: `scripts/validation/run_workflow_local_test.py --files .github/workflows/pytest.yml` exits 0 with `[WARN]` from a linked worktree at `origin/main`, and exits nonzero from a normal checkout.

## Fix shipped

Added the payload-gap signature to the downgrade list. Not event-scoped, since the payload omission is identical on every act event, matching how the missing-`.git` pattern is handled. Two tests cover it: one asserting the verbatim act line downgrades on `pull_request`, `workflow_dispatch`, and the unscoped call, and a negative control asserting that unrelated errors merely mentioning `base` keep blocking.

## Alternative considered and rejected

Injecting a synthetic event payload carrying `repository.default_branch` would make `paths-filter` actually run under act. That is more faithful, but it means maintaining a payload fixture that drifts against GitHub's schema, and the two adjacent limitations (`pull_request` context, PR env vars) are already handled by downgrade rather than by simulation. Consistency with the existing approach wins.

## Notes

Related: #2719 (the missing-`.git` downgrade), #3265 (the PR-context downgrades).

rjmurillo:
### What

Wires two CI ratchet scripts that shipped with full test coverage but were invoked by nothing, fixes the scope bug in one of them, and unblocks pushes that touch a workflow using `dorny/paths-filter`.

### Why

Both scripts were written to satisfy an explicit acceptance criterion on an open issue. Both criteria read as met because the script existed and its tests passed. Neither gate had ever run in anger, so the ADR-006 count could have climbed from 87 back to any number with nothing failing, and repo-wide ruff debt had no ceiling at all.

Verified by grepping every call site across the workflows tree, the lefthook config, the git hooks, and the scripts tree. Zero hits for either script name outside its own test file.

## Specification References

| **Issue** | Refs #2993 | Ruff debt: regression guard criterion |
| **Issue** | Refs #2967 | ADR-006 run-block burn-down: no gate |
| **Issue** | Fixes #3331 | workflow-local blocks on an act-only payload gap |
| **Issue** | Refs #3329 | The tested-but-unwired class of defect |

### 1. Ruff count ratchet scope, then wiring

Running the count ratchet for the first time reported 767 violations against a baseline of 408 and declared a regression. The real tracked-file count is 361. It invoked `ruff check .`, which walks untracked nested worktrees and scratch directories that happen to be on disk, and a single hardcoded exclude did not cover them.

Replaced the directory walk with a `git ls-files` scan over tracked `.py` paths. The path list is roughly 70 KB across 1476 files, so invocations are chunked to a 24000-byte argv budget to stay under the Windows 32767-character `CreateProcess` ceiling. Counts now accumulate across batches rather than the last batch overwriting the total.

Baseline lowered from 408 to 361, locking in drain that had already happened but was never recorded.

Wired into the Python Tests workflow beside the existing changed-file ratchet, under the same `python-changed` gate, since ruff debt only moves when Python moves.

### 2. ADR-006 run-block ratchet wiring

Wired at `--max 87`, the verified current count, into the PR Validation workflow. That job is a required check and runs on every PR regardless of touched paths, which matters because workflow YAML can change in a PR that touches no Python and therefore skips the Python Tests workflow entirely.

Both new steps are single-line invocations of a tested script, so neither adds logic to workflow YAML. The scanner still reports 87 after the edit, confirming the steps do not trip their own gate.

### 3. act paths-filter payload gap

The pre-push `workflow-local` gate runs act against changed workflows and downgrades known act-only failure signatures to a warning. One was missing:

```
::error::This action requires 'base' input to be configured or 'repository.default_branch' to be set in the event payload
```

`dorny/paths-filter` resolves its comparison base from the event payload. GitHub always populates `repository.default_branch`, so the branch is unreachable in CI. act builds a synthetic payload that omits it.

The existing list already carries the missing-`.git` signature for the same root limitation. Which of the two fires depends on checkout shape: a linked worktree fails early on the missing repo and gets downgraded, a normal checkout gets further and aborts on the payload key with no matching pattern. So the gate passed in a worktree and blocked in a normal clone for the same workflow.

## Testing

Ruff count ratchet, 8 tests to 14. The shared subprocess fake now answers both the git and ruff legs, which the previous fake did not, so three count tests had been asserting against the wrong process. New cases:

- git failure maps to the external-error exit code
- empty tracked set counts zero rather than erroring
- multi-batch accumulation, which fails if batches overwrite instead of summing
- argv budget is respected
- a single path longer than the budget is still scanned rather than dropped
- a real git repository with one clean tracked file next to an untracked directory full of violations counts zero, which is the direct regression guard for the 767-versus-361 bug

act limitation hint, 2 new tests: the verbatim act line downgrades on `pull_request`, `workflow_dispatch`, and unscoped, plus a negative control asserting that unrelated errors merely mentioning `base` keep blocking.

Verification run:

```
uv run pytest tests/ci/ -q                              94 passed
uv run pytest tests/validation/test_run_workflow_local_test.py -q   98 passed
python3 scripts/ci/ruff_count_ratchet.py                OK (count == baseline 361)
python3 scripts/ci/adr006_run_block_scanner.py --max 87 exit 0
python3 scripts/ci/adr006_run_block_scanner.py --max 86 exit 1
uv run mypy, uv run ruff check                          clean on changed files
```

The full pre-push suite passed, including `build-all-check`, `pre-pr-validation`, `workflow-local-run`, and `python-tests`.

## Notes for Reviewers

The ratchets freeze debt; they do not drain it. #2993 and #2967 stay open for the burn-down itself. Lower `scripts/ci/ruff_count_baseline.txt` with `--update` after a cleanup, and lower `--max` after each ADR-006 extraction.

The ADR-006 step inherits the PR Validation bot-skip gate, so a Renovate PR that touches workflow YAML bypasses it. That matches how every other check in that job behaves and did not seem worth diverging over.

## References

The unwired scanner is `scripts/ci/adr006_run_block_scanner.py` with tests in `tests/ci/test_adr006_run_block_scanner.py`. The diff-scoped ruff gate it sits beside is `scripts/ci/ruff_ratchet.py`. The mypy diff-line ratchet referenced as prior art lives in `scripts/validation/git_hook_policy.py`.


rjmurillo:
Correct, and the comment was misleading enough to be worth fixing. It named one defense as if it were the whole story, and the line immediately below it expands `"$BASE_REF"` in the shell, so a reader who checked would conclude the comment was wrong and stop trusting the block.

Reworded in effe459b38 to name all three layers separately:

1. Keeping the ref in `env` stops the `\${{ }}` expression being substituted into the script source. That is the layer where shell syntax in a ref name would become code.
2. Quoting `"$BASE_REF"` at the one place the shell does expand it keeps it a single argument to git rather than a word-split command line.
3. The ratchet script is handed the fixed literal `FETCH_HEAD`, never the ref name, so nothing attacker-controlled reaches Python argv.

Comment only, no behavior change.

PR #3400, opened 2026-07-26
docs(governance): add worktree identity, path resolution, and skill contract rules

### LINKED ISSUE #3402: Prevent writes to untracked worktrees and stale helper resolution
## Problem

Repository automation can silently target the wrong code tree in two cases:

1. Git can redirect a checkout through a local core.worktree value. Writes then land outside the tree that git reports.
2. Skill resolvers that probe a bare relative .claude path can miss the repository copy outside the repository root and select an older installed plugin helper.

Both failures look successful while using or modifying the wrong files.

## Acceptance criteria

- Add a repository rule that verifies worktree identity before writes.
- Add a rule that anchors repository helper resolution to the absolute git top-level directory.
- Add a rule requiring tests when skill prose defines executable script contracts.
- Generate the Copilot instruction mirrors from canonical rule sources.
- Record the execution evidence in repository memory.
- Pass rule generation, plugin version, and PR description validation.

rjmurillo:
## Summary

- Add rules that verify worktree identity before writes.
- Require repository helpers to resolve from the active checkout.
- Require tests when skill prose defines executable contracts.
- Generate both Copilot instruction mirrors and record the audit evidence.

## Validation

- Rule mirrors regenerate without drift.
- Both project-toolkit manifests use the same version.
- Markdown and repository validation gates pass.

Fixes #3402

### DIFF
--- a/.claude/rules/harness-path-resolution.md
+++ b/.claude/rules/harness-path-resolution.md
@@ -1,0 +1,47 @@
+---
+paths: ".claude/skills/**,src/copilot-cli/skills/**,.claude/commands/**"
+priority: high
+---
+
+# Harness Path Resolution Rule
+
+A skill that shells out to a helper script must resolve that script's directory before invoking it. The resolver walks a candidate ladder: harness environment variables first, then a repo-relative path, then installed-plugin copies under `$HOME`. Every rung after the repo-relative one points at a *different copy of the code* — usually an older one.
+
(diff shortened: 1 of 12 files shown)


rjmurillo:
Closing in favor of #3404.

This PR encoded three findings as markdown rules. That is the defect it was describing: prose asserting something about the codebase, with nothing recompiling the claim when the codebase moves. Markdown rots, and rules-about-rot written in markdown rot at the same rate.

#3404 ships the same three findings as executable guards, wired into `validate-vendor-portability.yml`, with positive/negative/empty-scan controls and a fix for every violation detected.

The evidence that code beats prose here is the detection delta. This PR's hand audit reported 2 unanchored resolver sites in 1 skill. The guard in #3404 finds 8 across 3 skills, including `pr-review` and `ship`, which I never opened. Two further defects in the guard itself were caught by its own tests: a name-based filter that silently missed `resolve_pr_review_config()`, and an ordering check that never fired on an already-anchored resolver. Neither would have been caught by a document.

PR #3423, opened 2026-07-27
fix(merge-driver): bake portable interpreter name, not worktree venv path

### LINKED ISSUE #3418: bug(merge-driver): causal-graph merge driver uses absolute venv path from worktree origin
## Summary

The causal-graph merge driver (`merge.causal-graph.driver` git config) stores the absolute path to the Python interpreter from whichever worktree first ran `install-merge-drivers`. When a different worktree attempts a merge, the driver fails because the path does not exist.

## Evidence

During PR #3409 processing in an isolated worktree at `/tmp/pr-3409-worktree`, the configured driver was:

```
"/tmp/pr-3407-worktree/.venv/bin/python3" scripts/validation/merge_causal_graph.py "%O" "%A" "%B"
```

That path (`/tmp/pr-3407-worktree`) belonged to a different, already-removed worktree. The merge driver silently failed (exit 127: interpreter not found), leaving the causal-graph file in conflict.

## Workaround Applied

Manually ran the merge driver with the system Python:
```bash
python3 scripts/validation/merge_causal_graph.py /tmp/ancestor.json /tmp/ours.json /tmp/theirs.json
```

## Proposed Fix

The merge driver should use a relative or portable Python path. Options:
1. Use `python3` (system) instead of the venv absolute path
2. Use `$(git rev-parse --show-toplevel)/.venv/bin/python3` (per-worktree venv)
3. Register the driver in `install-merge-drivers` using the current worktree's path

## Labels

bug, area-skills, automation

rjmurillo:
## Summary

The causal-graph merge driver stored this process's absolute `sys.executable` in `git config --local`. Linked git worktrees share one `.git/config`, so the first worktree to run the installer registered its own venv path for every worktree. When that worktree or its venv was relocated or deleted, the baked path stopped resolving: the driver exited 127 and git silently fell back to the text merge, reintroducing the causal-graph conflict the driver exists to remove.

## Fix

`_interpreter()` in `scripts/maintenance/install_merge_drivers.py` now returns a bare, PATH-resolved name (`python3`, falling back to `python`, then a `python3` literal) instead of the absolute interpreter path. A bare name cannot dangle across a shared worktree config, and it keeps the sh-safety the POSIX absolute path had (no backslashes to lose on Windows). This matches option 1 in the issue's Proposed Fix, and the reporter's own workaround, which ran the driver with the system `python3`.

The driver needs no project environment (standard library only), so any Python >=3.10 on PATH runs it. The 3.10 floor is inherent to the driver (it imports `typing.TypeAlias` and evaluates a PEP 604 union at import time), not introduced by this change.

## Tests

Two tests in `tests/validation/test_merge_causal_graph.py` encoded the absolute-path assumption and were updated:

- one now asserts the registered token is a bare on-PATH name (not absolute, no venv segment) and still runs the driver end to end;
- the other covers the python3 then python then literal selection order via a mocked PATH.

All 91 tests in the suite pass; `ruff check`, `ruff format --check`, and `mypy` are clean. The pre-commit hook self-healed the local git config to the portable value, verifying the install path end to end.

## References

- Merge behavior itself lives in `scripts/validation/merge_causal_graph.py` (unchanged here; only the interpreter that launches it changed).

Fixes #3418

### DIFF
--- a/scripts/maintenance/install_merge_drivers.py
+++ b/scripts/maintenance/install_merge_drivers.py
@@ -14,6 +14,7 @@ if the driver is already registered with the value we want, nothing is written.
 
 from __future__ import annotations
 
+import shutil
 import subprocess
 import sys
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


rjmurillo:
owner: local:pr-autofix
session: session-3423-1785129241
acquired_at: 2026-07-27T05:14:02Z
expires_at: 2026-07-27T05:29:02Z
base_sha: a3f074a4984161867b7b9bcbe256401b02f6ee78

rjmurillo:
owner: none
session: session-3423-release
acquired_at: 2026-07-27T05:06:17Z
expires_at: 2026-07-27T05:06:17Z

PR #3462, opened 2026-07-27
fix(validation): harden skill byte-ceiling gate against under-count vectors

rjmurillo:
## Summary

Follow-up to #3444, which added the SKILL.md byte ceiling to the skill-size gate. That gate measures the staged index blob, but four paths let an oversized SKILL.md under-count and slip the ceiling. Under-counting is the cardinal failure for a size gate: reporting fewer bytes than will actually be committed.

This PR closes all four vectors and adds one regression test each.

Refs #3421

## The four under-count vectors

1. **Pathspec glob collision.** A skill directory name holding a glob metacharacter (`x?`, `a[x]`) was handed to git ls-files as a pathspec, so it matched a smaller sibling. git ls-files emits the decoy first when it sorts lower, so a first-match read measured the wrong, smaller blob. Fixed with `:(literal)` pathspec magic plus a returned-pathname equality check that binds the read to the exact path under test.

2. **Replace-refs substitution.** git cat-file honors refs/replace/ by default, so a replace ref could swap the indexed blob for a tiny substitute at read time. Fixed with `--no-replace-objects` so the true indexed object is measured.

3. **Newline-bearing path.** A newline is a legal git path byte that the -z discovery stream keeps inside one NUL record. The old anchored match (no DOTALL) let the wildcard stop at the newline, so the path failed to match and dropped from discovery. Fixed with a compiled DOTALL pattern anchored by end-of-string.

4. **Mirror-tree scope.** The lefthook hook stages both the canonical Claude tree and the generated Copilot mirror, but discovery matched only the Claude tree. An oversized staged mirror SKILL.md was silently dropped. Fixed by also matching the Copilot mirror tree via a shared prefix tuple.

## Testing

Four regression tests added, one per vector. Each was verified to fail on pre-fix code and pass with the fix:

- pathspec glob: pre-fix measured 5 bytes (the small decoy) instead of the oversized blob
- replace-refs: pre-fix measured 4 bytes (the substitute) instead of the real blob
- newline path: pre-fix dropped the newline path from discovery
- mirror tree: pre-fix dropped the mirror path from discovery entirely

Full suite: 44 passed. ruff, mypy, complexity (C901 at or below 10), and the live gate are all clean.

## Out of Scope

The byte-ceiling ratchet values and the decomposition of existing oversized skills are handled separately in #3444 and #3460.

## Files Changed

- `scripts/validation/skill_size.py`: the four gate fixes
- `tests/test_validation_skill_size.py`: four regression tests

### DIFF
--- a/scripts/validation/skill_size.py
+++ b/scripts/validation/skill_size.py
@@ -71,8 +71,22 @@ SKILL_BYTE_TARGET: int = 20_480  # 20 KiB, the documented goal (Issue #3421)
 SKILL_BYTE_LIMIT: int = 24_576  # 24 KiB, current ratchet (max body is 24,210 B)
 SKILL_BYTE_WARNING: int = 12_288  # 12 KiB, progressive-disclosure trigger
 
-# Pattern matching staged/changed SKILL.md files
-_SKILL_MD_PATTERN: str = r"^\.claude/skills/.*/SKILL\.md$"
+# Skill trees whose staged/changed SKILL.md bodies the gate measures. Both the
+# canonical Claude tree and the generated Copilot mirror ship skills and are
+# staged by the lefthook ``**/SKILL.md`` glob, so both must be discoverable here
+# or a staged oversized mirror SKILL.md slips the ceiling (under-count, exit 0).
+# ``evals/`` fixtures are excluded by omission: they intentionally reuse skill
+# bodies verbatim and are exempt from skill validation upstream.
+_SKILL_TREE_PREFIXES: tuple[str, ...] = (".claude/skills", "src/copilot-cli/skills")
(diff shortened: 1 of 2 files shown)


rjmurillo:
owner: local:pr-autofix
session: pr-3462-autofix-1785142210
acquired_at: 2026-07-27T08:50:10Z
expires_at: 2026-07-27T09:05:10Z
base_sha: 8783c73e52b7598ceee106c7e59c9cc861642eea

rjmurillo:
owner: none
session: pr-3462-autofix
acquired_at: 2026-07-27T08:47:55Z
expires_at: 2026-07-27T08:47:55Z

PR #3513, opened 2026-07-27
feat(eval): measure whether trigger phrases are things a user has ever said

### LINKED ISSUE #3509: Trigger phrases are validated in a closed loop; 10.4% have ever been said by a real user
## Problem

Every check in `.agents/governance/skill-description-trigger-standard.md` scores a phrase the author wrote against a rule the author wrote. Nothing in the pipeline consults a prompt the author did not write.

That is a closed loop, and a closed loop cannot measure whether anyone would ever type the phrase.

## Evidence this matters

The wiki concept `Skill Triggering Failure Modes` records the collapse. A practitioner's skill-activation classifier scored 93 percent precision and recall on 40 prompts he wrote, then 27 percent precision on 86 prompts mined from his own transcripts. Same classifier, same author. The only variable was who wrote the prompts.

## Measured locally

Mined 640 unique real user prompts from 480 local Claude Code transcripts, matched on word boundaries, excluded slash commands and single words:

| Phrase set | Ever said | Share |
|---|---|---|
| Documented in a `## Triggers` table | 22 of 212 | 10.4% |
| Promoted into a description | 8 of 140 | 5.7% |

The phrases that do occur are conversational and lifecycle-shaped (`complete session`, `create session log`, `start new session`, `your call`, `handle it`, `security review`, `threat model`). The ones that never occur are formal technical constructions.

## What this does and does not claim

It is evidence about **provenance**, not about the router. The router is a model doing semantic matching, so a phrase nobody typed verbatim can still route correctly. What the number establishes is that the phrases were **authored rather than observed**, which is exactly the condition under which the practitioner's classifier collapsed.

## Proposed fix

1. Add an eval that mines the local transcript store and reports the realism figure.
2. Add an Independent-Distribution Validation section to the standard carrying the measured numbers and two rules: prefer an observed phrase to an invented one, and never benchmark trigger phrases on prompts you wrote.
3. Record the finding as a Serena decision memory.

## Two matching rules that are load-bearing

- **Anchor on word boundaries.** A bare substring test counted `analyze` 60 times in this corpus, every one inside ordinary prose. The wiki records the identical failure: every incorrect hard block in the practitioner's hook traced to a pattern matching inside a word.
- **Exclude slash commands and single words.** A slash command is dispatched by name and never consults a description, so counting it measures the dispatcher.

rjmurillo:
## Summary

Adds an eval that answers a question none of the existing trigger-phrase checks can: **has a real user ever actually typed this phrase?**

Every check in the skill-description standard scores a phrase you wrote against a rule you wrote. That is a closed loop. The wiki concept `Skill Triggering Failure Modes` records what happens when the loop opens: a practitioner's skill-activation classifier scored 93 percent precision and recall on 40 prompts he wrote himself, then 27 percent precision on 86 prompts mined from his own transcripts. Same classifier, same author. The only variable was who wrote the prompts.

This eval opens the loop locally by scoring documented trigger phrases against the operator-typed prompts in the local Claude Code transcript store.

## Result

Against 203 unique operator-typed prompts mined from 480 transcripts:

| Phrase set | Ever said | Share |
|---|---|---|
| Documented in a `## Triggers` section | 17 of 422 | 4.0% |
| Promoted into a description | 8 of 226 | 3.5% |

The phrases that do occur are conversational and lifecycle-shaped (`complete session`, `create session log`, `start new session`, `your call`, `handle it`). The ones that never occur are formal technical constructions.

## What this does and does not claim

**Lexical-provenance diagnostic, not a routing measurement.** It never executes the router and reports no precision, recall, or activation rate. A phrase nobody typed verbatim can still route correctly under semantic matching. What the number establishes is narrower: these phrases were authored rather than observed, which is the precondition the practitioner's classifier failed under.

The percentages are deliberately **not** written into the governance standard as fixed numbers. They move with corpus scope, time window, and parser version; one measured run varied from 0 to 12.9 percent across single-project scopes. The standard carries a qualitative rule; the digits live here and in the session log with their scope recorded.

## Adversarial review found the first version was wrong twice

A review on a second model (gpt-5.6-sol) returned five Critical findings. Each premise was reproduced independently before acting. Two were fatal:

1. **The corpus was 68.3 percent machine-authored.** 437 of 640 unique entries carried `isSidechain` or `agentId`: prompts an agent wrote for a subagent. They hold the `user` role but nobody typed them. An eval built to detect a closed loop was running inside one. Corpus is now 203 operator-typed prompts.
2. **The denominator read only Markdown table rows** while the standard permits a trigger "table or list", so it saw 212 of 422 documented phrases. Descriptions were scanned for double quotes only, missing backticked phrases.
3. **The slash-command negative control was fake.** Both fixtures were single words, so the single-word rule caught them; deleting the slash rule still passed 54 of 54. A multiword fixture now fails 4 tests when the rule is cut.
4. The standard asserted "the same collapse is measurable here". Removed.
5. The fixed percentages were too scope-dependent for a governance document. Removed.

The originally published 10.4% and 5.7% were both wrong. Every fix carries a negative-control test.

## Verification

```
tests/eval:            794 passed
this file:              61 passed (54 before the review fixes)
coverage:              _trigger_realism.py 100%, CLI 99% (sys.path and __main__ guards only)
mutation test:         deleting the slash rule now fails 4 tests (0 before)
ruff, mypy:            clean
full pre-push suite:   15,443 passed / 26 skipped
dash bytes:            0 in all changed files
```

The eval never writes prompt text to stdout or JSON, and refuses to fall back to authored prompts (exit 3).

## Changed Files

* `scripts/eval/_trigger_realism.py`: pure scoring helpers, word-boundary matching, measurability rules
* `scripts/eval/eval-trigger-phrase-realism.py`: CLI, transcript loading, phrase collection, reporting
* `tests/eval/test_trigger_phrase_realism.py`: 61 offline tests with negative controls on every rule
* `.agents/governance/skill-description-trigger-standard.md`: Independent-Distribution Validation section, checklist line, references
* `.serena/memories/decision-trigger-phrases-need-real-prompts.md`: decision record
* `.agents/sessions/2026-07-27-session-3509-trigger-phrase-realism.json`: session log

No plugin root is touched, so no manifest bump applies.

Fixes #3509

### DIFF
--- a/.agents/governance/skill-description-trigger-standard.md
+++ b/.agents/governance/skill-description-trigger-standard.md
@@ -236,6 +236,60 @@ The SkillForge validator (`.claude/skills/SkillForge/scripts/validate-skill.py`)
 
 **Note**: Validator checks WORD COUNT (10+ words), not character count. 150-250 chars is recommended for readability but not enforced.
 
+### Independent-Distribution Validation
+
+Every check above scores a phrase you wrote against a rule you wrote. That is a
+closed loop, and a closed loop cannot tell you whether anyone would ever type
+the phrase.
+
+The wiki concept `Skill Triggering Failure Modes` records what happens when the
+loop is opened. A practitioner's skill-activation classifier scored 93 percent
+precision and recall on 40 prompts he wrote himself, then 27 percent precision
+on 86 prompts mined from his own transcripts. Same classifier, same author. The
+only variable was who wrote the prompts.
+
+The same collapse is measurable here. Run:
+
+```bash
+uv run python scripts/eval/eval-trigger-phrase-realism.py
+```
+
+It reads real prompts from the local Claude Code transcript store, matches on
+word boundaries, and reports what fraction of documented phrases a user has
+ever actually typed. Measured against 640 unique real prompts at `4850af73f`:
+
+| Phrase set | Ever said | Share |
+|---|---|---|
+| Documented in a `## Triggers` table | 22 of 212 | 10.4% |
+| Promoted into a description | 8 of 140 | 5.7% |
+
+Read that as evidence about provenance, not about the router. The router is a
+model doing semantic matching, so a phrase nobody typed verbatim can still
+route correctly. What the number establishes is that these phrases were
+**authored rather than observed**, which is exactly the condition under which
+the practitioner's classifier collapsed.
+
+Two rules follow.
+
+1. **Prefer an observed phrase to an invented one.** When a real prompt exists
+   for the behaviour, use its wording. The eval prints every phrase a user has
+   said, so the observed set is the first place to look.
+2. **Never benchmark trigger phrases on prompts you wrote.** A self-authored
+   benchmark measures your own consistency. If no transcript corpus is
+   available, say the phrase set is unvalidated; do not substitute authored
+   prompts. The eval refuses that substitution deliberately and exits 3.
+
+Two matching rules in the eval are load-bearing and worth reusing in any
+successor tool:
+
+- **Anchor on word boundaries.** A bare substring test counted `analyze` 60
+  times in this corpus, every one of them inside ordinary prose. The wiki
+  records the same failure: every incorrect hard block in the practitioner's
+  hook traced to a pattern matching inside a word.
+- **Exclude slash commands and single words.** A slash command is dispatched by
+  name and never consults a description, so counting it measures the
+  dispatcher. A single word collides with prose and inflates the score.
+
 ### Security Requirements
 
 Trigger phrases MUST use character whitelist: `[a-zA-Z0-9 \-:,.'"]`
@@ -251,6 +305,8 @@ Before marking skill complete:
 
 - [ ] Description has action verb
 - [ ] Description includes trigger keywords (how users will search)
+- [ ] Trigger phrases checked against the observed set from
+      `scripts/eval/eval-trigger-phrase-realism.py`, not invented
 - [ ] Description has "Use when" or equivalent
 - [ ] Description mentions outcome
 - [ ] Body has trigger table/list
@@ -297,7 +353,8 @@ Before marking skill complete:
 ## References
 
 - [Skill Description Trigger Review](../analysis/skill-description-trigger-review.md) - 28-skill analysis
-- `Skill Triggering Failure Modes` (wiki: Agent Instruction Patterns) - Over-triggering remediation backing the SKIP clause requirement
+- `Skill Triggering Failure Modes` (wiki: Agent Instruction Patterns) - Over-triggering remediation backing the SKIP clause requirement, and the 93 percent to 27 percent collapse backing Independent-Distribution Validation
+- [`scripts/eval/eval-trigger-phrase-realism.py`](../../scripts/eval/eval-trigger-phrase-realism.py) - Measures documented phrases against real transcript prompts
 - [SkillForge Specification](../../.claude/skills/SkillForge/SKILL.md) - Skill creation framework
 - [Session 372](../sessions/2026-01-03-session-372.json) - Standard creation session
 - [ADR Review Debate Log](../critique/skill-description-trigger-standard-debate-log.md) - P0 issues addressed
--- a/.agents/governance/skill-description-trigger-standard.md
+++ b/.agents/governance/skill-description-trigger-standard.md
@@ -248,26 +248,30 @@ precision and recall on 40 prompts he wrote himself, then 27 percent precision
 on 86 prompts mined from his own transcripts. Same classifier, same author. The
 only variable was who wrote the prompts.
 
-The same collapse is measurable here. Run:
+The same *provenance* condition is measurable here. Run:
 
 ```bash
 uv run python scripts/eval/eval-trigger-phrase-realism.py
 ```
 
-It reads real prompts from the local Claude Code transcript store, matches on
-word boundaries, and reports what fraction of documented phrases a user has
-ever actually typed. Measured against 640 unique real prompts at `4850af73f`:
+It reads operator-typed prompts from the local Claude Code transcript store,
+matches on word boundaries, and reports what fraction of documented phrases a
+user has ever actually typed. On this repository the answer is a low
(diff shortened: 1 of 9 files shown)


rjmurillo:
## Blocking: the corpus is wrong again, and the published number does not hold

A second adversarial review (gpt-5.6-terra, after the first by gpt-5.6-sol) found the provenance filter is still incomplete. I verified it independently and it is worse than reported. Moving to draft rather than leave a number standing that I can no longer defend.

### What the store actually contains

The Claude Code transcript store carries an explicit ground-truth field, `promptSource`, that the eval never read. Measuring the entries this eval accepts today:

```
accepted after the current sidechain/agentId guard : 3544
  isMeta=True                                      :  213
  promptSource='system'                            :   95
  promptSource='typed'                             :   26   <- the actual human prompts
  promptSource='sdk'                               :    2
  promptSource='suggestion_accepted'               :    1
  isCompactSummary / isVisibleInTranscriptOnly     :    5
```

Excluding meta and sidechain leaves 199 texts, 163 unique. Of those, **154 begin with `<`** (synthetic `<command-name>` and `<local-command-stdout>` envelopes) and **9 are prose**, mostly compaction boilerplate ("This session is being continued from a previous conversation") and `[Request interrupted by user for tool use]`.

The genuine operator corpus in this store is roughly **26 unique prompts**, not the 203 this PR claims and not the 640 the first version claimed. Rescored on that corpus the result is **0 of 422 and 0 of 226**. At n=26 a 0 percent reading and a 4 percent reading are indistinguishable, so the percentage is not a measurement.

### Why the reviewer's proposed fix is also wrong

The review suggests accepting only entries with explicit typed provenance. `promptSource` is present on just 120 of 3,331 non-meta user entries, and both the labelled and unlabelled sets span the same dates (2026-06-23 to 2026-07-26), so this is sparsity, not a schema boundary. Filtering on it alone discards real prompts.

### The corpus exists, in the other store

Copilot CLI keeps operator turns in its own session store, where the human turn is the human turn by construction. Scoped to this repository and excluding synthetic-tag turns:

```
ai-agents : 439 unique operator prompts  (206 of them 4 to 400 chars)
all repos : 552 unique
```

That is 17x the usable Claude corpus, over three months and 1,711 sessions.

### Plan

Rework rather than withdraw. The scoring helpers, the phrase parser, and the 61 tests are sound and the reviewer confirmed no prompt text reaches any output path. What is wrong is corpus selection.

1. Read `promptSource` when present; reject meta, compact-summary, transcript-only, system, sdk, and tool-originated entries; reject synthetic `<tag>` envelopes and compaction boilerplate. Fixtures for every rejected shape.
2. Add the Copilot session store as a second source so the eval measures where operator prompts actually live.
3. Add a minimum-corpus guard: below the threshold the eval refuses to report a percentage and exits non-zero rather than publish a number it cannot support.
4. Reword the conclusion from "authored rather than observed" to "unobserved in this corpus", which is all a lexical scan can establish.
5. Fix the three remaining Required findings: quoted first-column table cells, non-mapping YAML frontmatter, and unclassified `OSError` on transcript read and output write.

The count of tests that fail when the slash-command rule is deleted is 3, not the 4 stated in the PR body. Corrected here.

This is the third corpus defect in this PR. All three were found by review on a model other than the author's, and none by the author.

rjmurillo:
## Withdrawing this PR

Two adversarial reviewers on different model families (`gpt-5.6-terra`, `gemini-3.1-pro-preview`) independently returned WITHDRAW. I reproduced the decisive finding myself before acting. The measurement is wrong, for the fourth time, and the honest move is to stop publishing numbers from it.

### The corpus is 86% machine text by volume

The premise of the rework was that `turns.user_message` in `~/.copilot/session-store.db` is the human turn "by construction". It is not. Verified directly:

```
accepted unique copilot texts: 264
injected-prefix counts: {'PreToolUse-hook': 88, 'Autopilot-objective': 12}
injected=100 chars=304,939 of total_chars=353,725
clean remaining: 164
```

PreToolUse hooks write `Additional context from ...` into that column, and autopilot writes `Autopilot objective: ...`. 100 of 264 accepted texts, and **304,939 of 353,725 characters (86%)**, are machine-authored. The column has no author field. There is no provenance system here, and a prefix blacklist is not one.

After removing only the verified injections the operator corpus falls to roughly 186 unique prompts, below this PR's own `MINIMUM_CORPUS = 200` guard. The eval would refuse to report its own headline.

### The control was a volume artifact, not a control

Independently computed by the second reviewer, volume-matching the control down to the operator's character count:

| | reported | volume-matched |
|---|---|---|
| control phrases observed | 28 of 428 | mean 7.6, 95% CI [1.0, 13.0] |

The control corpus carried 2,614,603 characters against the operator's 358,399, a 7.3x exposure difference. More text means more chances to match. The claim "same phrases, same matcher, one variable" was false: the two arms differ in source (operator 92% Copilot, control 100% Claude), volume (7.3x), register, and mean prompt length (4,272 vs 1,253 chars).

Rolled up to independent units the phrases are not 428 trials but 93 skills, and the gap shrinks from 9.3x to 4.3x before any volume correction.

### Other findings, all confirmed

- The `MINIMUM_CORPUS` guard checks only the operator arm, so a zero-control report renders `0 of 428 phrases appear (0.0%)` and publishes it.
- The `sqlite3.Error` handler interpolates the raw exception, and a malformed-UTF-8 cell puts prompt text into that exception. The privacy tests only cover the success path.
- The rejection-rule tests parameterize over the production constants, so deleting a rule deletes its test case and the suite still passes. Tests that cannot fail, again.

### What I am keeping

Nothing published. The three phrases are not defensible either: they came from the same contaminated corpus.

The durable finding is the corpus fact itself, and it ships separately as a Serena memory so the next attempt does not repeat this. Issue #3509 stays open; it is a real question with no valid instrument yet.

### The pattern

Four attempts, one shape each time: I trusted a corpus premise without verifying provenance, built tooling on it, and got a number. The fix is not a better matcher. It is an authoritative typed-vs-machine label at the event level, symmetric source and time windows, equal exposure fixed in advance, and eligibility guards on both arms. Until that exists there is no instrument here.

PR #3573, opened 2026-07-27
fix(eval): reduce repeated rule judge samples

### LINKED ISSUE #3445: eval: rule adapter is single-shot against an LLM judge with no noise reduction
The optimizer gate refuses any pass-to-fail transition on the held-out group (ADR-087 requirement 4). That rule is only as good as the scorer under it: a held-out task that flips for reasons unrelated to the edit rejects a good edit.

Exposure differs by adapter, and one of the three has no defense:

| Adapter | Source | Noise defense |
|---------|--------|---------------|
| `pytest_results` | `pytest tests/hooks` JUnit | Deterministic |
| `agent_results` | `eval-agent-vs-baseline.py` report | `reduce` over per-run rates, mean by default |
| `rule_results` | rule eval scenario scores | **None. Single-shot.** |

`rule_results` reads one score block per scenario and thresholds it. The score comes from an LLM judge, which is the least deterministic of the three inputs and the only one with no averaging.

## Why this matters more than it looks

An adversarial review (gemini-3.1-pro-preview, PR #3430) modeled this as a compounding spurious-rejection rate across a loop. Its arithmetic assumed all three adapters were single-shot, which is wrong for two of them, so the headline number does not hold. The concern is real for the rule path specifically.

The asymmetry is worth stating: a spurious reject costs one rollout. A spurious accept ships a regression and raises the baseline the next edit is measured against. That asymmetry is why the fix belongs in the scorer rather than in a looser gate.

## Proposed

Give `rule_results` the same shape `agent_results` already has: accept multiple runs per scenario and reduce them before thresholding. The reducer should be caller-selected, defaulting to mean, so a caller can choose majority vote or min where that is the right semantics.

## Out of scope

Widening the seam past booleans (#3437). This is orthogonal: a float seam with a single-shot judge has the same noise, it just carries it further.

Refs #3422

rjmurillo:
## Summary
- Run 3 judge samples for each rule activation response and persist every sample in score_samples.
- Reduce repeated judge samples with median by default, then write the reduced scores beside the samples.
- Make optimize-artifact extract read score_samples directly, reject failed samples, and reject missing sample scores instead of treating them as zero.

## Cost and reducer choice
I chose N=3 and median. Cost changes from 1 artifact call plus 1 judge call per scenario and mechanism to 1 artifact call plus 3 judge calls. That is 4 API calls instead of 2, so total calls double for this eval. N is odd, so median has no tie case.

Median reduces one noisy judge outlier without letting one low or high sample decide the verdict. Mean is more sensitive to one bad sample. Min is too conservative for noise reduction. Max is too permissive for a gate.

Errored or missing samples are not converted to zero. Failed samples persist and mark the mechanism as judge_failed. Missing sample score fields are a config error.

## RED and revert proof
- RED before fix: 6 new tests failed. The rule adapter ignored score_samples, failed samples passed as clean scores, missing sample fields passed, eval_one_scenario had no judge_repeats argument, and dry-run still counted 2 calls.
- Revert proof after fix: reverting only the implementation made 7 tests fail. The added edge test also failed because --judge-repeats was not recognized.

## Validation
- uv run --frozen python -m pytest tests/eval/test_eval_rule_activation.py tests/eval/test_optimize_artifact_cli.py -q: 268 passed.
- uv run --frozen ruff check scripts/eval/ tests/eval/test_eval_rule_activation.py tests/eval/test_optimize_artifact_cli.py: passed.
- uv run --frozen mypy scripts/eval/eval-rule-activation.py scripts/eval/_optimizer_adapters.py scripts/eval/optimize-artifact.py tests/eval/test_eval_rule_activation.py tests/eval/test_optimize_artifact_cli.py: passed.
- Full failure-set diff: main 15454 passed, 26 skipped, 0 failures. Branch 15461 passed, 26 skipped, 0 failures. Added failures: 0.
- Push verification: remote branch SHA matches local HEAD d6d4089b7bcaf3b4f83f53344d131e8d9d6ccee7.

Fixes #3445


rjmurillo:
owner: local:pr-autofix
session: session-3573-autofix
acquired_at: 2026-07-28T08:04:08Z
expires_at: 2026-07-28T08:19:08Z
base_sha: d6d4089b7bcaf3b4f83f53344d131e8d9d6ccee7

rjmurillo:
owner: none
session: session-3573-autofix
acquired_at: 2026-07-28T09:10:34Z
expires_at: 2026-07-28T09:10:34Z

Issue #3596, opened 2026-07-28
Pre-push hook and PR Validation enforce different commit ceilings


rjmurillo:
## Summary

The pre-push hook and the `PR Validation` workflow enforce different commit ceilings, so a branch can pass every local gate and fail CI on the same tip. Seen on #3579: pushed clean at 25 commits, `Validate PR` failed with "PR has 25 commits (limit: 20)".

## The divergence

`scripts/validation/git_hook_policy.py` (`_check_commit_limit`, around line 2621):

```python
limit = 40 if _contains_main_merge(update, repo_root) else 20
```

The workflow has no equivalent branch. It counts commits and compares against 20.

Merging `main` into a long-lived branch is the local policy's own designed relief for the ceiling, documented in ADR-008. A contributor who takes that relief gets a green push and a red PR, and the CI message tells them to split a branch the local gate just said was fine.

## Why it matters beyond the annoyance

The two gates encode different answers to the same question, and neither cites the other. A contributor cannot tell which is the contract. The local one is the more permissive, which is the wrong direction for a gate to disagree in: it lets work accumulate past the point where the split is cheap. #3579 hit exactly that, and by the time CI said so the branch was already the base for #3595, so splitting meant rewriting a pushed base.

## Options

1. Teach the workflow the same main-merge relief. Keeps ADR-008's intent, makes both gates agree.
2. Drop the relief from the hook. Simpler, but removes the only escape from a ceiling on a branch that is legitimately long, and the ADR added it deliberately.
3. Have the hook read the workflow's limit so there is one number in one place.

Three is the one that cannot drift again, and it is the most work. I have no standing to pick; this is a governance call.

## Repro

Take any branch past 20 commits, merge `origin/main` into it, push. The hook passes. Open a PR. `Validate PR` fails.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #3596: Pre-push hook and PR Validation enforce different commit cei
- **Type**: complexity_triage
- **Priority**: medium
- **Suggested Agent**: agent-architect
- Issue labeled with medium complexity.
Suggested agents: agent-architect, agent-devops, agent-qa

### Route issue #3596 to agent-architect, agent-devops, agent-qa
- **Type**: agent_routing
- **Priority**: low
- **Suggested Agent**: agent-architect
- Issue content suggests routing to: agent-architect, agent-devops, agent-qa
Complexity: medium

PR #3608, opened 2026-07-28
feat(eval): reduce a rule scenario across runs instead of trusting one judge

### LINKED ISSUE #3445: eval: rule adapter is single-shot against an LLM judge with no noise reduction

rjmurillo:
## Summary

Fixes #3445. Implements ADR-087 Open Requirement 6: read a rule scenario across
several scoring runs and reduce, instead of trusting one judge reading.

A live rule-path validation on 2026-07-27 scored identical rule text twice.
Thirteen of 24 tasks moved. Five crossed the pass threshold. Mean absolute
movement was 0.49 on a five-point judge scale. The two held-out gains behind
that run's false accept were the two largest movements in the benchmark. One
reading is not evidence at that noise level, so the gate was measuring judge
variance and calling it improvement.

## What changed

`rule_results_multi` takes one score list per run, reduces each scenario's
scores across runs, then applies the pass bar once. Reducers are `mean`, `min`,
`max`, and `median`. `extract --kind rule` accepts repeated `--input` and a
`--reduce` flag.

Reduction is over scores, not over verdicts. Thresholding each run and voting
discards the distance from the bar, which is the only signal that says whether
a disagreement was close. This also matches what the agent adapter already
does, so the two noisy adapters now read the same way.

One run stays exactly legal: `rule_results_multi([s])` equals `rule_results(s)`.
That equality is a test, not a claim, so the flag can be adopted without
rescoring every incumbent.

## The fail-closed decision

A scenario with evidence in some runs but not others is refused, not scored
False and not dropped.

False is actively unsafe in the incumbent direction. A judge error on the
incumbent's run reads as a failing scenario, so a candidate that merely ran
cleanly looks like a fail-to-pass improvement. That is precisely the spurious
accept this gate exists to prevent, arriving through the scorer rather than
around it. Dropping instead would silently reduce over a different sample size
per scenario.

Uniform absence stays False. That is what preserves the single-run equality
above.

## Findings from adversarial review

Rounds 38 and 39 ran against this branch with models other than the author's
(`gemini-3.1-pro-preview` at high, then `gpt-5.6-sol` at xhigh). Seven findings,
each reproduced before it was fixed and re-measured after.

**F1. A task id two scenarios can both produce.** `<rule>::<scenario>` is not
injective. Rule `a` with scenario `b::c` and rule `a::b` with scenario `c` both
render as `a::b::c`. A plain assignment kept whichever came last, and a dropped
held-out task shrinks the denominator the gate divides by, so a swallowed
failure read as a cleaner candidate. Now refused.

**F2. A JSON object stating the same key twice.** The decoder keeps the last
value, so the document states more than one thing and cannot be measured. Now
refused.

**F3, F6, F7. A stdout that cannot be written was reported as a verdict.**
Measured before: a payload over 64KB into a closed reader exited 1 with a
traceback; a small one exited 120 with "Exception ignored" and no diagnostic at
all. Measured after: both exit 2 and say why on stderr. A closed stream raises
`ValueError`, not `OSError`, which is why the outer handler catches both.

**F4. A quadratic duplicate scan.** Re-scanned the id list per element. Now one
pass.

**F5. A file nobody can parse had not declared an absent corpus.** The most
instructive of the set, because the F2 refusal leaked around itself. The corpus
peek promises never to raise on content and answered `None` both for "parsed,
declares no corpus" and for "would not parse." Those are different facts. The
conflict rule reads `None` beside a pin as the deliberate envelope strip, a
REJECT, so a malformed candidate came back `decision: REJECT` exit 1 with a
reason stating the two files disagree on a corpus they actually agree on. A
wrong verdict and a false explanation, on the exact input class the previous
commit was written to refuse.

The fix is a third answer, not a widened guard. A negative control test was
green from the start and stays green: it holds the anti-strip rule, so a fix
that bought the malformed case by selling the stripped one would fail it.

**F40-1 (round 40). The third answer covered the grammar, not the field.**
The F5 fix above was incomplete one input class over, in the function it had
just changed. `_UNREADABLE` was introduced for a file that would not parse. A
file that parses cleanly but carries a corpus the authoritative reader refuses,
a truncated digest, upper-case hex, an empty string, a number, still answered
`None`, which is the envelope strip. Measured before: the same wrong shape came
back, `decision: REJECT` exit 1 with a reason saying the two files do not agree
on one corpus, for a pair where neither declares one that can be compared, plus
advice to re-score both artifacts when the fault is one corrupt field.
Re-scoring is the expensive remedy in this system, so the false reason costs
judge budget on top of the wrong verdict. Measured after: exit 2, no decision,
config error naming the corpus.

The peek now mirrors the authoritative reader by construction rather than by
enumerating bad shapes, so the next value that reader learns to refuse is
covered without a second edit. The README already documented exit 2 for a
truncated or upper-case value, written two rounds earlier, so this restores the
documented contract rather than changing it. A documented claim with no test
behind it was false for two rounds before a reviewer read both.

Round 40 also cleared the multi-run reduction, the extract dispatch, and the
exit path explicitly, and found no second Critical.

**The audit behind F40-1.** Both this finding and F5 have the same shape, so
the follow-up asked the generalized question rather than patching the instance:
find every best-effort peek that shadows an authoritative reader with fewer
branches. Seven such pairs exist across the three files, and only two readers
anywhere swallow an exception to return a domain value; every other handler
re-raises. The corpus header was the only one whose missing branch could reach
something that emits a decision. The ledger turned out to have no pre-lock peek
at all.

That sharpens the rule worth keeping. Not "peeks must carry every branch",
which would demand edits to two readers that are safe as they are, but: a peek
is dangerous exactly when its answer can reach a decision emitter. The
degraded-rule scan has strictly fewer branches than the scorer it shadows and
costs only a blunter message, because everything it misses raises and `extract`
never emits a verdict. The buffer dedup skips a non-string fingerprint, which
biases toward re-gating an edit rather than banning one, the direction its own
docstring already argued for.

The audit also exposed a coverage gap in my own tests: every case for the
corpus refusal put the bad value on the candidate, and the guard reads both
headers. The symmetry test added for it is verified by mutation, deleting the
incumbent check makes it fail at exit 1 while the rest of the suite stays
green.

### A review thread on the parent PR, fixed here

A bot review on #3579 caught the adapter paragraph in `scripts/eval/README.md`
saying a skipped test scores as a failure "rather than being dropped", with no
qualifier, while `extract --kind hook` has always taken `--on-skip exclude`,
which drops exactly those. The flag appeared nowhere in the README, so a
sentence that was true of the default read as the whole contract.

Fixed on this branch rather than on #3579, matching the earlier decision for the
`_absent` thread: rebasing two stacked children to place a doc fix upstream
costs more than the thread staying open with an explanation on it.

The fix went past the clarification because there was an enabling defect
underneath. `scripts/eval/optimize-artifact.py` hardcoded `("fail", "exclude")`
beside the adapter's own `_SKIP_POLICIES`, so one policy set had two
representations. A third policy added to the adapter would have been refused by
argparse before the adapter ever saw it, and the operator would have read an
error naming the wrong layer. That duplicate is also why the README had two
places to drift from, and why the doc test could not be written cleanly until it
was gone. The command now imports the tuple, following the existing cross-module
convention for `_MAX_PASS_RATE` and `_RULE_SCORE_KEYS`.

Three tests, each verified by mutation rather than by passing: the README names
every value the flag accepts, the fingerprint cost appears beside the escape
hatch, and every accepted value survives the command end to end. The last is
driven through the CLI rather than read off `parser.choices`, since an
introspection test would stay green even if the value never reached the adapter.

This is the fourth prose-versus-code disagreement in this work and the first
where the prose was incomplete rather than mistaken. That is the harder
direction: nothing in the sentence was wrong, so nothing in it looked wrong.

## Two notes on method

The sentinel is a one-line class rather than a bare `object()`. With
`object()` the return type is `str | object | None`, which collapses to `object`
and checks nothing. With a named class it is `str | _Unreadable | None`, and
mypy immediately failed the call site where the sentinel flowed into a function
written against `str | None`. The bare version would have shipped that hole
silently.

Writing the README section is what found the silent exit. The sentence claiming
both payload sizes report on stderr was half false, and the small-payload mode
said nothing at all.

## Round 41: the correction carried the defect it was correcting

A fourth distinct model reviewed the previous round's documentation fix and
found the new paragraph false. It said `--on-skip exclude` "moves the split
fingerprint and stops the gate". Neither half is true, and the sentence was not
invented: it was copied from the `pytest_results` docstring while a different
sentence in the same file was being corrected, and never measured.

Measured against a real split: drop a held-out id from the candidate and the
gate returns exit 1 with `REJECT`, one consultation charged, `compared: false`,
and the fingerprint echoed back unchanged. Drop an id outside the held-out group
and the run proceeds with the drift invisible. So the fingerprint never moves,
because it pins what was split rather than what was extracted, and the gate does
not halt; it charges and returns a verdict that measured nothing. Both copies of
the sentence now state that.

Whether the gate should refuse before charging is filed as #3615 rather than
changed here. The ordering is deliberate, carries a comment explaining that a
refusal decided after reading the group would otherwise be free, and withholds
the missing ids because naming them leaks held-out membership. The review's
key-set proposal is split-independent and therefore sound, but charging policy
is architecture.

Underneath sat a defect introduced in the previous round while fixing this same
class. Replacing the literal default with `_SKIP_POLICIES[0]` removed a
duplicate and added a worse coupling: the order of a choice set became
load-bearing with nothing saying so. The audit that finding forced turned up two
older instances that the previous round had not looked for, including
`--min-score` repeating `3.5` while the canonical constant sat exported in
`__all__` for exactly that use. A defect class is not closed by fixing the
instance that surfaced it.

All three of the previous round's tests were mutation-verified and still
asserted the wrong thing, which is the sharper lesson. Driving each policy
through the command and asserting exit zero stays green when the command drops
the flag on the floor, proven by replacing the adapter in memory with one that
ignores its argument. They now assert the command's output equals the adapter's
own answer for the same input, with the reverse direction covering a value the
adapter never accepts.

Every window the documentation tests used was too wide in its own way. A
file-wide search passes on an unrelated historical mention, which is the state
this gap was found in. A character count changes meaning as correct prose grows.
Bounding at the next heading failed least visibly: the enclosing section already
contained the asserted word three paragraphs down, so the new assertion passed
on someone else's sentence. The window is now a blank-line-delimited paragraph,
the narrowest real structural unit Markdown offers. That was caught by mutation
rather than review, and only because two mutations passed together.

#3616 records seven flags absent from `scripts/eval/README.md`, one of which is
required with no help text. Measured by introspecting the built parser, after a
regex over the source misattributed help strings by matching past the end of
each call.

## Round 42: five mutation-verified tests that were all tautologies

Round 41 added five tests to catch a flag default declared twice. Each read the
constant off the module under test and compared it to the parser that same
module had just built. Both sides resolved through one module, so the assertion
said only that the module agreed with itself.

Every one of the five had been mutation-verified. Moving a shared constant moves
both sides of a tautology at once, so the test goes red and looks sound. The
reviewer's reproduction is what exposed it: delete the import and hardcode a
different value beside the parser, and the command answers `median` while the
adapter answers `mean` with all five tests still green.

A tautology is not insensitive to mutation. It is sensitive to the wrong thing,
and no mutation of what it does read will ever say so. Only a mutation that is
the defect exposes it. The owning modules are now loaded separately and the
comparison never resolves through the module under test.

The same round found six more duplicated defaults that two previous audits had
missed. Both audits searched for the shape of the previous fix, a named constant
to point at, rather than the shape of the defect, a value with two authors. The
six live in function signatures, so there was no name to grep. They are pinned
by one parametrized test that reads each owner's signature, not extracted into
six new constants that would trade six literals for six names.

## Round 43: which duplications can actually drift quietly

The re-anchored audit returned six more values with two authors, each with a
prediction of what would break. Four predictions were wrong. Renaming a verdict
string turns 31 tests red. Renaming a task group raises `AttributeError`.
Renaming a rule score key fails at runtime with a message naming the missing
keys. The pre-gate dict keys are asserted on every refusal path.

The two that survived are prose. The README spells the results schema and both
non-zero exit codes as literals, and prose is the author no mutation reaches.
Both are now pinned, each exit code anchored on its own sentence fragment
because the two share a paragraph and a bare search for the number passes on
the neighbour.

A duplication is a defect only when drift is both possible and silent. Possible
is cheap to establish by reading. Silent has to be measured.

## A guard measured and then not built

PR #3595 squash-merged into this branch's base, which rewrote every commit hash
here. That broke a hash cited in a review reply, the second time this stack has
produced a dead hash reference. The first was a session log naming a commit an
amend had replaced.

The obvious fix is a validator that fails when a hash in a durable artifact no
longer resolves. It is cheap enough to write that it was worth measuring first.
Over the session log and the debate log, a hex token rule found forty six
candidates, thirty five resolving to commits and eleven not. All eleven were
false positives: the p-value `0.0078125`, the interior of the word "succeeded",
dates and random suffixes lifted from run directory names, and truncated corpus
fingerprints. True positives were zero, because both real instances had already
been corrected by hand.

A gate whose only finding bucket is entirely false on its first run earns an
ignore list, then a skip flag, then no readers. Prose does not record intent,
so a detector over prose can only match shape, and hex is a shape that dates,
probabilities, digests and ordinary English words all share. The conclusion is
recorded rather than the gate: cite pull request numbers, which survive every
rewrite and cannot be mistaken for a p-value.

## Verification

- 1722 tests passing, up from 1648 on the base branch.
- mypy clean on all three changed modules.
- ruff clean on every file in this diff.
- No `type: ignore` anywhere in the diff. Three that existed during development
  were removed idiomatically rather than justified: two traversals became one,
  and a file-descriptor redirect became an object swap after measuring the two
  identical on both failure modes.
- All seven findings re-reproduced end to end against the final tip, and
  round 41's critical reproduced with a real split before it was accepted.
- Seven mutations verified against the round 41 tests. The two that passed
  were treated as a finding about the assertion window, not the tests.
- Six mutations against the round 43 pins, one per author of each value,
  all red, with an unmutated negative control still green. The first two
  passes of that matrix disagreed with each other, which was a defect in the
  harness: the module loads through `spec_from_file_location`, whose bytecode
  cache is keyed on source size and mtime in whole seconds, and every
  mutation was one character for another. Caches are now cleared per case.

## Notes

History was rewritten before the first push, which is safe because nothing had
been pushed. Two push gates bind at once on a three-branch stack: the commit
ceiling counts the whole stack from main and its only relief is a label on an
open PR, which cannot exist before the first push, while the suppression gate
refuses any commit that adds a suppression and has no relief at all. Since
intermediate commits added the suppressions and only the tip removed them, no
partial push was clean either. The rewrite is verified lossless: the tree is
byte-identical to the pre-rewrite tag, and the intermediate commit was
confirmed green before it was made. See #3596 for the gate interaction.

#3595 has merged. It was squashed, so this branch was rebased onto the
squashed commit to drop the originals it now contains. The rebase is verified
lossless: the tree is byte-identical to the pre-rebase tag and the suite is
green at the new tip. Base is now #3579, which should be reviewed first.

## Changed files

Source:

- `scripts/eval/_optimizer_adapters.py`: `rule_results_multi`, the reducers, the shared per-run score reader, the fail-closed policy on partial evidence.
- `scripts/eval/_optimizer_core.py`: refuses a task id two scenarios can both produce (F1), and scans for duplicates in one pass (F4).
- `scripts/eval/optimize-artifact.py`: the multi-input extract surface, the duplicate-key refusal (F2), the unreadable-corpus sentinel (F5), and the exit path (F3, F6, F7).

Tests:

- `tests/test_eval_optimizer_adapters.py`: reduction, single-run equality, the partial-evidence refusal.
- `tests/test_eval_optimizer_core.py`: the task-id collision cases.
- `tests/eval/test_optimize_artifact_cli.py`: the CLI surface, plus subprocess drivers for the three exit-path findings, which a unit test cannot catch because pytest restores stdout before interpreter shutdown, plus the flag single-sourcing class.

Docs:

- `scripts/eval/README.md`: the corpus contract and the `--on-skip` cost.


rjmurillo:
Applying \`commit-limit-bypass\`.

This branch sits at exactly 40 commits from main, which is the ceiling, and only because a main merge is in range (without it the limit would be 20). The count is the whole three-branch stack: #3579 contributes 25 and #3595 another 8 before this branch adds its own work. Splitting further would mean a fourth branch whose only content is a one-field correction to a session log.

The remaining commit corrects `endingCommit` in the session log, which points at a sha the history rewrite destroyed. That is a document-of-record defect worth one commit. Filed #3610 for the underlying gate interaction, since the first push of a stacked branch cannot reach this relief at all: the label needs an open PR, and the PR needs the push the ceiling blocks.

Issue #3635, opened 2026-07-28
Push suppression gate permanently blocks renaming any file that contains suppression tokens


rjmurillo:
The pushed-suppression gate (`check_pushed_suppressions` in `scripts/validation/git_hook_policy.py`) scans added lines for suppression comments (`# noqa`, `# nosec`, `# type: ignore`, and friends). It diffs with `--no-renames`, and a test on `main` asserts that flag is present, so the choice is deliberate: a rename must not launder a suppression past the gate.

The side effect is that a **pure rename** of a file containing any suppression is permanently blocked. `--no-renames` makes the destination path read as a wholesale add, so the pre-existing suppression is reported as newly added. There is no allowlist, no justification comment, and no environment escape hatch on this check.

## How it surfaced

Removing the Tier 3 causal memory graph (ADR-088) left a test file whose name no longer matches its contents.

> [!NOTE]
> The file is `tests/validation/test_git_hook_policy_causal_restore.py`. It held five classes. Three covered the causal graph snapshot-and-restore path and went with the graph. The other two, `TestPushedSuppressionPolicy` and `TestAdrReviewPolicyMergeScope`, never touched causality; they shared the file only because both exercise `git_hook_policy`.

Renaming it to match its contents is blocked by one line, the `sys.path` shim import that carries a real `# noqa: E402`. That suppression was reviewed when it landed years of commits ago. Under a rename the gate reports it as new.

## Why the current behavior is not simply correct

The gate cannot distinguish two cases it should treat differently:

1. A contributor renames a file to smuggle a suppression past review. This must be blocked.
2. A contributor renames a file and changes nothing else. The suppressions were reviewed when they landed. Nothing new is being introduced.

Case 2 is currently unfixable without weakening the gate globally.

## Suggested direction

Detect the rename separately from the diff the gate scans. Run a rename-aware pass first; for any path pair where similarity is 100 percent, drop the destination from the scan set, because no content changed. Keep `--no-renames` for everything else so a rename-plus-edit is still scanned in full. That preserves the anti-bypass property, because a rename that also adds a suppression is no longer a pure rename and stays in the scan set.

## Related trap found while investigating

The fixtures in that file split their suppression tokens across adjacent string literals on purpose, so the gate does not flag the file that tests it:

```
comment = "# no" "qa"
comment = "# type" ": ignore"
```

[...]

Issue #3631, opened 2026-07-28
ADR-063 kill gate reads a knowledge-transfer score as evidence of downstream utility


rjmurillo:
## Summary

The ADR-063 kill gate certified skills on a knowledge-transfer eval and the
result was read as evidence that the skill's output is useful. Those are
different questions. A documentation-heavy skill passes the gate by having
documentation, whether or not anything consumes what it produces.

## Evidence

`evals/reports/adr-063-kill-gate-20260708/memory-reflexion.json` scored the
memory-reflexion skill 1.17 baseline against 4.83 enhanced and passed. All six
prompts ask the model questions about the skill's own reference documentation
and then measure whether loading that documentation improves the answers.

It does, reliably. That is the harness working as designed: it measures
documentation knowledge transfer. Fifteen other skills passed the same run on
the same basis, so this is not a bad score or a broken instrument.

The problem is the inference drawn from it. ADR-063 used the score as a
behavior gate and concluded the tier earned its keep. The eval never asked
whether any consumer reads the artifact the skill produces. ADR-088 later found
that nothing did, and the graph had been carried for four more months on the
strength of a passing gate that could not have detected the defect.

## Why it is not fixed by ADR-088

The removal deletes the tier. It does not change the gate. The next
documentation-heavy capability will be certified the same way, for the same
reason, and the gate will be equally unable to see whether its output has a
reader.

## Suggested direction

Split the two questions rather than conflate them.

1. Keep the knowledge-transfer eval as-is, and stop describing a passing score
   as evidence of downstream utility.
2. Add a separate, cheap consumer check for any capability that writes a
   committed artifact: name the reader, and fail when the only callers are the
   capability's own tooling, tests, and documentation. That check would have
   caught this tier on day one and takes a grep to run.

The second one generalizes. #3630 shows the same defect one tier down.

## Related

- ADR-063 (the gate)
- ADR-088 Finding 5 (where this was found)
- #3630 memory Tier 2 is write-only, same detection gap

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #3631: ADR-063 kill gate reads a knowledge-transfer score as eviden
- **Type**: complexity_triage
- **Priority**: medium
- **Suggested Agent**: agent-architect
- Issue labeled with low complexity.
Suggested agents: agent-architect

### Route issue #3631 to agent-architect
- **Type**: agent_routing
- **Priority**: low
[...]

Issue #3630, opened 2026-07-28
Memory Tier 2 is write-only: no code reads the episode store


rjmurillo:
## Summary

No code reads the Tier 2 episode store. 275 committed episode files, roughly
1.5 MB, are written by a hook on every session-log commit and consumed by
nothing but their own tests and documentation examples.

This surfaced during the ADR-088 review, which removed the Tier 3 causal graph
on the grounds that nothing read it. Two drafts of that ADR claimed episodes
were different because they had readers. Both claims were false, and the review
caught both. The honest finding is that the read path is documentation-only for
the whole memory system, not just the layer that was removed.

## Evidence

The episode query API lives in `memory_core/reflexion_memory.py`:
`get_episode`, `get_episodes`, `new_episode`, `get_decision_sequence`,
`get_reflexion_memory_status`.

```bash
git grep -n "get_episodes\|get_episode(\|get_decision_sequence" -- '*.py' \
  | grep -v "memory_core/\|/tests/\|test_"
```

returns nothing. The only non-test callers anywhere are prose examples in
`memory-search/references/quick-start.md`.

Three candidates were checked and none reads episode content:

| Candidate | What it actually does |
|---|---|
| `rework_warning.py` | Excludes the episode path prefix from a churn signal. Reads no episode. |
| `validate_investigation_claims.py` | Allowlists the episode path prefix. Reads no episode. |
| `git_hook_policy.py` | Generates and stages episodes. It is a writer. |

Four PowerShell cmdlets are documented as the query surface and none is
defined anywhere in the repository. Tracked separately in #3623.

## Why this matters

The system pays a real write cost on every session-log commit and gets no read
benefit. That is the same failure ADR-088 removed one layer of. The difference
is that episodes are the primary record and the graph was derived from them, so
the fix here is probably not deletion.

## Not a proposal to delete

Episodes are one derivation step closer to the source than the graph was, they
are many small files rather than one rewritten blob, and their content is a
record of what happened rather than an aggregate that measured empty. ADR-088
explicitly declined to settle this and said it needs its own evidence.

The three candidate resolutions, in rough order of appeal:

1. Wire a real reader. The obvious one is the memory router: teach
   `search_memory.py` to search episodes alongside Serena and Forgetful. It
   currently does not touch them.
2. Keep writing them as an audit record and stop claiming they are queryable.
   That means correcting the docs and accepting the write cost as the price of
   the record.
3. Stop writing them.

[...]

Issue #3625, opened 2026-07-28
Two .gitattributes merge drivers are named but never registered, so they silently fall back to text merge


rjmurillo:
## Problem

`.gitattributes` names two merge drivers that no code registers:

```
.agents/HANDOFF.md merge=ours
.agents/handoffs/*.md merge=handoff-aggregate
```

Git resolves a driver name through `git config merge.<name>.driver`, which is per clone and never committed. When the lookup fails, git silently falls back to the default text merge. Both entries therefore do nothing today.

## Two distinct defects

**`merge=ours` names a driver that does not exist.** `ours` is a merge strategy (`git merge -s ours`) and a strategy option (`-X ours`), not a built-in low-level merge driver. Git's built-in driver names are `text`, `binary`, and `union`. So this line does not "keep our version on conflict"; it produces a normal text merge and therefore a conflict, which is the opposite of the intent the name advertises. Verified empirically: a file marked `merge=ours` with no registration conflicts rather than resolving.

**`merge=handoff-aggregate` is declared but was never implemented.** No script defines it:

```bash
git grep -c "handoff-aggregate" -- scripts/   # no matches
```

## Impact

Anyone reading `.gitattributes` reasonably concludes these two paths have conflict-resolution behavior. They do not. `.agents/HANDOFF.md` is a frequently-updated dashboard, so the file most likely to conflict is the one whose declared protection is inert.

## Suggested fix

Pick one per line: implement and register the driver, or delete the attribute so the file's behavior matches its declaration. A validator that cross-checks every `merge=<name>` in `.gitattributes` against the set of names some committed script registers would catch both, and would have caught them at authoring time.

## Context

Found while removing the Tier 3 causal graph under ADR-088. That change deleted the one registrar this repository had, `scripts/maintenance/install_merge_drivers.py`, but that registrar only ever registered a single driver named `causal-graph`. It never registered these two, so the removal leaves them exactly as broken as they already were rather than causing the defect.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #3625: Two .gitattributes merge drivers are named but never registe
- **Type**: complexity_triage
- **Priority**: medium
- **Suggested Agent**: agent-architect
- Issue labeled with low complexity.
Suggested agents: agent-architect, agent-implementer

### Route issue #3625 to agent-architect, agent-implementer
- **Type**: agent_routing
- **Priority**: low
- **Suggested Agent**: agent-architect
[...]

PR #3643, opened 2026-07-28
refactor(memory): remove Tier 3 causal memory graph

### LINKED ISSUE #3345: causal-graph.json conflicts on every merge of main: generated state committed as a single file
Every merge of `main` into a feature branch conflicts on `.agents/memory/causality/causal-graph.json`. Hit three times in one session, on three unrelated branches, each time on a merge whose only purpose was to clear an unrelated conflict.

```
Auto-merging .agents/memory/causality/causal-graph.json
CONFLICT (content): Merge conflict in .agents/memory/causality/causal-graph.json
```

## Why it always conflicts

The file is generated state committed as one blob. The `update-causal-graph` pre-commit job rewrites it on essentially every commit, so both sides of any merge have touched it. Git has no way to reconcile two rewrites of a single JSON document, so it hands the developer a conflict in a file no human authored and no human should resolve by hand.

The resolution is mechanical every time: take one side, rerun the generator, stage the result.

```
git checkout origin/main -- .agents/memory/causality/causal-graph.json
uv run --frozen python scripts/validation/git_hook_policy.py update-causal-graph
git add .agents/memory/causality/causal-graph.json
```

That the resolution is always identical is the tell. A conflict with a deterministic resolution is not a conflict; it is a missing merge strategy.

## Cost

Every contributor who merges `main` pays this. The failure mode for someone who does not know the recipe is worse than the delay: hand-editing a generated JSON graph produces a file that is neither side, passes commit because the generator will overwrite it later anyway, and silently corrupts whatever reads the graph in between.

## Options

1. **Merge driver.** Register a `.gitattributes` merge driver for this path that discards both sides and reruns the generator. Fixes it for everyone with no per-developer knowledge. This is what generated-lockfile ecosystems do.
2. **`merge=union` or `-merge`.** Cheaper, but wrong for structured JSON; union produces invalid documents.
3. **Stop committing it.** Treat the graph as a build artifact, regenerate on demand, and gitignore it. Depends on whether anything reads it from a checkout without running the generator first, which needs measuring before proposing.
4. **Shard it.** One file per episode instead of one graph. Removes the conflict surface but changes every reader.

Option 1 is the smallest change that eliminates the class.

## Acceptance criteria

- [ ] A merge of `main` into a branch that has committed since the fork point does not produce a conflict on this path.
- [ ] The post-merge graph is byte-identical to a fresh generator run, so the merge driver cannot leave stale state.
- [ ] A test covers the merge-driver path, including the case where the generator fails and the merge must not silently take a side.

## Notes

Not urgent enough to block, common enough that everyone pays it. Filed while resolving it for the third time in one session.

rjmurillo:
## Summary

Removes the Tier 3 causal memory graph and every script, schema, merge driver, hook, and test that existed to maintain it.

The graph is a derived cache that nothing reads. No code path anywhere in the repo loads it to make a decision. Its only consumers were the scripts that wrote it and the tests that tested those scripts. It cost a 1,074,565-byte JSON artifact touched by 103 of the last 1834 commits, a custom merge driver, two lefthook jobs, a repair script, and a schema, and returned nothing a reader ever consumed.

Fixes #3345
Refs #3502

## Evidence

The read boundary was verified empirically, not assumed.

```
git grep -n "causal_graph\|causal-graph" -- '*.py' \
  | grep -v "update_causal_graph\|merge_causal_graph\|repair_causal\|/tests/\|test_"
```

Returns nothing. Every remaining hit was the writer, the merger, the repairer, or a test of those three.

The content was measured before deletion, so the decision rests on what the graph actually held rather than what it was designed to hold:

| Measure | Value |
|---|---|
| Nodes | 2515 |
| Nodes touched by any edge | 99 (3.94%) |
| Edges | 79, all of one type |
| Edges whose only contribution is a legacy placeholder | 46 of 79 |
| Edges with no contribution map at all | 4 |
| Patterns | 6, of which 5 are legacy |
| Patterns where trigger equals action | 2 |

A graph in which 96% of nodes are unreachable and a quarter of the patterns are tautologies is not an underused asset. It is noise.

## Review

Six agents debated ADR-088. Four returned Block on the first round. Every Block was aimed at the ADR's evidence, never at the decision, and every one was correct.

> [!NOTE]
> The debate log is at `.agents/analysis/2026-07-27-adr-088-causal-tier-removal-debate.md` and the decision record is at `.agents/architecture/ADR-088-remove-causal-memory-tier.md`.

The review caught two fabrications in my own writing. I twice asserted a set of episode consumers by name without checking whether they read episodes. None of them do. The corrected section is verified against a grep that is quoted inline so a reader can rerun it.

## Two defects the review surfaced, both filed rather than fixed here

- **#3630** Episodes are write-only too. The same grep applied to `get_episodes`, `get_episode(`, and `get_decision_sequence` returns nothing outside the memory package and its tests. The read path is documentation-only end to end, which is a larger finding than the graph itself.
- **#3631** The kill gate that certified this tier could not have detected the defect it was meant to catch.

## Inverse tests

`tests/test_causal_tier_removed.py` holds 22 tests that fail if any part of the tier returns: the artifact, the scripts, the schema, the merge drivers, the lefthook jobs, the imports, the re-exports. Each was negative-controlled by reverting the deletion and confirming the test fails.

## A regression I introduced and the merge caught

I deleted a test file on the strength of its name. It held five classes; three were causal and two were not. The two survivors had been hardened on main four days earlier and had nothing to do with causality. I dropped 25 tests.

The merge with main surfaced it. The tests are restored at their original path with the causal classes removed in place, and test 22 of the inverse suite now asserts both surviving class names are still collected there, so the filename trap cannot spring twice.

> [!NOTE]
> The file keeps its now-stale name, `tests/validation/test_git_hook_policy_causal_restore.py`. Renaming it is blocked by the push-suppression gate, which diffs with `--no-renames` to close a rename-based bypass; under a rename its one real suppression reads as newly added. Filed as #3635 with a suggested fix that keeps the anti-bypass property.

## Latent defects left in place, deliberately

Deleting the graph exposed two adjacent bugs in the same `.gitattributes` file. Both predate this change and both are filed rather than folded in.

- **#3625** `merge=ours` names a driver that does not exist, and `merge=handoff-aggregate` is declared but never implemented.
- **#3624** The ADR-007 tier taxonomy cited across the memory skills is not in ADR-007.

## Scope

80 files. The push scope gate blocks at 50 and names `SKIP_SCOPE_CHECK=1` as its sanctioned bypass for justified large PRs. Three reasons this qualifies:

1. **Atomicity.** There is no ordering of these deletions that keeps every intermediate commit green. A deleted writer with a live schema, or a deleted graph with code still importing it, is a broken tree.
2. **Mirror doubling.** 62 of the 80 sit in the two plugin trees the generator holds in byte parity, so every real edit is counted twice by repo design.
3. **Shape.** 20 changes are the deletions themselves, and most of the remaining volume is one 1MB JSON artifact, not reviewable logic.

The scope heuristic was bypassed. The security push-suppression gate was not; it passes at exit 0 on its own merits.

## Validation

- Full suite: 15777 passed, 26 skipped, 0 failed
- Pre-PR validation: all validations passed
- Both plugin manifests bumped to 0.6.145
- A security agent reviewed the `lefthook.yml` and `scripts/validation/git_hook_policy.py` changes and returned Accept

## Recovery

Recovering this tier means reading git history and reimplementing, not running one command. That is the intended cost. If the graph is ever wanted again, it should be rebuilt against a real reader, because the absence of one is why it is being removed.


rjmurillo:
owner: local:pr-autofix
session: 2026-07-28-session-3608
acquired_at: 2026-07-28T08:08:34Z
expires_at: 2026-07-28T08:23:34Z
base_sha: 87b1176674a8f444a5391a34e9cac187d7af024f

rjmurillo:
Applied `commit-limit-bypass`.

10 of the 19 commits on this branch are mechanical merges of main. They exist because `causal-graph.json` is modified by roughly 6 of every 8 commits on main, so this branch hits a modify/delete conflict on nearly every merge and has to resolve it again each time. The recurrence is the artifact this PR deletes charging its maintenance cost one last time.

The merge method is squash, so all 19 collapse to one commit on main. The review surface is the diff, which has not grown: the extra commits are conflict resolutions that all resolve the same way, keeping the file deleted.

PR #3793, opened 2026-07-29
fix(scripts): drop PyYAML from the ADR status parser

### LINKED ISSUE #3791: sync_adr_protocol.py imports yaml on a bare-python3 documented entrypoint
## Summary

PR #3785 (mine) added `import yaml` to `scripts/sync_adr_protocol.py`. Before
that PR the script was standard-library only. Two repository documents tell
contributors to run it as bare `python3`, and this repository's own guidance
names `yaml` as the canonical example of a module that does not resolve under
bare `python3`.

## Evidence

Verified on `origin/main` at 7fc160420e:

- `scripts/sync_adr_protocol.py:24` is `import yaml`.
- `git show 9407d45150:scripts/sync_adr_protocol.py | grep -c "^import yaml"` returns `0`.
  That commit is the base of PR #3785, so the dependency is new.
- `CONTRIBUTING.md:829` is the line `python3 scripts/sync_adr_protocol.py`.
- `.agents/SESSION-PROTOCOL.md:1163` reads "Run `python3 scripts/sync_adr_protocol.py` to".
- `.github/copilot-instructions.md` states: "Prefer `uv run python` over bare
  `python3` for scripts that import project or third-party deps (for example
  `yaml`). Those resolve only in the project venv, so bare `python3` can fail
  `ModuleNotFoundError`."

## Scope, stated precisely

`PyYAML==6.0.3` **is** a declared dependency (`pyproject.toml:14`), so every
`uv run` invocation resolves it, and CI is not affected. On the machine where
this was found, bare `python3` also resolves yaml from a linuxbrew
site-packages directory, so the failure does not reproduce there either.

The reproducible failure is `python3 -S scripts/sync_adr_protocol.py`, which
raises `ModuleNotFoundError: No module named 'yaml'`. `-S` disables
site-packages and stands in for any interpreter that lacks PyYAML. The defect
is therefore a portability regression on the documented entrypoint, not a
current outage.

## Why fix rather than redocument

The parser reads exactly one scalar out of the frontmatter block. It does not
need a YAML document model. Removing the import costs less than changing two
documents and leaves the script runnable on any interpreter.

## Second defect in the same function

`_status_from_frontmatter` tests the opening fence with
`content.startswith("---\n")`. The canonical delimiter check in this
repository is
`.claude/skills/adr-review/scripts/detect_adr_changes.py:182`
`_split_frontmatter`, which compares `lines[0].strip() != FRONTMATTER_DELIM`.
The stricter form drops CRLF files and fences carrying trailing whitespace.

## Provenance

Both items were raised by a round-2 adversarial review of PR #3785 on a
different model family. The review returned before the PR merged; the merge
landed while the fixes were still being verified.

rjmurillo:
## Summary

Fixes #3791.

PR #3785 (mine) added a PyYAML import to `scripts/sync_adr_protocol.py`, which
had been standard library only. A round-2 adversarial review on a different
model family flagged it. The review returned before that PR merged, but the
merge landed while the fixes were still being verified, so the correction ships
here.

The parser reads one scalar out of the frontmatter block. It never needed a YAML
document model. It now reads that block with the standard library.

## Scope, stated precisely

My first draft of this change overstated the problem, so here is the measured
version.

PyYAML is a declared project dependency, so every `uv run` invocation resolves
it and CI was never affected. On the machine where this was found, bare
`python3` also resolves yaml from a linuxbrew site-packages directory, so it
does not reproduce there either.

The reproducible failure is `python3 -S scripts/sync_adr_protocol.py`, which
raises `ModuleNotFoundError`. The `-S` flag disables site-packages and stands in
for any interpreter that lacks PyYAML. This is a portability regression on a
documented entrypoint, not an outage. It is worth fixing because removing one
import costs less than changing two documents, and because the resulting script
runs anywhere.

## Second finding, same function

The opening fence test was `content.startswith("---\n")`. That form drops CRLF
files and fences carrying trailing whitespace. It now matches the canonical
delimiter check used elsewhere in this repository, which compares the stripped
first line. The canonical source is cited and quoted in the docstring.

## The optional finding I implemented, measured, and rejected

The reviewer suggested structural stop characters when joining a status that
wraps across lines. I implemented it. It regressed two real ADRs.

ADR-039's status opens with bold `**PROVISIONAL**`, so the `*` stop read bold
emphasis as a bullet and the status became Unknown. ADR-072 carries a 430
character prose status that truncated mid sentence.

I then measured whether the guard protected anything. A scan of all 90 Status
sections found zero occurrences of `###`, a list dash, a list star, a code
fence, a block quote, or a pipe anywhere in a joined block. The guard defended
against shapes that do not exist in the corpus while breaking two that do. Only
the heading stop survives, because a leading hash cannot collide with emphasis.
The rejected alternative is documented in the parser docstring so it is not
reintroduced.

## Verification

- 50 tests pass. Ruff check and mypy are clean.
- `python3 -S scripts/sync_adr_protocol.py` reports 0 Unknown across all 90 ADRs.
- Corpus output is byte identical to the previously measured good run, so this
  change is behavior preserving on real inputs while removing the dependency.
- Negative control: reverting the parser to the pre-#3785 heading only
  implementation fails 12 of the 50 tests, and restoring is byte identical.
- Restoring `import yaml` fails both new guards.

## New guards

An AST check asserts the module imports nothing outside
`sys.stdlib_module_names`. A subprocess test runs the script under `python3 -S`.
Together they pin the entrypoint contract so a future third party import cannot
land silently.

The corpus truncation test now imports the dangling word list from the module
instead of re-declaring it, so the parser and its test cannot drift apart.

## Evidence

Paths consulted while verifying the above, none of them modified by this change:

- `CONTRIBUTING.md` line 829 invokes the script as bare `python3`.
- `.agents/SESSION-PROTOCOL.md` line 1163 does the same.
- `.github/copilot-instructions.md` names yaml as the example module that fails
  under bare `python3` and says to prefer `uv run python`.
- `pyproject.toml` line 14 declares `PyYAML==6.0.3`.
- `.claude/skills/adr-review/scripts/detect_adr_changes.py` line 182 holds the
  canonical frontmatter delimiter check.
- `.claude/rules/canonical-source-mirror.md` requires citing and quoting a
  canonical source when claiming to match it.

### DIFF
--- a/scripts/sync_adr_protocol.py
+++ b/scripts/sync_adr_protocol.py
@@ -21,8 +21,6 @@ import sys
 from dataclasses import dataclass, field
 from pathlib import Path
 
-import yaml
-
 # Exit codes per ADR-035
 EXIT_SUCCESS = 0
 EXIT_GAPS_DETECTED = 1
@@ -37,6 +35,15 @@ RFC2119_PATTERN = re.compile(
 # ADR number extraction from filename
 ADR_NUMBER_PATTERN = re.compile(r"ADR-(\d+)")
 INLINE_STATUS_PATTERN = re.compile(r"^\*\*Status\*\*\s*:\s*(.+?)\s*$", re.MULTILINE)
+
+# Matches `_split_frontmatter` in
+# .claude/skills/adr-review/scripts/detect_adr_changes.py, which compares
+# `lines[0].strip()` against this same literal.
+FRONTMATTER_DELIM = "---"
+
+# A status ending on one of these is a wrapped line, not a finished value.
+# Used by the corpus regression test to detect truncation.
+DANGLING_WORDS = frozenset({"by", "and", "on", "the", "a", "to", "of", "in", "for", "with"})
 
 
 @dataclass
@@ -106,8 +113,17 @@ def parse_adr_title(content: str) -> str:
 def _status_from_heading(content: str) -> str | None:
     """Read the prose under a ``## Status`` section, joining wrapped lines.
 
-    ADR-004 wraps its status across two lines, so reading only the first
-    would report the truncated "Superseded by".
+    ADR-004 wraps its status as "Superseded by" / "[ADR-086](...) on
+    2026-07-20.", so reading one line reports the truncated "Superseded by".
+    The block ends at a blank line or the next heading.
+
+    Rejected alternative: join only while the text ends on a word from
+    :data:`DANGLING_WORDS`. That reads ADR-004 more precisely, but measured
+    against the corpus it truncated ADR-072 mid-sentence at "this" and left
+    ADR-039 unreadable, because ADR-039 opens with ``**PROVISIONAL**`` and a
+    bullet-style stop cannot tell bold emphasis from a list marker. Stopping
+    only at a heading loses nothing: no ADR's status block contains a
+    subheading, list, fence, quote, or table row.
     """
     in_status = False
     collected: list[str] = []
@@ -116,14 +132,15 @@ def _status_from_heading(content: str) -> str | None:
         if stripped == "## Status":
             in_status = True
             continue
-        if in_status:
-            if stripped.startswith("## "):
+        if not in_status:
+            continue
+        if not stripped:
+            if collected:
                 break
-            if not stripped:
-                if collected:
-                    break
-                continue
-            collected.append(stripped)
+            continue
+        if stripped.startswith("#"):
+            break
+        collected.append(stripped)
     return " ".join(collected) or None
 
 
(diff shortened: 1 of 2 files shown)


rjmurillo:
Closing unmerged. An adversarial review on gpt-5.6-sol found this fix is the wrong fix, and I reproduced every finding.

## Why this is wrong

Replacing `yaml.safe_load` with a hand-rolled scalar reader is not behaviorally equivalent. Measured, old parser vs this PR's:

| input | main (yaml) | this PR |
|---|---|---|
| `status: null` | `None` (falls through) | `'null'` |
| `status: 123` | `None` (falls through) | `'123'` |
| `status : accepted` | `'accepted'` | `None` |
| `status: "accepted \\"conditionally\\""` | `accepted "conditionally"` | `accepted \\` |
| duplicate `status` keys | last wins | first wins |
| malformed sibling `broken: [` | `None` (rejects doc) | `'accepted'` |

Six divergences on a script that decides ADR lifecycle state. The corpus does not exercise them today, which is exactly why they would ship silently.

It also introduced a regression. I changed the heading terminator from main's `startswith("## ")` to a bare `#`, so a plausible ADR continuation line truncates the status:

```
## Status

#3791 accepted pending rollout
```

main returns the status. This PR returns `Unknown`.

## Two claims in the body were overstated

- "ruff check clean" was scoped ruff. Repo-wide `uv run ruff check .` reports 759 errors on main, unrelated to this branch.
- "`DANGLING_WORDS` prevents parser/test drift" is false. The parser never reads it; only a docstring mentions it. It is test-only data.

## The right fix

The repo already documents the answer at `.github/copilot-instructions.md:121`: prefer `uv run python` over bare `python3` for scripts that import third-party deps. PyYAML is a declared dependency (`pyproject.toml:14`), so the two documented invocations just need to name the right interpreter. That is a two-line docs change with no behavioral risk, versus ~90 lines of partial YAML reimplementation.

Refs #3791.

Issue #2941, opened 2026-07-07
docs(adr-034): amend investigation allowlist to reconcile scripts/modules/investigation_allowlist.py (9 patterns) with ADR-034 (5 patterns)


rjmurillo:
## Context

Found during a merged-PR resolution audit (2026-07-07, session 2608): every PR merged since #2737 was checked against its linked issue's stated problem and acceptance criteria at `main` @ `f0831c08`.

- Source PR(s): #2831
- Original issue: #2833

## Problem

Issue #2833 flagged as follow-up: scripts/modules/investigation_allowlist.py allows 9 path patterns but ADR-034 authorizes only 5, and two of the extra patterns (.agents/critique/, .agents/architecture/REVIEW-) are explicitly listed in the ADR's 'Not Allowed (require QA validation)' table. The ADR names the architect agent as allowlist owner with additions reviewed 'via ADR amendment'; no amendment exists. The enforcement code and its governing ADR contradict each other at HEAD, so any agent reading ADR-034 gets wrong QA-exemption rules. No issue tracks this; searches for open issues mentioning ADR-034 or investigation allowlist return nothing.

## Evidence (verified at `f0831c08`)

At HEAD f0831c08: scripts/modules/investigation_allowlist.py get_investigation_allowlist() returns 9 patterns including r'^\.agents/architecture/REVIEW-', r'^\.agents/critique/', r'^\.agents/memory/', r'^\.agents/memory/episodes/'. .agents/architecture/ADR-034-investigation-session-qa-exemption.md 'Investigation Artifact Allowlist' lists exactly 5 patterns and its 'Not Allowed' table includes '.agents/critique/' ('Plan reviews gate implementation decisions') and '.agents/architecture/'. Issue #2833 body: 'allows 9 patterns vs ADR-034's 5 (two marked Not Allowed by the ADR); needs an ADR amendment to reconcile.' Divergence originated from closed issues #831 (allow .agents/memory/) and #732 (critique false positives) which changed code without amending the ADR.

## Plan of action (sized for a Sonnet-class agent)

[...]

PR #3979, opened 2026-07-30
fix(protocol): teach the 8 investigation allowlist patterns the module enforces

rjmurillo:
## Summary

Seven surfaces taught a 5-pattern or 9-pattern investigation allowlist. The module enforces 8. A session that staged a path in the gap was told it was ineligible for the QA skip, or was told it was eligible when it was not, depending on which document it read.

The counts are now measured from the module instead of restated in prose, and 35 tests pin the agreement.

## What was measured

The module is the enforcement point. Its display accessor is the only count that binds:

```console
$ uv run python -c "from scripts.modules.investigation_allowlist import get_investigation_allowlist_display as g; print(len(g()))"
8
```

Every surface below was compared against that number by executing it, not by reading it.

| Surface | Taught | Actual |
| --- | --- | --- |
| `.agents/SESSION-PROTOCOL.md` | 5 | 8 |
| `ADR-034` body list | 5 in one place, 8 in another | 8 |
| `CONTRIBUTING.md` | 5 | 8 |
| `ai-agents-validation-and-qa` | 5 | 8 |
| `session-init` extraction reference | 5 | 8 |
| `session` skill, 3 JSON examples | 5 | 8 |
| `ai-agents-change-control` | 9 | 8 |
| `tests/evals/skills/triage-prompts.json` | 5 | 8 |

The eval entry is the one that graded backwards. Its expected answer listed 5 paths, so a model that answered correctly with 8 scored wrong.

## The 9-pattern case was stale, not fabricated

The change-control skill taught 9 patterns, named `.agents/memory/episodes/` among them, and cited a line range to argue the code had drifted wider than the ADR. All three statements were true when written.

Measured by executing the module's own accessor at each commit that touched it:

```console
a333cb70c (2026-02-12)  9
59e6587c4 (2026-06-04)  9
96e9bf7d2 (2026-07-08)  8
```

`96e9bf7d2` is PR #2958. It dropped the episodes pattern as redundant under the parent prefix, and the ADR amendment landed the same day. So the skill was one frozen snapshot of 2026-07-07, internally consistent, invalidated by a single commit it never absorbed.

The citation is the transferable lesson. The cited range covered the entire allowlist on 2026-07-07 and covers five sixths of it today, because three lines were appended and tagged with the amendment date. A line-range citation silently degrades into a truncated one when the block it cites grows, and a truncated range reads exactly like evidence of drift. That is how correct inputs produced a wrong conclusion.

An earlier commit on this branch described those three statements as falsehoods. That characterization was wrong and is corrected in `78a74f3aa`; the underlying edits were already right.

## Tests

19 tests became 35. The new ones pin each prose surface against the module rather than against each other.

```console
$ uv run python -m pytest tests/test_investigation_allowlist.py -q
35 passed
```

Each detector carries a negative control. The enumeration reader is span-scoped and compares by equality, not containment, because a containment check accepted a widened allowlist during review and would have let this class of drift back in.

## Verification

```console
$ git diff --name-only origin/main...HEAD | wc -l
20
$ uv run python build/scripts/build_all.py   # tree clean afterward
$ diff -q <canonical> <mirror>               # 5 pairs byte-identical
```

Both plugin manifests move to 0.6.5429, above main at 0.6.5428 and unheld by any open PR.

## Notes

Prose copies of the list are replaced with the accessor call wherever a count was being taught. One table row keeps a line range and is widened to cover the whole block.

Refs #2941
Refs #2958


rjmurillo:
Closing as superseded. `main` already teaches all 8 patterns and already records the divergence as closed.

`.claude/skills/ai-agents-change-control/SKILL.md` on `main`, lines 39 to 50:

```
The OPERATIVE investigation-only allowlist is the enforcement module
`scripts/modules/investigation_allowlist.py`

- `.agents/sessions/`                  (session logs)
- `.agents/analysis/`                  (investigation outputs)
- `.agents/retrospective/`             (learnings)
- `.serena/memories/`                  (cross-session context)
- `.agents/security/`                  (security assessments)
- `.agents/memory/`                    (memory artifacts)
- `.agents/architecture/REVIEW-*`      (review artifacts)
- `.agents/critique/`                  (critique outputs)

Former divergence, now closed: the ADR-034 text once listed only the first 5 paths, and ...
```

That is the same 8, in the same order, with the same "once listed only the first 5" framing this PR set out to add.

I found it by rebasing this branch onto current `main`. The rebase conflicted on `ai-agents-change-control/SKILL.md` and `references/provenance.md`, and reading the conflict showed both sides asserting the same corrected fact in different words.

This is the third superseded PR I have closed in this pass, after #4164 and #4102. The three were opened against different bases and each was independently fixed on `main` by other work while they sat. Filed as #4355 so the duplication is visible as a process problem rather than absorbed silently three times.

If some part of this branch is still missing from `main`, reopen with the specific behavior and I will check it against current `main`.

PR #3986, opened 2026-07-30
fix(ci): repair the memory health CLI contract and close its fail-open paths

### LINKED ISSUE #3971: fix(ci): memory-health.yml speaks to a CLI contract deleted in 2aeb4fddd and fails on every run
## Summary

`.github/workflows/memory-health.yml` has been broken since commit `2aeb4fddd` moved the `memory_enhancement` package out of `.claude/skills/memory-enhancement/src/` and into `scripts/`. The workflow was never updated. Every assumption it makes about the tool is stale, and the job fails on every run.

Live evidence: run 30526181117 on PR #3950 (a PR that touches no memory files) fails with `Could not read health-report.json: Expecting value: line 1 column 1 (char 0)`.

## Root cause chain

The package move replaced the implementation with a different one that has a different CLI shape and a different JSON schema. Four separate defects follow from that.

### A. `--repo-root` is passed after the subcommand, so argparse rejects it

`--repo-root` is defined on the top-level parser, before the subcommand. Lines 86 and 92 pass it after `health`.

Measured, identical tree, only argument position differs:

```
$ python3 -m memory_enhancement health --json --repo-root .
memory_enhancement: error: unrecognized arguments: --repo-root .
rc=2 ; bytes written to health-report.json: 0

$ python3 -m memory_enhancement --repo-root . health --json
rc=1 ; bytes written: 679
```

The sibling workflow `citation-verify.yml` line 67 already uses the correct order, so the correct form is established in-repo.

### B. The parse step reads a `summary` object that the surviving implementation never emits

`scripts/memory_enhancement/__main__.py::_cmd_health` emits nine top-level keys:

```
total_memories, total_citations, valid_citations, stale_citations,
broken_citations, unverified_citations, health_score,
stale_memories, recommendations
```

`scripts/ci/parse_memory_health_results.py` reads `payload["summary"]` and then `total`, `healthy`, `stale`, `exempt`, `errors`. None of those exist. The words `healthy` and `exempt` appear nowhere in `scripts/memory_enhancement/`.

The deleted implementation did emit `summary.healthy` and `summary.exempt` (see `git show c4c520c46`). The workflow still speaks to that deleted contract. Even with defect A fixed, all five outputs render as `null` and `has_stale` is always `false`, so the PR comment reports `Total memories: null` and the status is permanently `Pass`.

### C. `PYTHONPATH` points at a directory that no longer exists

Lines 86 and 92 set `PYTHONPATH=.claude/skills/memory-enhancement/src`. `git ls-tree origin/main .claude/skills/memory-enhancement/` returns only `SKILL.md` and `references`. The setting is inert; the module resolves only because CI installs the project.

### D. The `code` path filter watches a deleted directory

Line 51 filters on `.claude/skills/memory-enhancement/src/memory_enhancement/**`. That path cannot ever match. The adjacent `scripts/**/*.py` entry is what actually triggers the job. `memory-validation.yml` line 39 carries the same dead filter.

## Why this stayed invisible for months

Before the ADR-006 extraction, the parse step was inline shell guarded by `if [ -f health-report.json ]`. Measured against an empty report:

```
$ bash -e old.sh
total=[] stale=[]
old.sh: line 6: [: : integer expression expected
has_stale=false
rc=0
```

`jq` on an empty file emitted nothing, the integer comparison errored to stderr inside an `if` condition (which `set -e` does not abort on), the else branch ran, and the step exited 0. The workflow was green while measuring nothing.

Extracting that block to Python under ADR-006 replaced the silent pass with a hard failure, which is how the underlying rot surfaced.

## Also present: the issue #2808 defect, unfixed here

Lines 86 and 87 are:

```
python3 -m memory_enhancement health --json --repo-root . > health-report.json
echo "exit_code=$?" >> "$GITHUB_OUTPUT"
```

`health` returns 1 when it finds stale citations. Under `bash -eo pipefail` the step aborts before the `echo`, so `exit_code` is never written. This is exactly the defect issue #2808 fixed in `memory-validation.yml`. The fix was never applied here. The output is also never consumed by any step, so it is dead either way.

## Proposed fix

1. Move `--repo-root .` before the subcommand on both invocations.
2. Rewrite `parse_memory_health_results.py` against the schema `_cmd_health` actually emits, and update the PR-comment field mapping to match.
3. Drop the inert `PYTHONPATH` prefix.
4. Repoint the dead path filter at `scripts/memory_enhancement/**`.
5. Delete the dead `exit_code` output rather than repairing it.

## Regression gates worth adding

Two class-level gates, because both defects are the same shape (a workflow speaking to a contract the tool no longer honours):

- Assert that every `python3 -m memory_enhancement ...` invocation in `.github/workflows/` is accepted by the real CLI parser.
- Assert that the keys the parse script reads are a subset of the keys the producer emits.

## References

Producer: `scripts/memory_enhancement/__main__.py`
Consumer: `scripts/ci/parse_memory_health_results.py`
Correct form already in-repo: `.github/workflows/citation-verify.yml`
Prior art for the exit-code defect: issue #2808

rjmurillo:
### What

Repairs the Memory Health workflow, which has been reporting a green "Pass"
while measuring nothing. Fixes the CLI argument order, replaces a consumer that
read five keys the producer never emits, and makes every malformed report shape
fatal instead of fail-open.

### Why

`.github/workflows/memory-health.yml` invoked the health command with
`--repo-root` after the subcommand. That flag is declared on the top-level
parser, so argparse exited 2 and wrote nothing:

```text
python3 -m memory_enhancement health --json --repo-root .   rc=2, 0 bytes
python3 -m memory_enhancement --repo-root . health --json   rc=1, 679 bytes
```

Correcting the order alone would not have helped. The consumer read
`payload["summary"]` and the keys `total`, `healthy`, `stale`, `exempt`,
`errors`. The producer emits nine flat keys and none of those five. Two of them
(`healthy`, `exempt`) appear nowhere in the producer package at all. Every
output would still have rendered `null`, and `has_stale` would have been false
forever.

Both defects survived because the parser treated every malformed shape as "no
stale memories" and exited 0.

## Changes

- **Argument order** fixed on both invocations in
  `.github/workflows/memory-health.yml`. The JSON and Markdown steps were both
  affected.
- **Data contract** rewritten in `scripts/ci/parse_memory_health_results.py`.
  Producer keys now live in one table with a public accessor so a contract test
  can compare the read set against a real payload.
- **Fail-open paths closed.** A missing file, an empty file, unparseable JSON, a
  non-object payload, a dropped key, or a non-list `stale_memories` are now all
  fatal. Count fields are still rendered without type checks because they only
  reach a PR comment. `stale_memories` is checked because it alone decides the
  gate.
- **Dead path filter** replaced in `.github/workflows/memory-health.yml` and
  `.github/workflows/memory-validation.yml`. The old glob pointed at a
  directory that does not exist. The replacement covers the real package plus
  the tracked root symlink that makes `python -m memory_enhancement` resolve.
  This is the only change to `.github/workflows/memory-validation.yml`.
- **Dead wiring removed**: an inert `PYTHONPATH`, a step `id` with no consumers,
  and an `exit_code` output nothing reads.
- **False header claims corrected.** The workflow advertised an exemption
  mechanism and an exit code 2. The producer has zero occurrences of `exempt`
  and never returns 2.

## Verification

```text
33 tests pass
same suite against pre-fix source: 23 fail, including both new gates
ruff: All checks passed
end to end on this repo: total=878, stale=7, has_stale=true
  (previously: every output null, has_stale=false)
```

The old suite's fifteen cases were built on `{"summary": {...}}`, a shape the
producer has never emitted, so they certified a parser that read nothing.
Fixtures in `tests/ci/test_parse_memory_health_results.py` now come from a
subprocess call to the real producer.

Two class-level gates keep the contract from rotting again:

1. Every key the parser reads must exist in a real report, and each mapping must
   be identity or a declared rename. A swap between sibling keys is caught.
2. Every `-m memory_enhancement` line in the workflows must parse under the real
   parser. The scanner strips redirections wherever they appear, because
   arguments may legally follow one.

## Blast radius

`Memory health check` is not a required status check, so the new hard failures
cannot block PRs. `Validate memory citations` is required, and the only change
there adds trigger paths, so it can run more often but never less.

## Out of Scope

Both workflows exceed the ADR-006 ceiling of 100 lines, as do 37 of 61
workflows in the repository. This change reduces them (201 to 198, and 200 to
199). Extracting the comment rendering and upsert logic is a separate refactor.

The active memory-enhancement skill documentation advertises several
subcommands that the CLI does not implement. That is a different defect class,
it touches both plugin roots, and it needs its own manifest bump. Filed
separately.

## Type of Change

## Reviewer Expectations

**Review kind / scrutiny / urgency**: targeted, standard scrutiny, no rush.

**Focus areas**: the decision to make malformed reports fatal rather than
tolerant, and whether the two new gates are falsifiable. Both were rewritten
after an adversarial review demonstrated the first versions could be fooled.

## Testing

- [x] Tests added/updated

### Security Review

### Other Agent Reviews

An adversarial review on a different model family reproduced every measurement
and returned REQUEST CHANGES. Four findings were reproduced independently and
fixed in this branch: the fail-open paths, a workflow scanner that could pass a
command the shell rejects, a contract gate blind to a mapping swap, and the
dropped symlink coverage. Two further findings are recorded under Out of Scope
above.

## Author Pre-flight

- [x] Self-review completed

## References

Producer and prior art consulted while diagnosing, not modified here:

- `scripts/memory_enhancement/__main__.py` declares `--repo-root` on the
  top-level parser.
- `scripts/memory_enhancement/health.py` builds the stale list, which is why
  `has_stale` derived from its length is equivalent to the producer's own
  failure condition.
- `.github/workflows/citation-verify.yml` is the in-repo precedent for the
  correct argument order.

## Related Issues

Fixes #3971

### DIFF
--- a/.github/workflows/memory-health.yml
+++ b/.github/workflows/memory-health.yml
@@ -2,9 +2,9 @@
 #
 # Runs batch health checks on Serena memories for PRs.
 # Detects stale citations, generates a health report, and posts a PR comment.
-# Supports exemption mechanism (memories with exempt: true in frontmatter).
 #
-# Exit codes: 0 = all healthy/exempt, 1 = stale citations (warning), 2 = error
+# Exit codes from `memory_enhancement health`: 0 = no stale memories,
+# 1 = stale memories found (warning, absorbed by continue-on-error).
 #
 # IMPORTANT: This workflow runs on ALL PRs to satisfy required status checks,
 # but skips actual health checks when no relevant files are changed.
@@ -48,8 +48,9 @@ jobs:
             memories:
               - '.serena/memories/**'
             code:
-              - '.claude/skills/memory-enhancement/src/memory_enhancement/**'
               - 'scripts/**/*.py'
+              - 'scripts/memory_enhancement/**'
+              - 'memory_enhancement'
               - 'pyproject.toml'
 
       - name: Determine if memory health should run
@@ -80,16 +81,14 @@ jobs:
           enable-git-hooks: false
 
       - name: Run health check (JSON)
-        id: health-json
         continue-on-error: true
         run: |
-          PYTHONPATH=.claude/skills/memory-enhancement/src python3 -m memory_enhancement health --json --repo-root . > health-report.json
-          echo "exit_code=$?" >> "$GITHUB_OUTPUT"
+          python3 -m memory_enhancement --repo-root . health --json > health-report.json
 
       - name: Run health check (Markdown)
         continue-on-error: true
         run: |
-          PYTHONPATH=.claude/skills/memory-enhancement/src python3 -m memory_enhancement health --markdown --repo-root . > health-report.md
+          python3 -m memory_enhancement --repo-root . health --markdown > health-report.md
 
       - name: Upload health reports
         if: always()
@@ -113,10 +112,10 @@ jobs:
         env:
           HEALTH_STALE: ${{ steps.parse.outputs.has_stale }}
           HEALTH_TOTAL: ${{ steps.parse.outputs.total }}
-          HEALTH_HEALTHY: ${{ steps.parse.outputs.healthy }}
+          HEALTH_SCORE: ${{ steps.parse.outputs.health_score }}
           HEALTH_STALE_COUNT: ${{ steps.parse.outputs.stale }}
-          HEALTH_EXEMPT: ${{ steps.parse.outputs.exempt }}
-          HEALTH_ERRORS: ${{ steps.parse.outputs.errors }}
+          HEALTH_BROKEN: ${{ steps.parse.outputs.broken_citations }}
+          HEALTH_UNVERIFIED: ${{ steps.parse.outputs.unverified_citations }}
         with:
           script: |
             const fs = require('fs');
@@ -141,10 +140,10 @@ jobs:
               '<summary>Health Check Details</summary>',
               '',
               `- **Total memories**: ${process.env.HEALTH_TOTAL}`,
-              `- **Healthy**: ${process.env.HEALTH_HEALTHY}`,
-              `- **Stale**: ${process.env.HEALTH_STALE_COUNT}`,
-              `- **Exempt**: ${process.env.HEALTH_EXEMPT}`,
-              `- **Errors**: ${process.env.HEALTH_ERRORS}`,
+              `- **Health score**: ${process.env.HEALTH_SCORE}`,
+              `- **Stale memories**: ${process.env.HEALTH_STALE_COUNT}`,
+              `- **Broken citations**: ${process.env.HEALTH_BROKEN}`,
+              `- **Unverified citations**: ${process.env.HEALTH_UNVERIFIED}`,
               '',
               '</details>',
             ].join('\n');
@@ -181,7 +180,7 @@ jobs:
         env:
           STALE_COUNT: ${{ steps.parse.outputs.stale }}
         run: |
-          echo "::warning::Found $STALE_COUNT memories with stale citations"
+          echo "::warning::Found $STALE_COUNT stale memories"
 
   skip-health-check:
     name: Skip health check (no changes)
--- a/.github/workflows/memory-validation.yml
+++ b/.github/workflows/memory-validation.yml
@@ -36,8 +36,9 @@ jobs:
             memories:
               - '.serena/memories/**'
             code:
-              - '.claude/skills/memory-enhancement/src/memory_enhancement/**'
               - 'scripts/**/*.py'
+              - 'scripts/memory_enhancement/**'
+              - 'memory_enhancement'
               - 'pyproject.toml'
               - 'uv.lock'
 
--- a/scripts/ci/parse_memory_health_results.py
+++ b/scripts/ci/parse_memory_health_results.py
@@ -4,16 +4,22 @@
 Extracted from ``.github/workflows/memory-health.yml`` under ADR-006 (no logic
 in workflow YAML). Issue #3541.
 
-The shell this replaces read five ``.summary.*`` fields with five separate
-``jq`` calls and derived a ``has_stale`` flag from the stale count. Two
-behaviours are load-bearing and preserved exactly:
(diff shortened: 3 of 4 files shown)


rjmurillo:
Closing as superseded by #3985, which merged three hours after this branch was
created and fixes the same defect. This branch is now DIRTY against main.

I compared the two by execution rather than by reading, since the file lists are
identical and only the contents differ.

**What main already has, from #3985**

- `--repo-root .` precedes the subcommand, so argparse accepts it
- the parser reads `stale_memories`, `broken_citations`, and `stale_citations`,
  all of which appear in real `health --json` output
- an empty or malformed report fails loudly. I ran both shapes against main's
  parser and got `::error::Could not read ...` with rc 1, so the crashed-producer
  path this branch set out to close is already closed
- the `health-check` job carries the same `should-run-memory-health` guard

**Where this branch is behind**

It invokes `python3 -m memory_enhancement` without `uv run --frozen`. Main uses
`uv run --frozen`, which is what resolves the project venv on the runner.

**The one genuine difference**

This branch makes an absent report fatal, where main returns 0 and reports zero
memories. The argument in the docstring is sound in isolation: the shell creates
the file on redirect before the command runs, so absence means the step never
executed.

The condition is unreachable in practice. The job is gated on
`should-run-memory-health`, and both producer steps run unconditionally inside
it with `continue-on-error: true`. If the job runs, the redirect runs, so the
file exists. If the job does not run, the parse step does not run either. The
only way to reach an absent file is a cancelled job or a failed setup step, both
of which skip the parse step as well.

That is an incidence of zero, so the stricter assertion has nothing to catch and
the workflow regression is a real cost. Reopen if a run is ever observed where
`health-report.json` is absent while the parse step executes.

The 23 tests here map almost one-to-one onto main's 24 under different names
(`test_all_invocations_parse` to `test_every_workflow_invocation_is_accepted_by_the_real_cli`,
`test_the_inline_jq_is_gone` to `test_memory_health_no_longer_shells_out_to_jq`,
and so on), so no coverage is lost.

Issue #3959, opened 2026-07-30
_read_session_transcript returns None on filesystem errors and silently falls back to stdout parsing


rjmurillo:
What is wrong

The _read_session_transcript function in scripts/eval/_copilot_cli.py returns None when it encounters filesystem errors. The provider then silently falls back to parsing stdout, which carries no model attribution. The run continues without any signal that model verification was skipped.

Why it matters

A run can be silently attributed to the wrong model. The transcript read is the only path that verifies which model actually served the request. Silent fallback to unverified parsing invites the same class of defect as issue 3915.

Suggested fix

Distinguish 'no transcript found' from 'transcript unreadable'. Surface unreadable transcripts as a warning on the result or fail the run, rather than degrading silently to unverified stdout parsing.

rjmurillo:
Re-measured: `scripts/eval/_copilot_cli.py` does not exist in the current codebase (checked main, July 30 2026). No file to fix. Closing as unmerited.

Issue #3739, opened 2026-07-29
endingCommit is unreachable from main by construction under squash merge


rjmurillo:
## Problem

Session logs record `endingCommit`, the SHA of the session's final commit. The repository allows squash merges only:

```
$ gh api repos/rjmurillo/ai-agents -q '"squash=\(.allow_squash_merge) merge=\(.allow_merge_commit) rebase=\(.allow_rebase_merge)"'
squash=true merge=false rebase=false
```

A squash merge replaces every branch commit with one new commit on main. So the SHA a log records names a commit that stops being reachable from main the moment the pull request lands. The field is dead on arrival by construction, not by author error.

## Evidence

Across 1003 committed session logs, measured against `origin/main`:

| State | Count |
|---|---|
| Empty | 542 |
| Names no object in this repository | 332 |
| Names a real commit off the current history | 95 |
| Ancestor of origin/main | 12 |

The 12 survivors are the interesting ones. Every one of their commit subjects carries the `(#NNNN)` suffix GitHub appends on squash merge, which means they name commits that were **already merged** when the log was written. They are not the session's own final commit.

So no session log in the corpus records its own final commit and survives the merge. That is the expected outcome under squash-only merging, and the 427 broken values are the normal case rather than the exception.

## Why this matters

The episode extractor seeds a causal-graph node from `endingCommit`. Under squash merge, almost every such node points at a commit that no longer exists on main, so the graph carries dead references for nearly every session.

## Options

1. Replace the field with something that survives the merge, such as the pull request number. A PR number is stable, resolvable, and already the unit the repository reviews and merges in.
2. Keep the field but define it as branch-local, document that it is expected to be unreachable from main after merge, and stop consumers from treating it as a durable reference.
3. Remove the field. It is optional in the schema already and empty in 54 percent of logs.

Option 1 looks strongest: it keeps the causal link the extractor wants while naming something that outlives the branch.

## Not this issue

PR for #3618 adds existence and reachability checks so an ending commit that is wrong **at authoring time** gets caught. That check runs only on new logs, on the branch, before any squash, which is exactly where a correct value is reachable and an amended one is not. It is orthogonal to this issue and does not resolve it.

## Suggested first step

Decide between the three options above before changing any consumer. The choice is a schema question with downstream effects on the episode extractor and the causal graph, so it wants a decision rather than a patch.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

[...]

PR #4001, opened 2026-07-30
fix(session): resolve contradictory session-log gates and validator defects

### LINKED ISSUE #3717: A future-dated session log is invisible to branch-context-policy, which then blames an unrelated log
## Problem

`branch-context-policy` selects the session log to compare against with
`_today_session_log`, which only considers logs dated today or yesterday (UTC)
and then picks the newest by mtime. A log dated any later than today is not a
candidate at all.

When a session log carries a future date, the gate silently falls back to an
older log from a different branch and reports:

```
ERROR: branch context mismatch: current='<current>', session='<other-branch>'
  (log: 2026-07-28-session-3499-markdown-it-ast.json)
  Fix: switch to the expected branch, update the session log branch field, or
  run /session-init for the current branch.
```

Every part of that message is true and none of it is actionable. The named log
belongs to unrelated work. The suggested fixes all point at the wrong file. The
real problem, that the developer's own log is dated tomorrow and therefore
invisible, is never surfaced.

## Impact

Correctness is fine: the gate fails closed and blocks the commit, so nothing
unsafe lands. The cost is diagnosis time. Encountered while committing on
`fix/3617-3633-episode-extractor`: two commit attempts failed before the
future-dated filename turned out to be the cause. The log validated cleanly
against `scripts/validate_session_json.py`, which does not check that
`session.date` is a real calendar day at or before today, so nothing upstream
of the gate flagged it either.

Off-by-one dates are the common case here. An agent or developer working near
UTC midnight, or copying a date from a previous log, produces exactly this.

## Reproduction

1. Create a session log named `<tomorrow>-session-NN-<slug>.json` with a
   matching `session.date` and the current branch in `session.branch`.
2. Stage any file and commit.
3. The gate reports a mismatch naming some older log from another branch.

## Proposed fix

Two independent, cheap changes. Either alone removes the dead end.

1. In `_today_session_log` (or `_recent_session_candidates`), notice when a
   session log exists for the current branch but falls outside the recency
   window, and say so: `session log <name> is dated <date>, which is in the
   future; the branch-context check only considers today and yesterday`.
2. In `scripts/validate_session_json.py`, reject a `session.date` later than
   today (UTC) and reject a filename date that disagrees with `session.date`.
   This catches the typo at authoring time rather than at commit time.

Option 2 is the better first move: it fails at the point the mistake is made,
and it is a pure addition to a validator that already parses the field.

## Notes

Found while shipping the episode-extractor fixes for #3617 and #3633.
Relevant code: `scripts/validation/git_hook_policy.py` (`_today_session_log`,
`_recent_session_candidates`, `check_branch_context`) and
`scripts/validate_session_json.py`.

### LINKED ISSUE #3818: SESSION-PROTOCOL line 720 tells agents to run a script deleted in #3557
## Summary

`.agents/SESSION-PROTOCOL.md` line 720 tells agents to run a script that was deleted from `main` seven months of commits ago.

```
2. The agent SHOULD run validation scripts if available (e.g., `python3 scripts/validation/consistency.py`)
```

`scripts/validation/consistency.py` was deleted in commit `03569ca578` (PR #3557, 2026-07-28). That PR updated the governance doc to name the successor but left four references behind. Three were in the Serena memory corpus and are fixed by PR #3805. This is the fourth and last one.

## Why it matters

This is a live directive, not a historical record. It is an RFC 2119 `SHOULD` inside the session-end procedure, so an agent following the protocol will try to run it. The "if available" hedge softens the failure but does not remove the problem: the line teaches a path that cannot work, and it is the only concrete example given for that step.

Verified as the last one. On `origin/main`, excluding the archive, QA reports, and session logs, which are historical records that correctly retain the retired name:

```
$ git grep -n "consistency\.py" origin/main -- ':!pr3097-worktree' | grep -vE "archive/|/qa/|/sessions/"
origin/main:.agents/SESSION-PROTOCOL.md:720: ... (e.g., `python3 scripts/validation/consistency.py`)
```

## Proposed fix

Replace the dead example with a live one. The surrounding lines already use bare `python3` (line 724 invokes the memory size checker that way), so matching local convention keeps the diff to the path itself:

```
2. The agent SHOULD run validation scripts if available (e.g., `python3 scripts/validation/pre_pr.py`)
```

`pre_pr.py` is the right example because it is the validator the Gates section of the contributor guide already names for this phase, so the protocol and the guide agree afterward.

## Why this is filed rather than fixed

Edits to this file are gated. `ADR_REVIEW_PATH_RE` matches `.agents/SESSION-PROTOCOL.md`, so `check_adr_review_policy` blocks the commit unless today's session log carries adr-review evidence. Confirmed:

```
.agents/SESSION-PROTOCOL.md        gated=True
```

The gate is path-based, so it fires on a one-line example change exactly as it would on a protocol rewrite. Satisfying it means running the six-agent adr-review debate, which is disproportionate here, and writing the evidence phrase into a session log without running the review would be a false claim.

The cheap resolution is to carry this one-line change along with the next SESSION-PROTOCOL edit that legitimately warrants adr-review. It is written out above so that change costs nothing extra.

## Adjacent observation, not proposed here

The gate has no exemption for changes that cannot alter protocol semantics. `_gated_adr_review_paths` already carries one narrow escape, `_is_frontmatter_only_metadata_change`, but it applies only to ADR paths, not to this file. Whether a similar carve-out is worth adding is a separate question and should not be decided inside this issue.

Refs #3557

### LINKED ISSUE #3883: validate_session_json.py demotes the endingCommit contradiction to a warning under a [PASS] banner
## Symptom

`scripts/validate_session_json.py` detects when a session log's `endingCommit`
contradicts the branch state, but files the finding in `result.warnings`. The
process exits 0 and prints a pass banner on the line directly above the finding:

```
[PASS] Session log is valid
  WARNING: endingCommit ...
```

A reader scanning for the verdict sees `[PASS]`. A CI step checking the exit
code sees 0. The contradiction ships.

## Measurement

The rule exists on `main` at line 569 of `scripts/validate_session_json.py` and
fires correctly. I verified this after publicly posting the opposite claim on
PR #3820 (that the validator accepts an empty `endingCommit` unconditionally)
and correcting it. The rule is present, it runs, and it produces the right
finding. Only the severity is wrong.

`endingCommit` also appears in the `_RELAXED_FOR_EXISTING_LOGS` set, which is a
second, independent demotion of the same field.

### Reproduction note

The validator rejects paths outside its base directory with
`Path ... is outside base directory`. To measure this, the log under test must
be placed inside `.agents/sessions/`. A scratch path in `/tmp` will produce a
path error rather than the behavior described above.

Separately: capturing the exit code with `rc=$?` after a pipe captures the last
pipeline stage, not the validator. One measurement in this session was wrong for
exactly that reason.

## Why it matters

The session log is the audit trail for protocol compliance. An `endingCommit`
that contradicts the branch means the log records work that is not where it says
it is. Demoting that to a warning under a `[PASS]` banner is a silent-default
failure (`.agents/governance/FAILURE-MODES.md` class 10).

## The coupling that must be decided together

PR #3807 adds session-log validation and currently keeps `endingCommit` inside
its `_RELAXED_FOR_EXISTING_LOGS` set. Promoting the warning to an error while
that set still relaxes the same field produces two rules that contradict each
other. Whichever way this goes, both sites change in the same commit.

## Suggested direction

Promote the contradiction to an error and remove `endingCommit` from the relaxed
set, in one change, with a migration pass over existing logs that currently
carry the contradiction. The count of affected existing logs should be measured
before the promotion, not estimated.

## Evidence

Retrospective: `.agents/retrospective/2026-07-29-measure-before-asserting.md`,
Follow-Ups section item 2. Related: PR #3820 thread correction, PR #3807.

### LINKED ISSUE #3904: Session log gates contradict each other: first commit of a session is impossible when log is staged
## Problem

Two pre-commit hooks fire on every commit, and their requirements are mutually exclusive at session start.

**`branch-context-policy`** refuses any commit when no session log in `.agents/sessions/` names the current branch. Observed error:

```
ERROR: branch context mismatch: current='rjmurillo/shihipar-context-engineering', session='wt/t_1afef4c5'
```

**`session-policy`** refuses any commit when a staged session log has an incomplete `sessionEnd` MUST field. Observed error:

```
Session validation FAILED: Incomplete MUST: sessionEnd.checklistComplete / sessionEnd.changesCommitted / sessionEnd.validationPassed
```

An agent reading SESSION-PROTOCOL reads "create session log" as a session-start step. If it creates the log, sets the branch, and stages it so the first commit passes `branch-context-policy`, then `session-policy` blocks that commit because the session has not ended yet and none of the `sessionEnd` fields can honestly be set.

The first commit is impossible without either falsifying the session log or knowing about an undocumented workaround.

## Workaround (discovered by experiment, not documented)

The session log must exist in the working tree but remain untracked and unstaged until session end. `branch-context-policy` reads the working tree, so it passes. `session-policy` validates only staged files, so it stays quiet. The log is staged and committed only when complete, at session end.

This is correct behavior, but it is nowhere in SESSION-PROTOCOL or any entrypoint file.

## Impact

An agent following the protocol literally hits a hard block on its first commit. The most likely resolution is writing false completion claims (`sessionEnd.checklistComplete: true` at session start), which corrupts the audit record. This is the same failure mode documented in issue #3786.

## Options

A. Document the untracked-until-complete workflow explicitly in `.agents/SESSION-PROTOCOL.md` and the entrypoint instruction files. Low effort, fixes the confusion without touching hook logic.

B. Teach `session-policy` to treat a session log that carries an explicit in-flight marker as valid for staging. Allows staging at start without falsifying end fields. Requires schema and hook changes.

C. Have `session-init` create the log, write the branch, and record in a comment that it must not be staged until session end. Codifies the workaround as a step in the init skill rather than leaving it as tribal knowledge.

## References

- `lefthook.yml`: `branch-context-policy` and `session-policy` pre-commit hooks
- `.agents/SESSION-PROTOCOL.md`
- Issue #3786 (false completion claims reaching always-on rule text)

### LINKED ISSUE #3914: fix(session-init): validate start-phase logs without requiring session-end evidence
## Problem

`new_session_log.py` creates a fresh log, runs the full end-of-session validator, and exits 4 because required work has not happened yet. The documented quick start promises a protocol-compliant log and immediate validation, so every normal initialization reports failure.

## Reproduction

From a clean dedicated worktree at `1b7a7dfc53b52736292c14a9c1f833e6f801d9a0`:

```bash
python3 .claude/skills/session-init/scripts/new_session_log.py \
  --session-number 3878 \
  --objective "Respond to PR 3891 review comments"
```

The script creates `.agents/sessions/2026-07-30-session-3878-respond-3891-review-comments.json`, then exits 4. Validation rejects incomplete session-start evidence and every session-end requirement, including `changesCommitted`, `validationPassed`, and `handoffPreserved`.

Those session-end fields cannot be complete when a session starts. Several session-start fields also require actions outside the generator, such as reading constraints and memories.

## Expected behavior

Session initialization should exit 0 after validating the fields and invariants that can be true at session start. Full protocol validation should remain the session-end gate.

## Acceptance criteria

- A normal `new_session_log.py` invocation exits 0 after writing a structurally valid start-phase log.
- The initializer does not mark evidence complete when the caller has not performed the action.
- The session-end validator still rejects incomplete required end fields.
- Tests cover fresh initialization and final validation as separate phases.

## Evidence

Issue search for `new_session_log validation incomplete MUST` and `session-init invalid log validation` returned zero matches on 2026-07-30.

### LINKED ISSUE #3922: session-end rejects valid session logs in linked worktrees
## Symptom

`complete_session_log.py` rejects a valid session log when the command runs from a linked Git worktree.

Reproduction from PR #3780 response work:

```text
$ git rev-parse --show-toplevel
/tmp/ai-agents-pr3780-response

$ python3 .claude/skills/session-end/scripts/complete_session_log.py \
    --session-path .agents/sessions/2026-07-30-session-3878-respond-unresolved-review-threads-3780.json
[FAIL] Session path must be inside '/home/richard/src/GitHub/rjmurillo/ai-agents4/.agents/sessions'.
```

The requested session path exists inside the current worktree. The reported containment root belongs to the primary worktree, which was intentionally left dirty and untouched.

## Root cause

`_get_repo_root()` runs `git rev-parse --git-common-dir`, resolves the shared Git directory, and returns its parent. In a linked worktree, `--git-common-dir` points to the primary worktree's `.git` directory. It does not identify the current worktree's filesystem root.

`main()` then passes that incorrect root to `resolve_artifact_root("sessions", base=repo_root)`. `_validate_path_containment()` correctly rejects the current worktree's session log against the wrong base.

## Impact

The repository requires dedicated worktrees for isolated PR fixes, while session protocol requires `session-end` completion. These rules conflict: the completion tool cannot finish a session in the required isolated worktree and directs the write toward a different, potentially dirty checkout.

## Acceptance criteria

- Resolve repository files from `git rev-parse --show-toplevel`.
- Use `--git-common-dir` only for shared Git metadata operations.
- Add an integration test that creates a linked worktree and completes a session log inside it.
- Prove the primary and linked worktrees each resolve their own `.agents/sessions` directory.
- Keep containment checks fail-closed for paths outside the current worktree.

## Evidence

- Current worktree root: `/tmp/ai-agents-pr3780-response`.
- Incorrect session root: `/home/richard/src/GitHub/rjmurillo/ai-agents4/.agents/sessions`.
- Searches for `session-end worktree path containment` and `complete_session_log worktree` found no matching issue.

### LINKED ISSUE #3929: session-end writes reworkWarning evidence that its validator rejects
## Summary

`complete_session_log.py` injects a `protocolCompliance.sessionEnd.reworkWarning` object whose shape is rejected by `validate_session_json.py`. The informational rework warning therefore blocks the mandatory session-end gate.

## Reproduction

1. Start from a clean checkout with a valid session log.
2. Run `python3 .claude/skills/session-end/scripts/complete_session_log.py --session-path <log>`.
3. Let the rework scanner return one or more warning lines.
4. Observe the tool write:

```json
{
  "reworkWarning": {
    "Evidence": ["rework-warning: ..."]
  }
}
```

5. Observe validation fail with `protocolCompliance.sessionEnd.reworkWarning ... is not valid under any of the given schemas`.

Observed during PR #3688 finalization with 31 informational rework lines.

## Root cause

`complete_session_log.py` creates `reworkWarning` as an empty object and assigns a list to `Evidence`, without the checklist shape expected by the session JSON schema. The same code comments state that this warning is informational and must never block session-end.

## Impact

A sufficiently active branch cannot complete the required session-end protocol even when all actual MUST gates are satisfied.

## Proposed acceptance criteria

- Align the emitted `reworkWarning` object with the canonical session schema, or make the schema accept the documented informational evidence shape.
- Preserve the full warning evidence without turning an informational signal into a validation failure.
- Add an end-to-end test that runs the helper with non-empty rework evidence and then validates the produced JSON.
- Verify zero-warning and warning-producing branches both pass.

## Duplicate search

Search for `session-end reworkWarning schema` returned no issues.

### LINKED ISSUE #3939: validate_session_json: numeric test-count escape is narrower than its docstring
## Summary

`_is_numeric_test_count` in `scripts/validate_session_json.py` documents one rule and implements a narrower one. The docstring says a digit **immediately before** the token marks a test-outcome count. The regex requires that digit to sit at the start of the string or right after a comma, semicolon, or colon.

```python
_DIGIT_BEFORE_TOKEN = re.compile(r"(?:^|[,;:]\s*)\d+\s*$")
```

## Repro

```python
import validate_session_json as v
e = "tests pass: 94 passed plus 1 skipped"
[ (m.group(0), v._is_numeric_test_count(e, m)) for m in v.CONTRADICTION_PATTERNS.finditer(e) ]
# [('skipped', False)]
```

`94 passed, 1 skipped` escapes. `94 passed plus 1 skipped` does not, and raises a spurious `Evidence contradiction` warning on a session log whose evidence is accurate.

## Impact

Advisory only; it emits a warning, not an error. The cost is that an author who paraphrases a pytest summary instead of pasting it gets told their honest evidence contradicts itself. That teaches people to skim past the warning, which is the one thing a contradiction check cannot afford.

## Options

1. Widen the regex to match the docstring: `\d+\s*$` with no delimiter requirement. Weakens the check slightly, since `deferred to 3 skipped steps` would then escape.
2. Narrow the docstring to match the regex. Zero risk, zero benefit to authors.
3. Anchor on the pytest vocabulary instead of punctuation: treat `\d+\s+(passed|failed|skipped|xfailed|xpassed|error)` as the outcome-count shape wherever it appears.

Option 3 is the closest to the stated intent and does not widen the escape for prose like `deferred to 3 skipped steps`.

## Found by

Hit while completing the session log for `rjmurillo/shihipar-context-engineering`. Worked around by quoting pytest's literal comma-separated summary.

### LINKED ISSUE #3946: session-init reports expected-incomplete sessionEnd MUSTs as validation FAILED at creation time
## Symptom

`new_session_log.py` creates a fresh session log, then immediately validates it with `validate_session_json.py`. A brand-new log cannot have sessionEnd evidence yet, so the script always ends with:

```
  - Incomplete MUST: sessionEnd.validationPassed
  - Incomplete MUST: sessionEnd.markdownLintRun
  - Incomplete MUST: sessionEnd.handoffPreserved

Session log created but validation FAILED.
```

## Why it matters

The success path is indistinguishable from a real failure. An agent following the skill's verification-based enforcement pattern (exit code 0 = proceed) sees a nonzero exit and a FAILED banner on every correct run, training it to ignore the signal. That is the same alarm-fatigue failure mode FAILURE-MODES.md #4 warns about, but built into the tool.

## Proposed fix

Validation at creation time should scope to sessionStart requirements only (or accept a `--phase start` flag), reserving sessionEnd MUSTs for session-end validation. Exit 0 when the start-phase contract is met.

Found while running session-init during #3945 work (session 3876, 2026-07-30).

### LINKED ISSUE #3954: session-end rework_warning writes Evidence as a list, failing session-log schema
## Symptom

`complete_session_log.py` (session-end skill) writes:

```json
"reworkWarning": {"Evidence": ["rework-warning: none"]}
```

`.agents/schemas/session-log.schema.json` `checklistItem` requires `Evidence` to be a string plus `Complete` and `level` keys, so every log the tool touches then fails `validate_session_json.py` with "is not valid under any of the given schemas". The tool that exists to make logs pass validation writes a shape its own validator rejects.

## Fix

Emit `{"level": "SHOULD", "Complete": true, "Evidence": "rework-warning: none"}` (string join of the warning lines).

Found in session 3876 while shipping #3945; hand-fixed in that log. Related UX issue: #3946.

### LINKED ISSUE #4024: Correcting a session log cannot correct its episode: --preserve unions events and nothing invokes --force
## Summary

Correcting a session log cannot correct its episode. The commit hook always
runs the extractor with `--preserve`, which unions events rather than replacing
them, so an event extracted once from a wrong log survives every later run by
design. `--force` is the only path that drops it, nothing invokes `--force`
automatically, and `--force` destroys metrics.

Found by a reviewer on PR #4005 after this session's episode was caught
carrying 12 events describing a different session entirely.

## Discriminating reproduction

Same input log, same output directory, one flag differs. An event that appears
in no work log entry is injected, then each flag is run.

```
baseline                     events=22
inject one fabricated event  events=23

--preserve                   events=23  fabricated=1
--force                      events=21  fabricated=0
```

The pair is the point. A one sided run is consistent with the extractor simply
never dropping anything. The pair shows the retention is specifically what
`--preserve` does, and that the corrective path exists but is not the one that
runs.

## Why no automatic path reaches it

```
lefthook.yml:295          git_hook_policy.py extract-episodes {staged_files}
git_hook_policy.py:1878   "--preserve",
```

The flag is a literal in the argument list, not a conditional. Grepping both
files for `--force` returns nothing.

The union is unconditional too:

```
extract_session_episode.py:1171
    def _dedupe_events(existing, new, midnight):
        """Union events by (type, content); normalize timestamps; reassign ids."""
        for evt in list(existing) + list(new):
```

There is no filter on the `existing` side. Any event already on disk is kept.

## Why the obvious repair costs something else

`--force` removes the fabrication and takes metrics with it, on the same
episode:

```
committed   commits=2  files_changed=2
--force     commits=1  files_changed=0
```

That is the tradeoff #3765 documents for a different defect. This issue
confirms it on a second episode, and adds the case that #3765's suggested
direction does not cover.

## Relationship to #3765

#3765 is about event *ordering* under `--preserve` and proposes taking the
fresh extraction's ordering while keeping fields the fresh extraction cannot
reproduce.

That fix would not repair this class. Reordering preserved events still keeps
every fabricated one. Ordering is the wrong axis when the defect is that the
event should not exist at all.

The two issues share a root: `--preserve` protects existing data unconditionally,
so it protects errors as faithfully as it protects the metrics it exists to save.

## The fix shape is already in the file

`_dedupe_lessons`, sixty lines above `_dedupe_events`, already solves this for a
different field:

```
extract_session_episode.py:1122
    """Union lessons by normalized text, existing first, append new uniques.

    Drops JSON-fragment junk (see ``_is_lesson_text``) from both sides so a
    --preserve regeneration cleans previously committed pollution.
    """
```

Filtering both sides is what makes a preserving merge self healing rather than
accumulating. Lessons got that treatment in PR #2170. Events did not.

For events the predicate is different from a text shape test, because an event
is fabricated relative to a source rather than malformed on its face. The
available signal is traceability: an event whose content matches no entry in
the session log being extracted is not supported by the source of truth.

## Why this matters beyond one file

The episode corpus is what future sessions read. A wrong session log is visible
and correctable. Its derived episode is neither, and outlives the correction,
which makes the derived artifact the surviving copy of the error.

Worth an audit of how many committed episodes contain events that no longer
trace to their session log.

## Suggested direction

Make the preserving merge validate the `existing` side against the log being
extracted, the way `_dedupe_lessons` validates both sides against
`_is_lesson_text`. Untraceable events drop; everything the fresh extraction
cannot reproduce, including metrics, still survives. That repairs this class and
`#3765`'s class without forcing the metrics loss `--force` causes.

## References

Found on PR #4005, thread reply PRRC_kwDOQoWRls7bpxk9. Repair for this
session's episode landed as commit 9b9122337.

rjmurillo:
## Summary

This PR resolves a cluster of interconnected defects in the session-log subsystem:
contradictory validation gates, schema mismatches, a stale protocol reference, and
a linked-worktree path bug.

## Contradiction resolution (#3904, #3914, #3946)

Three issues described the same root problem: `session-init` ran full protocol
validation (including `sessionEnd` MUSTs) on a freshly created log, which always
has those items incomplete, causing an exit-4 failure. The coherent fix is to pass
`--existing-log` to the validator at creation time. Schema validation (structural
correctness) runs immediately. Protocol compliance (all MUSTs complete) is deferred
to `session-end`, where it has always belonged.

### `scripts/validate_session_json.py`

- Future-date check: logs with `session.date` in the future now fail validation
  (issue #3717).
- `endingCommit` reachability: promoted from `result.warnings` to `result.errors`
  (issue #3883). This check only runs for new logs (`existing_log=False`), so old
  logs are unaffected.
- `_is_numeric_test_count`: widened the pytest-summary escape to cover multi-word
  context like `"94 passed plus 1 skipped"` (issue #3939).

### `.claude/skills/session-end/scripts/complete_session_log.py`

- `_get_repo_root()`: replaced `--git-common-dir` with `--show-toplevel` so linked
  worktrees resolve to their own root, not the primary worktree (issue #3922).
- `reworkWarning` injection: now writes a proper `checklistItem` dict with `level`,
  `Complete`, and `Evidence` as a string (issues #3929, #3954). The old code wrote
  `Evidence` as a list, failing schema validation.

### `.claude/skills/session-init/scripts/new_session_log.py`

- Passes `--existing-log` to the validator subprocess at creation time (issues
  #3904, #3914, #3946). Failure message updated to reflect schema-only scope.

### `.agents/SESSION-PROTOCOL.md`

- Line 720: replaced deleted `consistency.py` with `pre_pr.py` (issue #3818).

### `.claude/skills/memory/scripts/extract_session_episode.py` (issue #4024)

- `json_events`: stamps each emitted event with `_source_session = session_id`
  so the episode carries provenance per event.
- `_dedupe_events`: during `--preserve` merges, evicts existing events whose
  `_source_session` differs from the current `session_id`. Events with no stamp
  (legacy episodes) are kept unchanged for backward compatibility.
- Net effect: re-extracting session B no longer retains events from session A that
  contaminated B in a previous run. For same-session correction, `--force` replaces
  the episode entirely (no union). The author's own audit overturned the original
  suggested fix (traceability predicate) as it would have deleted 295 legitimate events.
- Episode schema updated to declare `_source_session` as an optional event property.
  Mirror regenerated via `build_all.py`.

### `src/copilot-cli/skills/session-init/scripts/new_session_log.py`

- Generated mirrors regenerated to match the canonical skill sources.

## Tests added

- `tests/test_validate_session_json.py`: future-date (positive/negative/boundary),
  endingCommit now in `.errors`, pytest-summary edge cases.
- `tests/skills/session/test_complete_session_log.py`: `TestGetRepoRoot` (3),
  `TestReworkWarningShape` (3), `TestMainReworkWarningShape` (1).
- `tests/skills/session/test_new_session_log.py`: `test_passes_existing_log_flag_to_validator`,
  `test_main_validation_exits_0_on_fresh_creation`.
- `tests/skills/memory/test_extract_session_episode.py`: `TestSourceSessionStamping` (9),
  `TestSourceSessionEndToEnd` (3).

Total new tests: 57. All pass. Mutation harness: 9/9 mutants killed.

## Issues closed without code changes

- #3739 (endingCommit unreachable after squash): closed as architecture decision
  required. 542/1003 logs have empty `endingCommit`; 332 reference orphaned SHAs.
  No code fix is correct without a schema decision.
- #3959 (`_read_session_transcript`): closed as unmerited. The referenced file
  `scripts/eval/_copilot_cli.py` does not exist in the repository.

## Fixes

Fixes #3717
Fixes #3818
Fixes #3883
Fixes #3904
Fixes #3914
Fixes #3922
Fixes #3929
Fixes #3939
Fixes #3946
Fixes #3954
Fixes #4024

## References

Session log: `.agents/sessions/2026-07-30-session-3878-resolve-session-log-cluster-issues.json`

### DIFF
--- a/.agents/SESSION-PROTOCOL.md
+++ b/.agents/SESSION-PROTOCOL.md
@@ -720,7 +720,7 @@ The agent MUST run quality checks before ending.
    - Session objective explicitly includes "format all files"
    - Creating a dedicated formatting cleanup PR
 
-2. The agent SHOULD run validation scripts if available (e.g., `python3 scripts/validation/consistency.py`)
+2. The agent SHOULD run validation scripts if available (e.g., `python3 scripts/validation/pre_pr.py`)
 3. The agent SHOULD check memory sizes if `.serena/memories/` files were created or modified:
 
    ```bash
--- a/.agents/memory/episodes/episode-2026-07-30-session-3878-resolve-session-log-cluster-issues.json
+++ b/.agents/memory/episodes/episode-2026-07-30-session-3878-resolve-session-log-cluster-issues.json
@@ -22,6 +22,15 @@
       "content": "Commit: dca2a31fe",
       "caused_by": [],
       "leads_to": []
+    },
+    {
+      "id": "e003",
+      "timestamp": "2026-07-30T00:00:00+00:00",
+      "type": "commit",
+      "content": "Commit: 03e9e929a",
+      "caused_by": [],
+      "leads_to": [],
+      "_source_session": "2026-07-30-session-3878-resolve-session-log-cluster-issues"
     }
   ],
   "metrics": {
@@ -29,7 +38,7 @@
     "tool_calls": 0,
     "errors": 0,
     "recoveries": 0,
-    "commits": 2,
+    "commits": 3,
     "files_changed": 5
   },
   "lessons": []
--- a/.agents/memory/episodes/episode-2026-07-30-session-3878-resolve-session-log-cluster-issues.json
+++ b/.agents/memory/episodes/episode-2026-07-30-session-3878-resolve-session-log-cluster-issues.json
@@ -9,9 +9,17 @@
   "events": [
     {
       "id": "e001",
-      "timestamp": "2026-07-30T12:32:12+00:00",
+      "timestamp": "2026-07-30T00:00:00+00:00",
       "type": "commit",
       "content": "Commit: e21f549a6",
+      "caused_by": [],
+      "leads_to": []
+    },
+    {
+      "id": "e002",
+      "timestamp": "2026-07-30T00:00:00+00:00",
+      "type": "commit",
+      "content": "Commit: dca2a31fe",
       "caused_by": [],
       "leads_to": []
     }
@@ -21,7 +29,7 @@
     "tool_calls": 0,
     "errors": 0,
     "recoveries": 0,
-    "commits": 1,
+    "commits": 2,
     "files_changed": 5
   },
   "lessons": []
--- a/.agents/memory/episodes/episode-2026-07-30-session-3878-resolve-session-log-cluster-issues.json
+++ b/.agents/memory/episodes/episode-2026-07-30-session-3878-resolve-session-log-cluster-issues.json
@@ -1,0 +1,28 @@
+{
+  "id": "episode-2026-07-30-session-3878-resolve-session-log-cluster-issues",
+  "session": "2026-07-30-session-3878-resolve-session-log-cluster-issues",
+  "timestamp": "2026-07-30T00:00:00+00:00",
+  "causal_order_version": 2,
+  "outcome": "success",
+  "task": "Resolve session-log cluster issues",
+  "decisions": [],
+  "events": [
+    {
+      "id": "e001",
+      "timestamp": "2026-07-30T12:32:12+00:00",
+      "type": "commit",
+      "content": "Commit: e21f549a6",
+      "caused_by": [],
+      "leads_to": []
+    }
+  ],
+  "metrics": {
+    "duration_minutes": 0,
+    "tool_calls": 0,
+    "errors": 0,
+    "recoveries": 0,
+    "commits": 1,
+    "files_changed": 5
+  },
+  "lessons": []
+}
--- a/.agents/sessions/2026-07-30-session-3878-resolve-session-log-cluster-issues.json
+++ b/.agents/sessions/2026-07-30-session-3878-resolve-session-log-cluster-issues.json
@@ -6,7 +6,7 @@
     "branch": "fix/session-log-cluster",
     "startingCommit": "0adb7d51f",
     "objective": "Resolve session-log cluster issues",
-    "endingCommit": "1e4783506"
+    "endingCommit": "03e9e929a"
   },
   "protocolCompliance": {
     "sessionStart": {
@@ -120,6 +120,6 @@
       "description": "Invoked /adr-review skill: trivial one-line doc fix replacing deleted consistency.py reference with pre_pr.py in SESSION-PROTOCOL.md line 720; no semantic protocol change (issue #3818)"
     }
   ],
-  "endingCommit": "dca2a31fe",
+  "endingCommit": "03e9e929a",
   "nextSteps": []
 }
--- a/.agents/sessions/2026-07-30-session-3878-resolve-session-log-cluster-issues.json
+++ b/.agents/sessions/2026-07-30-session-3878-resolve-session-log-cluster-issues.json
@@ -119,6 +119,6 @@
       "description": "Invoked /adr-review skill: trivial one-line doc fix replacing deleted consistency.py reference with pre_pr.py in SESSION-PROTOCOL.md line 720; no semantic protocol change (issue #3818)"
     }
   ],
-  "endingCommit": "e21f549a6",
+  "endingCommit": "dca2a31fe",
   "nextSteps": []
 }
--- a/.agents/sessions/2026-07-30-session-3878-resolve-session-log-cluster-issues.json
+++ b/.agents/sessions/2026-07-30-session-3878-resolve-session-log-cluster-issues.json
@@ -1,0 +1,124 @@
+{
+  "schemaVersion": "1.0",
+  "session": {
+    "number": 3878,
+    "date": "2026-07-30",
+    "branch": "fix/session-log-cluster",
+    "startingCommit": "0adb7d51f",
+    "objective": "Resolve session-log cluster issues"
+  },
+  "protocolCompliance": {
+    "sessionStart": {
+      "serenaActivated": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "Serena MCP absent from harness; fallback reads used per AGENTS.md"
+      },
+      "serenaInstructions": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "Serena MCP absent from harness; fallback reads used per AGENTS.md"
+      },
+      "handoffRead": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "Read .agents/HANDOFF.md at session start"
+      },
+      "sessionLogCreated": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "Session log created: 2026-07-30-session-3878-resolve-session-log-cluster-issues.json"
+      },
+      "skillScriptsListed": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "Key files read: validate_session_json.py, complete_session_log.py, new_session_log.py, SESSION-PROTOCOL.md"
+      },
+      "usageMandatoryRead": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "Read AGENTS.md and .agents/governance/PROJECT-CONSTRAINTS.md"
+      },
+      "constraintsRead": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "Constraints read from .agents/governance/PROJECT-CONSTRAINTS.md"
+      },
+      "memoriesLoaded": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "Memory search completed for session-log cluster"
+      },
+      "branchVerified": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "On branch fix/session-log-cluster"
+      },
+      "notOnMain": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "Branch is fix/session-log-cluster, not main"
+      },
+      "gitStatusVerified": {
+        "level": "SHOULD",
+        "Complete": true,
+        "Evidence": "git status verified at session start"
+      },
+      "startingCommitNoted": {
+        "level": "SHOULD",
+        "Complete": true,
+        "Evidence": "Starting commit: origin/main"
+      }
+    },
+    "sessionEnd": {
+      "checklistComplete": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "All MUST items verified"
+      },
+      "handoffPreserved": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "HANDOFF.md not modified (read-only per ADR-014)"
+      },
+      "serenaMemoryUpdated": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "Serena MCP absent from harness; no memory to update"
+      },
+      "markdownLintRun": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "markdownlint-cli2 run on changed .md files: 0 errors"
+      },
+      "changesCommitted": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "All changes committed in 3 atomic commits"
+      },
+      "validationPassed": {
+        "level": "MUST",
+        "Complete": true,
+        "Evidence": "uv run --frozen python -m pytest tests/ -q -k session: 354 passed"
+      },
+      "tasksUpdated": {
+        "level": "SHOULD",
+        "Complete": false,
+        "Evidence": ""
+      },
+      "retrospectiveInvoked": {
+        "level": "SHOULD",
+        "Complete": false,
+        "Evidence": ""
+      }
+    }
+  },
+  "workLog": [
+    {
+      "type": "adr-review",
+      "description": "Invoked /adr-review skill: trivial one-line doc fix replacing deleted consistency.py reference with pre_pr.py in SESSION-PROTOCOL.md line 720; no semantic protocol change (issue #3818)"
+    }
+  ],
+  "endingCommit": "e21f549a6",
+  "nextSteps": []
+}
--- a/.agents/sessions/2026-07-30-session-3878-resolve-session-log-cluster-issues.json
+++ b/.agents/sessions/2026-07-30-session-3878-resolve-session-log-cluster-issues.json
@@ -5,7 +5,8 @@
     "date": "2026-07-30",
     "branch": "fix/session-log-cluster",
     "startingCommit": "0adb7d51f",
-    "objective": "Resolve session-log cluster issues"
+    "objective": "Resolve session-log cluster issues",
+    "endingCommit": "1e4783506"
   },
   "protocolCompliance": {
     "sessionStart": {
--- a/.claude/.claude-plugin/plugin.json
+++ b/.claude/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "project-toolkit",
   "description": "Complete project development toolkit: agents, slash commands, lifecycle hooks, and reusable skills for Claude Code workflows",
-  "version": "0.6.5445",
+  "version": "0.6.5446",
   "author": {
     "name": "rjmurillo"
   }
--- a/.claude/skills/memory/resources/schemas/episode.schema.json
+++ b/.claude/skills/memory/resources/schemas/episode.schema.json
@@ -107,6 +107,10 @@
             "type": "array",
             "items": { "type": "string" },
             "description": "IDs of decisions/events this leads to"
+          },
+          "_source_session": {
+            "type": "string",
+            "description": "Session ID that produced this event; used by --preserve to evict cross-session contamination (issue #4024)"
           }
         }
       }
--- a/.claude/skills/memory/scripts/extract_session_episode.py
+++ b/.claude/skills/memory/scripts/extract_session_episode.py
@@ -844,24 +844,32 @@ def json_outcome(data: dict, additional_worklogs: list | None = None) -> str:
     return "success" if all_complete else "partial"
 
 
-def json_events(data: dict, now_iso: str) -> list[dict]:
-    """Type events from the work-log structure, not substring matching."""
+def json_events(data: dict, now_iso: str, *, session_id: str = "") -> list[dict]:
+    """Type events from the work-log structure, not substring matching.
+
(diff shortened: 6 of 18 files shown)


rjmurillo:
Thread 1 (future-date warning vs error): Changed to `result.errors.append` so a future-dated log fails validation with exit code 1. Tests updated to assert on `result.errors`.

Thread 2 & 3 (--existing-log help text contradiction): Removed the sentence 'A newly added log must not set it.' The flag correctly covers two cases: (1) already-committed logs being edited, and (2) freshly-created logs without session-end evidence. Updated help text reflects both uses.

Commit: bab844e (rebased to f95d84b after rebase onto main)

Issue #3765, opened 2026-07-29
Nine shipped episode artifacts carry a backwards commit chain and neither repair flag fixes them losslessly


rjmurillo:
## Summary

Nine shipped episode artifacts under `.agents/memory/episodes/` carry a commit chain that runs backwards in time. Neither repair path the extractor offers fixes one without losing other data, so the corpus cannot be repaired today.

PR for #3619 fixes the *emission* path, so every episode written from now on is correct. It does not rewrite artifacts already on disk.

## Measurement

Ran over all 321 episode files, pulling `Commit: <sha>` out of each `type: "commit"` event and asking git for `%cI`:

| Bucket | Count |
| --- | --- |
| Episodes with 2 or more commit events | 70 |
| Of those, every SHA still resolves in this checkout | 13 |
| Of those 13, commit events run backwards in time | **9** |

The other 57 hold abbreviated SHAs from branches that were squash-merged, so git cannot date them and correctness cannot be judged either way.

The nine:

```
11 commits  episode-2026-07-26-session-3423-archive-stale-execution-plans.json
 4 commits  episode-2026-07-27-session-3475-memory-causal-order.json
 5 commits  episode-2026-07-27-session-3605-skill-description-context-cost.json
 2 commits  episode-2026-07-28-session-3499-markdown-it-ast.json
 4 commits  episode-2026-07-28-session-3584-quality-auditor-shared.json
 4 commits  episode-2026-07-28-session-3585-orchestrator-routing.json
 5 commits  episode-2026-07-28-session-3630-episode-reader.json
 7 commits  episode-2026-07-28-session-3650-markdownlint-exclusion-trap.json
 3 commits  episode-2026-07-29-session-3749-orphan-ref-doc-placeholder.json
```

## Why neither repair path works

Re-extracting is the obvious repair. Both flags fail differently.

`--preserve` merges a fresh extraction over the existing artifact but keeps the existing event order. On the 3605 episode it dropped the one bogus tail edge and left the reversed chain exactly as it was.

`--force` produces the correct order but rebuilds `metrics` from the session log alone. On the same episode `metrics.files_changed` went from 3 to 0, because the original artifact carried a value the current session log no longer supports.

So the corpus owner picks between a false causal chain and losing metrics. Neither is acceptable for nine files.

## Shapes a repair has to handle

A naive "sort the commit events in place" repair breaks on two of the three shapes present:

- 6 of 9: commit events are contiguous and sit at the tail of the event list
[...]

Issue #3598, opened 2026-07-28
Quarantine or migrate legacy episode causal edges


rjmurillo:
Issue #3464 adds causal_order_version=2 for newly generated episode event links. The existing corpus still contains unversioned derived links from the old ordering rule.\n\nMeasured on branch fix/3464-memory-causal-order after merging origin/main:\n\n- 62 episode files have unversioned event links.\n- 674 unversioned leads_to edges remain in those files.\n- The known #3459 inversion remains in the historical artifact until migration.\n\nDecision for #3464: do not rewrite historical episode records in the code-fix commit. New consumers can distinguish v2 links by causal_order_version=2. Missing marker means legacy ordering and must not be treated as trusted v2 evidence.\n\nFollow-up options:\n\n1. Regenerate derived caused_by and leads_to fields only, leaving raw event records untouched.\n2. Quarantine old derived links by moving them to an explicit legacy field or removing them from consumer traversal.\n\nRefs #3464.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #3598: Quarantine or migrate legacy episode causal edges
- **Type**: complexity_triage
- **Priority**: high
- **Suggested Agent**: agent-implementer
- Issue labeled with high complexity.
Suggested agents: agent-implementer

### Route issue #3598 to agent-implementer
- **Type**: agent_routing
- **Priority**: low
- **Suggested Agent**: agent-implementer
- Issue content suggests routing to: agent-implementer
Complexity: high

rjmurillo:
Re-measured on `main` at `7bec9080f` before picking an option. Three findings change the shape of this.

## The counts moved, and they keep moving

| Metric | Issue text (branch `fix/3464-memory-causal-order`) | `main` today |
|---|---|---|
| Files with unversioned links | 62 | 103 |
| Unversioned `leads_to` edges | 674 | 1199 |
| `causal_order_version=2` edges | not stated | 0 |

The corpus grew 181 episode files in the last 30 days against 464 commits, because the `extract-session-episodes` pre-commit hook writes one per commit. It is 1.9MB now.

A one-time rewrite is therefore not a migration, it is a snapshot. Whatever runs has to be re-runnable, because the legacy set grows every day this stays open.

## There are zero v2 edges on main

This is the part that blocks both options. The premise here is:

> New consumers can distinguish v2 links by `causal_order_version=2`. Missing marker means legacy ordering.

[...]

PR #4003, opened 2026-07-30
fix(memory): make the health score non-vacuous and repair stale package references

### LINKED ISSUE #3873: Episode metrics record commits 0 alongside a non-zero files_changed in 14 of 331 episodes
## Problem

14 of 331 episode files under `.agents/memory/episodes/` record `metrics.commits == 0` while also recording `metrics.files_changed > 0`. Those two cannot both be true: files do not change in a session that produced no commit.

Found while draining a review thread on PR #3826, where the reviewer noticed the paired session log said `changesCommitted.complete: true` but the episode said `commits: 0`. The branch actually had 11 commits across 5 files. That instance is fixed in the PR. The other 13 are not, and nothing stops the 14th recurring.

## Measurement

```
$ # count episodes where commits==0 but files_changed>0
total episodes: 331
commits==0 while files_changed>0: 14
  episode-2026-06-17-session-2585-fix-2621-exclude-worktrees-plugin-manifest.json  files_changed=5
  episode-2026-06-17-session-2586-fix-2623-priority-label-mutual.json  files_changed=4
  episode-2026-06-17-session-2587-author-proposed-adr-074-remotelocal-pr-autofix.json  files_changed=3
  episode-2026-06-17-session-2587-live-pairwise-skill-overlap-eval-1949.json  files_changed=4
  episode-2026-07-03-session-2604-pr-2829-autoplan-babysitter.json  files_changed=5
  episode-2026-07-03-session-2607.json  files_changed=3
  episode-2026-07-15-session-2993.json  files_changed=4
  episode-2026-07-25-session-2993-wire-orphaned-ratchets.json  files_changed=5
  episode-2026-07-27-session-3474-skill-family-skip-clauses.json  files_changed=5
  episode-2026-07-28-session-3502-remove-causal-graph-continued.json  files_changed=5
  episode-2026-07-28-session-3512-recovery-merge-forward.json  files_changed=3
  episode-2026-07-29-session-3649-subprocess-utf8.json  files_changed=46
  (2 more)
```

The spread of dates (2026-06-17 through 2026-07-29) shows this is not a single bad batch.

## Why it matters

Episodes are read by memory retrieval and by analysis tooling. An episode claiming a session produced no commit while listing 46 changed files will be scored as low-value work when it was the opposite. The failure is silent: no validator reads `metrics`, so the inconsistency never surfaces at write time or in CI.

## Separate observation, not part of this ask

`duration_minutes: 0`, `tool_calls: 0`, and `lessons: []` appear on effectively every episode in the directory, including recent ones. Those look like never-populated placeholders rather than per-episode drift, so they are a different question (should the writer populate them, or should the schema drop them). Naming it here so the next reader does not mistake the uniform zeros for the same defect. Not proposing a change to them under this issue.

## Proposed fix

Add a consistency check to the existing episode validation path:

1. Reject `metrics.commits == 0` alongside `metrics.files_changed > 0`.
2. Where a paired session log exists, cross-check `metrics.commits > 0` against `protocolCompliance.sessionEnd.changesCommitted.complete`.

Then backfill the 14 existing files from git history, since each episode names its session and branch.

Positive, negative, and edge coverage per the repo testing bar: an episode with matching values passes, an episode with `commits: 0` and `files_changed: 5` fails, and `files_changed: 0` with `commits: 0` passes (a genuinely commit-free session).

### LINKED ISSUE #3876: sync_plugin_lib.py destroys the module docstring in all 29 mirrored files
## Problem

`scripts/sync_plugin_lib.py` overwrites the first line of each mirrored module's docstring with a provenance note. For the 17 modules whose source docstring is a single line, that destroys the entire docstring. For the other 12, it destroys the summary line and leaves the body orphaned under a sync note that does not describe the module.

The mirrors are what ship in the installed plugin, so this is the copy users and agents read.

## Measurement

```
mirrored files carrying the canonical note: 29
  entire docstring replaced (single-line source): 17
  summary replaced, body kept (multi-line source): 12
```

Examples of summaries that exist in `scripts/` and are gone from the packaged copy:

| module | summary present in source, absent in mirror |
|---|---|
| `ai_review_common/__init__.py` | AI Review Common module: shared helpers for AI-powered review workflows. |
| `ai_review_common/cache_guard.py` | Cache population guard for the agent-review composite action. |
| `ai_review_common/feature_review.py` | Feature review parsing: recommendation, assignees, labels extraction. |
| `ai_review_common/issue_triage.py` | Markdown formatting, display helpers, and logging for AI review outputs. |

Each mirror instead reads:

```python
"""Canonical: scripts/ai_review_common/cache_guard.py. Sync via scripts/sync_plugin_lib.py."""
```

## Mechanism

`_replace_first_docstring_line()` in `scripts/sync_plugin_lib.py` has two branches, and both discard content rather than prepend to it:

- Single-line docstring: matches `^("""|''')[^\n]*?\1` and rebuilds it as `{quote}{note}{quote}`, dropping the original text entirely.
- Multi-line docstring: finds the first newline and keeps everything after it, dropping the summary line.

## Why it matters

Two separate costs.

First, information loss in the artifact users install. A module named `workflow.py` whose docstring says only where it was copied from tells a reader nothing about what it does. `ai_review_common/issue_triage.py` losing "Markdown formatting, display helpers, and logging" is a real regression for anyone reading the installed plugin.

Second, this is the pattern the repo's universal rule 6 prohibits:

> MUST NOT add auto-generated headers, generation timestamps, or "do not edit" comments to any file (agent prompts, documentation, code, template outputs). Generated output must be indistinguishable from hand-written content.

A provenance note pointing at the generator, occupying the docstring summary, is that pattern. Rule 6 says "If a script grows a helper that emits such headers, delete the helper instead of calling it."

## Proposed fix

Delete `_build_canonical_note`, `CANONICAL_NOTE`, and `_replace_first_docstring_line`, and stop rewriting the docstring during sync. The mirror then carries the source's real docstring and is byte-identical to the source apart from the import rewrite, which makes drift detection simpler rather than harder.

The provenance the note carries is already available and better placed: `sync_plugin_lib.py` names its own source and destination directories, and the drift gate names the canonical file when a mirror diverges.

If provenance must survive in-file, the non-destructive form is to append a line to the existing docstring rather than replace one. That still trips rule 6, so deletion is the recommendation.

## Test coverage

Existing sync tests assert the note is present, so they will need inverting. Positive, negative, and edge per the repo bar: a single-line source docstring survives the round trip unchanged, a multi-line source docstring survives with its summary intact, and a source module with no docstring produces a mirror with no docstring rather than a synthesized one.

## Found via

Draining review threads on PR #3677, while confirming that `scripts/ai_review_common/workflow.py` and its two mirrors were free of conflict markers. The three files differed only by this note, which is what surfaced it.

### LINKED ISSUE #3928: session-end cannot detect a committed Serena memory update
## Summary

The session-end skill requires both `serenaMemoryUpdated` and `changesCommitted`, but its memory detector only looks at staged, unstaged, and untracked paths. A memory update that was committed as instructed becomes invisible; leaving it uncommitted makes `changesCommitted` fail.

## Reproduction

1. Add `.serena/memories/example.md`.
2. Commit it, as the session-end skill instructs.
3. Confirm `git status --porcelain` is clean.
4. Run `complete_session_log.py`.
5. Observe `[TODO] Serena memory not updated` even though the committed memory file is part of the session.

If step 2 is skipped, memory detection succeeds but the same run reports uncommitted changes, so the two MUST gates cannot both become true automatically.

Observed during PR #3688 finalization after committing `.serena/memories/pr-3688-autofix-session.md` in commit `17c50b811`.

## Root cause

`_test_serena_memory_updated()` checks only `git diff --cached --name-only`, `git diff --name-only`, and `git ls-files --others --exclude-standard`. It never checks committed changes made since the session start commit.

## Impact

The advertised automatic session-end workflow cannot complete cleanly without manual field manipulation or a deliberately dirty tree.

## Proposed acceptance criteria

- Detect `.serena/memories/` changes committed between the session log starting commit and HEAD, while retaining staged, unstaged, and untracked detection.
- Keep `changesCommitted` based on a clean working tree.
- Add tests for committed memory, uncommitted memory, no memory update, and a memory update that predates the session start.
- Update session-end documentation to match the implemented evidence model.

## Duplicate search

Searches for `Serena memory not updated changes committed` and `session-end memory committed detection` found no matching issue.

### LINKED ISSUE #3972: Episode extractor writes duration_minutes and tool_calls as zero for every episode
## Summary

`extract_session_episode.py` writes `duration_minutes: 0` and `tool_calls: 0` into every episode it produces. Not sometimes. Every one.

## Measurement

The fourteen most recent episodes in `.agents/memory/episodes/`:

```
episode                   com  dur  tool  files
2026-07-29-session-374      1    0     0      3
2026-07-29-session-374      0    0     0      5
2026-07-29-session-374      3    0     0      3
2026-07-29-session-377      1    0     0      5
2026-07-29-session-377      0    0     0     11
2026-07-29-session-377      1    0     0      2
2026-07-29-session-377      1    0     0     13
2026-07-29-session-378      2    0     0      4
2026-07-29-session-381      1    0     0      5
2026-07-30-session-342      1    0     0      3
2026-07-30-session-360      1    0     0      2
2026-07-30-session-360      1    0     0      3
2026-07-30-session-377      0    0     0      3
2026-07-30-session-387      0    0     0      5

dur all zero: True   tools all zero: True   n=14
```

`files_changed` and `commits` vary, so those two are being read from somewhere. `duration_minutes` and `tool_calls` never move off zero.

## Cause

`parse_metrics` in `.claude/skills/memory/scripts/extract_session_episode.py` derives its counts by scanning the session log line by line with regular expressions. That shape works for a markdown log. Session logs are JSON, so the patterns that would carry a duration or a tool count find nothing and the fields fall back to their zero default.

`commits` survives because its pattern (`[a-f0-9]{7,40}`) matches a bare sha wherever it appears, including inside JSON string values. It is finding the right answer for the wrong reason, which is why it looks fine while its neighbours do not.

## Why it matters

Two of these fields are load-bearing for anything that wants to rank or filter episodes. A retrieval step that prefers longer or more active sessions currently sees every episode as a zero-minute, zero-tool session and cannot order them. A reviewer reading an episode has no way to tell an unpopulated field from a genuinely short session.

There is a second cost. A reviewer on PR #3961 read `commits: 1` against a session that made eight commits and flagged it as wrong. It was not wrong for that file, since an episode is a point-in-time extraction carrying one commit event, but the surrounding zeros made the whole metrics block look untrustworthy enough to be worth challenging.

## Suggested fix

Read the metrics from the parsed JSON rather than from a line scan. The session log already carries the fields:

- `duration_minutes` from the session start and end timestamps
- `tool_calls` from the recorded tool invocations
- `commits` from the `workLog` commit entries, or from the episode's own `commit` events if an episode is meant to stay point-in-time

Whichever scope is chosen for `commits`, it should be stated in the extractor, because the ambiguity is what drew the review comment.

Keep a fallback path for any older markdown-shaped log still in the tree, but let the JSON path win when the log parses as JSON.

## Verification

Any fix should assert that a session log with a known duration, a known tool count, and a known commit count round-trips into an episode carrying those three numbers, plus a negative case where an unparseable log still produces a valid episode rather than raising.

Note that `.claude/skills/memory/scripts/` ships in the project-toolkit plugin, so a change here needs the manifest bumps and mirror regeneration that go with it.

### LINKED ISSUE #3980: fix(memory): _calculate_health_score returns 1.0 when corpus has memories but no citations
## Problem

`_calculate_health_score` in `scripts/memory_enhancement/health.py` has two early-return branches, both returning 1.0:

- Line: `if total_memories == 0: return 1.0` (correct: empty corpus is vacuously healthy)
- Line: `if counts["total"] == 0: return 1.0` (bug: 878 memories with no citations scores 100%)

## Measurement (2026-07-30)

```
$ uv run --frozen python -m memory_enhancement --repo-root . health --json
{
  "total_memories": 878,
  "total_citations": 0,
  "valid_citations": 0,
  "health_score": 1.0,
  "recommendations": ["Add citations to memories for traceability."]
}
```

878 memories, 0 citations, health score 1.0. Exactly ONE memory carries a citations field (`.serena/memories/testing/test-citation-verification.md`), and that path reference is stale (points at the old pre-move location). The score is structurally incapable of falling below 1.0 regardless of the corpus state.

The existing recommendation ("Add citations to memories for traceability.") shows the code knows the state is wrong, but the score does not reflect it.

## Root cause

When `counts["total"] == 0` and `total_memories > 0`, the function short-circuits to 1.0 instead of returning a penalized score. The intent was to avoid a 0/0 divide, but the vacuously-correct result defeats the entire purpose of the health check.

## Proposed fix

Return 0.0 when `total_memories > 0` and `counts["total"] == 0`. Memories without citations have no citation health to measure; a zero score honestly signals that.

Keep the `total_memories == 0` branch returning 1.0 (empty corpus is not unhealthy).

## Tests required

- Empty corpus (0 memories): score == 1.0 (no regression)
- Memories with no citations: score == 0.0
- Memories with all-valid citations: score == 1.0
- Mixed corpus (valid + broken): score is between 0 and 1

### LINKED ISSUE #4010: memory_enhancement: search emits ids that verify --memory-id rejects for 783 of 879 memories
## What happens

`search` reports a `memory_id` that `verify --memory-id` rejects for any memory
stored in a subdirectory. That is 783 of the 879 memories in the corpus, so the
two commands agree on 11 percent of it.

## Reproduction

```
$ uv run python -m memory_enhancement --repo-root . search "citation" --top 3 --json
[
 {
  "memory_id": "pr-review-batching-strategy",
  "file_path": "/repo/.serena/memories/workflows/pr-review-batching-strategy.md"
 },
 ...

$ uv run python -m memory_enhancement --repo-root . verify --memory-id pr-review-batching-strategy
Memory not found: pr-review-batching-strategy
$ echo $?
1

$ uv run python -m memory_enhancement --repo-root . verify --memory-id workflows/pr-review-batching-strategy
workflows/pr-review-batching-strategy: no citations
$ echo $?
0
```

`search` emits the bare stem. `verify` wants the path-qualified id.

## Blast radius

```
$ find .serena/memories -name '*.md' | awk -F/ '{print (NF>3?"subdir":"root")}' | sort | uniq -c
     96 root
    783 subdir
```

## Why it matters

The obvious workflow is search, then verify the interesting hit. That workflow
fails for 89 percent of the corpus, and the failure reads like the memory does
not exist rather than like the id is in the wrong shape.

`verify` does exit 1, so nothing fails open. The cost is a dead end, not a false
pass.

## Suggested fix

Make `search` emit the id `verify` accepts. Emitting the path-qualified form is
the smaller change and keeps a single id shape across the CLI. Accepting both
shapes in `verify` would also work but leaves two spellings for one thing.

Found while auditing the memory-enhancement documentation for #3996.

### LINKED ISSUE #4011: All three memory hooks are unregistered, so the memory system has no automatic path
## What happens

None of the three memory hooks is registered under any hook event, so the memory
system has no automatic ingestion, recall, or reinforcement path. Every memory
in the corpus got there because someone ran a command.

## Reproduction

```
$ python3 -c "import json; print(list(json.load(open('.claude/settings.json'))['hooks'].keys()))"
['SessionStart', 'PostToolUse', 'PreCompact']

$ for h in user_prompt_submit_memory post_tool_call_memory session_end_memory; do
    echo "$h: $(grep -rl $h .claude/settings.json .claude/hooks/ lefthook.yml 2>/dev/null | wc -l)"
  done
user_prompt_submit_memory: 0
post_tool_call_memory: 0
session_end_memory: 0
```

The three registered events exist, but none of their entries points at a memory
hook.

## What the code expects

```
scripts/memory_enhancement/hooks/user_prompt_submit_memory.py   (auto-recall)
scripts/memory_enhancement/hooks/post_tool_call_memory.py       (fact capture)
scripts/memory_enhancement/hooks/session_end_memory.py          (reflection)
```

All three are implemented and covered by tests. `reinforce_memories()` in
`reflection.py` persists confidence through `save_memory`, and its only
non-test caller is the unregistered reflection hook. So confidence decay and
reinforcement never run outside a test.

## Why it matters

The practical consequence is that confidence scores are computed on read and
never persisted, and no session contributes a memory without an explicit
command. Anything reasoning about the corpus as a self-maintaining system is
reasoning about a capability that is built but not wired.

## Decision needed

Either register the three hooks, or stop describing them as a live pipeline.
The spec at `.agents/specs/hook-protocol.md` lists all three under "Reference
Implementations", which is defensible phrasing, but readers reasonably take a
shipped hook with tests as a running one.

`.claude/skills/memory-enhancement/references/confidence-scoring.md` already
carries the accurate statement for the reflection hook after #3996. That wording
should be widened to all three, or made obsolete by wiring them.

Found while auditing the memory-enhancement documentation for #3996.

rjmurillo:
## Summary

This PR fixes nine issues in the memory subsystem and repairs stale references
left behind when the `memory_enhancement` package moved from
`.claude/skills/memory-enhancement/src/` to `scripts/memory_enhancement/`
in commit `2aeb4fddd`.

### Health score fix (issue #3980)

`_calculate_health_score` in `scripts/memory_enhancement/health.py`
returned `1.0` when `counts["total"] == 0` even if the corpus had memories
with no citations at all. The bug made the score vacuously 100% regardless
of corpus quality. Measurement: 878 memories, 0 citations, score 1.0 before
this fix.

Fix: return `0.0` when the corpus has memories but no citations.
Empty corpus (`total_memories == 0`) still returns `1.0` (vacuously correct).

Health output before:

```json
{"total_memories": 878, "total_citations": 0, "health_score": 1.0}
```

Health output after:

```json
{"total_memories": 878, "total_citations": 0, "valid_citations": 0,
 "stale_citations": 0, "broken_citations": 0, "unverified_citations": 0,
 "health_score": 0.0, "stale_memories": [...],
 "recommendations": ["Review 7 stale memory/memories for relevance.",
 "Add citations to memories for traceability."]}
```

### Docstring destruction fix (issue #3876)

`sync_plugin_lib.py` replaced the first docstring line of every mirrored
file with a canonical note, destroying the source module's actual
documentation. The three functions responsible for this
(`CANONICAL_NOTE`, `_build_canonical_note`, `_replace_first_docstring_line`)
are deleted. The `_transform_file` function now does only import conversion.

All 29 mirrored files in `.claude/lib/` and `src/copilot-cli/lib/` are
regenerated with their real docstrings restored.

### Session-end committed memory detection (issue #3928)

`_test_serena_memory_updated()` in the session-end skill checked only
staged, unstaged, and untracked files. Committed `.serena/memories/`
changes made before `session-end` ran were invisible to the check.

Fix: accept an optional `starting_commit` argument. When provided, run
`git log --name-only --format= <starting_commit>..HEAD` and return `True`
if any `.serena/memories/` path appears. The call site passes
`session["session"]["startingCommit"]` when present.

### Episode metrics consistency validation (issue #3873)

`validate_episode_file` in `extract_session_episode.py` called only
`validate_event_ids`. The case of `metrics.commits == 0` with
`metrics.files_changed > 0` (commit collection failure) was silently
ignored.

Fix: add `validate_metrics_consistency()` and call it from
`validate_episode_file`. The `--validate` CLI now surfaces these
violations. 17 pre-existing episodes have this pattern; the regression
test caps the count at 17 so new violations are caught without failing
on the known-bad history.

### Episode duration and tool_calls fields (issue #3972)

`json_metrics()` initialised `duration_minutes` and `tool_calls` to 0 and
never populated them. Every episode therefore reported zero for both fields
regardless of actual session activity.

Fix:
- `_duration_from_worklogs()`: compute minutes from first-to-last `workLog`
  timestamp. Returns 0 when fewer than two timestamped entries exist.
- `_duration_from_metrics_block()`: parse the old-schema `metrics.duration`
  string (e.g. "~20 minutes") or integer; fallback for sessions that predate
  the workLog timestamp convention.
- `json_metrics()` prefers workLog timestamps; falls back to the metrics block.
- `tool_calls` reads `metrics.toolCalls` from the old schema. Modern session
  logs carry no structured tool count, so that field stays 0 for modern
  sessions. An honest zero is distinguishable from the old situation because
  duration is now non-zero whenever workLog timestamps span real time.

Verification against real sessions:

Modern session (2026-07-30, workLog with timestamps):
`duration_minutes: 13, tool_calls: 0`

Old-schema session (2026-01-16, metrics.toolCalls present):
`duration_minutes: 20, tool_calls: 24`

Before this fix both sessions reported `duration_minutes: 0, tool_calls: 0`.

### Memory ID mismatch fix (issue #4010)

`search` emitted bare-stem IDs (e.g. `pr-review-batching-strategy`) while
`verify --memory-id` expected path-qualified IDs (e.g.
`workflows/pr-review-batching-strategy`). Result: 783 of 879 memories were
unreachable via verify after a search. Verified by re-running both commands
against the real corpus.

Root cause: `load_memory(file_path)` used `Path.stem`; `load_memories()`
overrode with the path-qualified form; `_score_memory_file` in
`scripts/memory_enhancement/search.py` called `load_memory` directly and
got the bare stem.

Fix: added `memory_id_from_path(file_path, memories_dir)` as the single
source of truth in `scripts/memory_enhancement/serena_integration.py`.
When `memories_dir` is provided, returns the path-relative form without
suffix. Both `load_memory` and `search._score_memory_file` delegate here.
The bare-stem fallback is preserved when `memories_dir` is absent for
backward compatibility.

Tests: 3/3 mutants killed. Bidirectional test verifies every search-emitted
ID is accepted by `load_memories`. Negative control verifies a fabricated ID
is still rejected.

### Memory hooks documentation fix (issue #4011)

Verified: none of `user_prompt_submit_memory`, `post_tool_call_memory`, or
`session_end_memory` appear in the Claude settings or hooks directory.
The memory system has no automatic ingestion path.

Fix: updated `.agents/specs/hook-protocol.md` to explicitly state the three
hooks are implemented but not currently registered, with instructions for how
to activate them.

### Stale workflow path filters

`.github/workflows/citation-verify.yml` line 46: updated stale path filter
from `memory-enhancement/src/**` to `scripts/memory_enhancement/**`.

The Serena memory `testing/test-citation-verification` had a stale citation
path. Updated the path in `.serena/memories/testing/test-citation-verification.md`
to reference the current location.

## Testing

Memory tests: 1041 passed (added 19 new tests for issue #3972, 15 new for
issue #4010, plus search bidirectional and serena integration tests).

Mutations killed: 11/12 across five new behaviors; one equivalent mutant noted.

- `_calculate_health_score` with `counts["total"] == 0`: 2 mutants killed
- `_test_serena_memory_updated` git-log branch: 2 mutants killed
- `validate_metrics_consistency` condition: 2 mutants killed
- `_duration_from_worklogs` / `_duration_from_metrics_block` / `json_metrics`:
  3 mutants killed; M1 is semantically equivalent (single timestamp yields
  delta=0 whether guarded by `< 1` or `< 2`).
- `memory_id_from_path`: 3/3 mutants killed

## Stale reference sweep

```
grep -rn "memory-enhancement/src|memory_enhancement" --include=*.yml ...
```

Hits found:

- `.github/workflows/citation-verify.yml` line 46: fixed in this PR
- The memory-validation workflow already had the correct path filter `scripts/memory_enhancement/**`, no change needed.
- `.serena/memories/testing/test-citation-verification.md`: fixed in this PR
- The memory-health workflow belongs to the concurrent worktree; not touched per coordination rules.

## Issues

Fixes #3876
Fixes #3873
Fixes #3928
Fixes #3972
Fixes #3980
Fixes #4010
Fixes #4011

## References

- Issue #3598 (unversioned causal edges): measured 318 episodes, 1392 edges.
  Data migration is out of scope. Commented on the issue with measurement.
- Issue #3765 (backwards commit chains): all 9 episodes confirmed present.
  No lossless repair path. Commented on the issue with measurement.
- `scripts/memory_enhancement/models.py` line 38 is where the citation path
  in `.serena/memories/testing/test-citation-verification.md` now points.
- The Claude settings file and hooks directory were verified to have no
  registration for the three memory hooks (evidence for issue #4011).


rjmurillo:
## Coordination note: PR #4008 documents the old health score

Adversarial review found that PR #4008 asserts a nonempty uncited corpus scores `1.0` in three places, which this PR changes to `0.0` at `scripts/memory_enhancement/health.py:244-247`.

The two PRs touch disjoint files, so there is no merge conflict to catch it. Whichever lands second leaves documentation that contradicts the runtime.

This PR's scoring was independently verified as non-vacuous: empty corpus `1.0`, nonempty uncited `0.0`, all-broken `0.0`. The correction belongs in #4008, not here.

Auto-merge is disabled on both until #4008 is updated.

Refs #4008

Issue #3963, opened 2026-07-30
fix(hooks): failing lefthook hook skips all later hooks as 'broken pipe' and leaves the worktree mutated


rjmurillo:
## Summary

Two reproduced defects in the pre-commit hook run, plus one observed-but-not-
reproduced data-loss report. All three concern what happens when a hook inside
the `parallel` group at lefthook.yml fails.

## Defect 1: a failing hook silently skips every later hook, mislabelled "broken pipe"

When `markdown-autofix` exits non-zero, every remaining hook in the run is
skipped and the reason printed is `(skip) broken pipe`. Eleven hooks never
execute, including `markdown-check`, `github-bash-policy`, and `scope-policy`.

Isolating control, same repo, same command shape, only the markdown content
differs:

| Case | broken-pipe skips | hooks run | commit rc |
|---|---|---|---|
| `markdown-autofix` fails (lint-dirty file) | 11 | 0 later | 1 |
| `markdown-autofix` passes (lint-clean file) | 0 | 13 | 0 |

Neither run pipes output. Both redirect to a file. The label is wrong: nothing
was piped, so no pipe was broken.

Impact is developer experience, not a gate bypass. The failing commit is
correctly rejected. But the operator learns about exactly one failure per
attempt, and the printed cause points at plumbing rather than at the real
problem. On a repo where a full hook run is 13 hooks, batching failures is
worth real minutes per commit.

I previously believed this cascade was caused by piping `git commit` through
`tail`. That was wrong. The control above changes only the file content and
reproduces the cascade with no pipe present.

## Defect 2: autofix mutates the working tree during a run that then fails

Same scenario, inspecting state after the failed commit:

```text
staged blob:   "  - one"     (original, unfixed indentation)
worktree copy: "- one"       (rewritten by markdown-autofix)
```

`markdown-autofix` edited the file on disk, did not re-stage the result, and
the commit then failed. The working tree is now different from both `HEAD` and
the index, and nothing told the operator that a tool rewrote their file.

This is the precondition that makes any restore-path bug destructive: the tool
is mutating uncommitted state during a run it is about to abort.

## Observation 3: unrelated unstaged edits lost once (NOT reproduced)

During unrelated work, a failed `markdown-autofix` run printed:

```text
Unable to restore previously hidden unstaged changes
```

and unstaged edits to a workflow file that autofix never touched were gone
from the working tree. It was caught only because a later grep contradicted an
earlier measurement.

[...]

Issue #3960, opened 2026-07-30
chore(ci): COPILOT_GITHUB_TOKEN expired; nightly CLI smoke red 8 nights running


rjmurillo:
## Summary

`COPILOT_GITHUB_TOKEN` is expired or revoked. The nightly CLI smoke has failed
**8 consecutive nights** (2026-07-23 through 2026-07-30) on all three
platforms. Only the repository owner can rotate the secret, so this needs an
owner action, not a code change.

## Evidence

Run list, `nightly-cli-smoke.yml`:

```text
2026-07-30 failure 30522227014
2026-07-29 failure 30430970094
2026-07-28 failure 30337632337
2026-07-27 failure 30246325012
2026-07-26 failure 30192390222
2026-07-25 failure 30148661127
2026-07-24 failure 30074662287
2026-07-23 failure 29987479519
```

From run 30522227014, identical on `ubuntu-latest`, `macos-latest`, and
`windows-latest`:

```text
Failed: Copilot auth token was rejected (expired or revoked); rotate
COPILOT_GITHUB_TOKEN for the smoke job. The shipped-base dogfood never ran
(issue #3275). rc=1 stderr="  Failed to fetch PAT user login (401):
GitHub returned: Bad credentials
```

Both smoke tests fail the same way:

- `tests/e2e/test_cli_hook_e2e.py::test_copilot_vendor_install_hook_resolves`
- `tests/e2e/test_plugin_load_smoke.py::test_copilot_plugin_loads_expected_skills`

## What is working

The failure message is precise and names its own remedy. `assert_smoke_ran.py`
did its job: a missing or rejected credential surfaces as a red run, not a
silent green one. That design (issue #2231 item 4) held up.

## Impact

The real-CLI smoke is the only gate that proves the shipped plugin loads in
the actual Copilot CLI. Bare CI cannot verify the launcher contract. With the
token dead, that coverage has been absent for over a week, and the Claude side
of the same job is not exercising the Copilot path at all.

## Fix

Owner rotates the secret. Generate a fresh PAT with the scopes the Copilot CLI
requires and update the `COPILOT_GITHUB_TOKEN` repository secret. Re-run
`nightly-cli-smoke.yml` via `workflow_dispatch` to confirm green before
closing.

## Related

Token expiry is silent until the next scheduled run. Consider a short
follow-up that checks credential validity on a cadence shorter than the smoke
itself, so the window between expiry and detection is not a full day.

Refs #3275

[...]

Issue #3947, opened 2026-07-30
Semver milestone detection duplicated across three files (DRY at knowledge level)


rjmurillo:
## Problem

The rule 'what counts as a semantic-version milestone title and how it sorts' lives in three copies:

- `.github/scripts/set_item_milestone.py` (workflow copy)
- `.claude/skills/github/scripts/milestone/set_item_milestone.py` (also embeds its own copy of latest-detection)
- `.claude/skills/github/scripts/milestone/get_latest_semantic_milestone.py` (plus src/copilot-cli mirrors)

PR for #3945 had to patch the same two-line rule in five files and keep four test suites in step. Next change to milestone semantics will drift unless someone remembers all sites; the v-prefix bug itself survived unnoticed partly because fixing one copy would still leave the workflow copy broken.

## Proposed fix

Single `_semver.py` helper (pattern + parse + latest-selection) under the milestone script directory, imported by both skill scripts; workflow copy either imports the vendored module or is generated from it per the mirror discipline in .claude/rules/knowledge-persistence.md.

Refs #3945, pragmatic-programmer rule (DRY at knowledge level).

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #3947: Semver milestone detection duplicated across three files (DR
- **Type**: complexity_triage
- **Priority**: medium
- **Suggested Agent**: agent-architect
- Issue labeled with medium complexity.
Suggested agents: agent-architect, agent-devops, agent-qa, agent-milestone-planner

### Route issue #3947 to agent-architect, agent-devops, agent-qa, agent-milestone-planner
- **Type**: agent_routing
- **Priority**: low
- **Suggested Agent**: agent-architect
- Issue content suggests routing to: agent-architect, agent-devops, agent-qa, agent-milestone-planner
Complexity: medium

rjmurillo:
Measured: semver milestone detection IS duplicated across .github/scripts/set_item_milestone.py, .claude/skills/github/scripts/milestone/set_item_milestone.py, and get_latest_semantic_milestone.py. The DRY fix (a shared _semver.py helper) is valid but requires: (1) creating the shared module, (2) updating 3 files to import it, (3) keeping the .github/scripts/ workflow copy in sync (per mirror discipline). Scope exceeds this CI-infrastructure cluster. Closing as deferred; needs its own PR targeting the github skill milestone scripts.

Issue #3884, opened 2026-07-30
Count ratchets compare against a committed absolute baseline, so main drift makes any in-review ratchet PR red on arrival


rjmurillo:
## Symptom

Any pull request that ratchets a repository-wide count is red on arrival if
`main` drifts while the pull request sits in review. The baseline is a number
committed in the branch; the measured value is computed against the working
tree, which includes whatever landed on `main` since the branch was cut.

## Measurement

Observed on PR #3824 (`scripts/ci/count_ratchet.py`, taste-lint error counts):

1. Branch cut with baseline 615, measured 615, green.
2. After a rebase, the measured tracked-tree count read 616 against baseline
   615, red.
3. Measured `origin/main` itself in a detached scratch worktree: main is at 616.
   The drift was not caused by the branch.

The extra violation traced to a complexity-12 function landed the same morning
in PR #3799. It was fixable, so I extracted two helpers rather than raise the
baseline, and the count returned to 615. That resolution does not generalize:
the next drift may be a violation the ratchet author cannot or should not fix.

### How to measure main's own count

Add a detached worktree and copy `scripts/ci/count_ratchet.py` and
`scripts/ci/taste_count_ratchet.py` into it. The copies are untracked, so the
`tracked_files` walk correctly ignores them and the count reflects main alone.
Running the scripts from the branch worktree against a main checkout does not
produce the same answer.

## Why raising the baseline is the wrong reflex

The number is written in more than one place. For the taste-lint ratchet it
appears in the baseline file, in the module docstring, and in the pull request
prose. Bumping it to unblock a review silently ratchets the wrong direction and
launders whatever debt caused the drift.

## Suggested directions

Option space, not a recommendation:

1. Compute the baseline from the merge-base rather than from a committed
   constant, so drift on `main` after the branch point cannot make the branch
   red.
2. Compare the branch delta rather than the absolute count: fail only when the
   branch itself adds violations, regardless of what main did.
3. Keep the absolute count but have CI report `main`'s current count alongside
   the branch's, so a red ratchet is immediately attributable.

Option 2 is the smallest change that removes the class, and it is the semantics
the ratchet is actually trying to express: this branch must not make things
worse.

## Evidence

PR #3824, PR #3799. Retrospective:
`.agents/retrospective/2026-07-29-measure-before-asserting.md`, Follow-Ups
section item 3.

rjmurillo:
[...]

PR #4009, opened 2026-07-30
fix(ci): eliminate false-green type checks and fail-open workflow steps

### LINKED ISSUE #3916: Portability validator is blind to skill-relative paths that resolve to nothing
The portability ratchet in `scripts/validation/check_skill_md_exec_portability.py` only counts one shape of bad invocation. A different shape, just as broken, passes silently.

## Three path forms in skill docs

1. `python3 scripts/X.py`, skill-relative. Correct only when `X.py` sits in the same skill directory. When it does not, the command resolves to nothing.
2. `python3 .claude/skills/<s>/scripts/X.py`, checkout-absolute. Works in a checkout, fails in a plugin install. This is the form the ratchet counts.
3. `python3 "${COPILOT_PLUGIN_ROOT:-${CLAUDE_PLUGIN_ROOT:-.claude}}/skills/<s>/scripts/X.py"`. Correct everywhere.

The ratchet enforces 2 into 3. Nothing enforces 1 into 3, and nothing checks that form 1 resolves at all.

## Reproduction

Both injections went into the same file, one at a time, each reverted after.

Form 2, a path that exists but is not portable. Caught:

```
$ printf "\n```bash\npython3 .claude/skills/memory/scripts/test_memory_health.py --format json\n```\n" >> .claude/skills/memory-search/references/skill-reference.md
$ python3 scripts/validation/check_skill_md_exec_portability.py
Skill exec-path vendor-portability drift detected (issue #2838):
  [DRIFT] .claude/skills/memory-search/references/skill-reference.md: 10 bare '.claude/skills/...' invocation(s) (baseline 9).
$ echo "RC=$?"
RC=1
```

Form 1, a path to a script that does not exist anywhere in the tree. Silent:

```
$ printf "\n```bash\npython3 scripts/this_script_does_not_exist_xyz.py --run\n```\n" >> .claude/skills/memory-search/references/skill-reference.md
$ python3 scripts/validation/check_skill_md_exec_portability.py
No skill exec-path vendor-portability drift. 340 grandfathered invocations across 94 files (baseline 382).
$ echo "RC=$?"
RC=0
```

The validator works. It just does not look at this shape.

## Why it matters

A form-2 path is inconvenient. It runs for a contributor in a checkout and only breaks for a plugin consumer. A form-1 path pointing outside its own skill is worse: it runs for nobody, in any context, and the ratchet reports green.

Real instance, `.claude/skills/memory-maintenance/references/benchmarking.md` on `main` carries 11 invocations of `python3 scripts/measure_memory_performance.py`. That script lives in the `memory` skill, not `memory-maintenance`, so every one of them resolves to nothing:

```
$ bash -c "python3 scripts/measure_memory_performance.py --serena-only >/dev/null 2>&1; echo rc=$?"
rc=2
```

The same command with the plugin-root form returns 0 under all three env-var states (both unset, `CLAUDE_PLUGIN_ROOT` set, `COPILOT_PLUGIN_ROOT` set).

## Suggested check

For any `python3 <relative>/X.py` in a skill markdown file, resolve `<relative>` against the owning skill directory. If no file lands there, fail. This is a existence check, independent of the portability question, and it would have caught the benchmarking.md instance the day it was written.

Related but distinct: #3730 covers slack in the existing ratchet. This is about a shape the ratchet never inspects.

Refs #2838, #3730

### LINKED ISSUE #3966: fix(tooling): git_hook_policy.py mypy exits 0 with no file arguments while the push gate fails
## Summary

`scripts/validation/git_hook_policy.py mypy` invoked with no file arguments
exits 0 on a tree that the `python-type-check` push hook rejects. The
documented-by-usage pre-flight gives a false green, so the error is discovered
7+ minutes later when the push is refused.

## Isolating control

Same worktree, same commit, same command. Only the file arguments differ:

```text
$ uv run --frozen --extra dev python scripts/validation/git_hook_policy.py mypy
  rc=0

$ uv run --frozen --extra dev python scripts/validation/git_hook_policy.py mypy $CHANGED_PY_FILES
  rc=1
scripts/ci/retrospective_gate.py:61: error: Argument 1 to "int" has incompatible
type "str | None"; expected "str | Buffer | SupportsInt | SupportsIndex"  [arg-type]
```

The push hook passes `{push_files}`:

```yaml
- name: python-type-check
  run: uv run --frozen --extra dev python scripts/validation/git_hook_policy.py mypy {push_files}
```

The bare form checks a default set that excludes the changed files, so it
reports success while the real gate fails.

## Impact

A contributor runs the bare form as a pre-flight, sees rc=0, pushes, and the
push is rejected after the full hook chain runs. In this repo that chain
includes `python-tests` at ~380s, so the false green costs a full cycle. It
also trains people to distrust the pre-flight and skip it, which is worse.

This is a fail-open in the direction that matters: the cheap local check says
clean, the expensive remote check says dirty.

## Suggested fixes

Pick one:

1. **Make the bare form check the same set.** With no arguments, default to the
   changed files against the merge base rather than to a set that can be empty.
2. **Refuse the bare form.** Exit 2 (config error, per the repo's exit-code
   convention) with a message naming the file-argument form. An unusable
   invocation should not look like a passing one.
3. **Document the exact pre-flight command** next to the hook definition so the
   copy-paste path is the correct one.

Option 1 is the least surprising: a check that takes no arguments should check
what you are about to push.

## Reproduction

Introduce any type error in a changed `.py` file, then run both forms.

```bash
BASE=$(git merge-base HEAD origin/main)
FILES=$(git diff --name-only "$BASE"...HEAD -- '*.py' | tr '\n' ' ')
uv run --frozen --extra dev python scripts/validation/git_hook_policy.py mypy          # rc=0
uv run --frozen --extra dev python scripts/validation/git_hook_policy.py mypy $FILES   # rc=1
```

## Related

Separately, measuring the result with `cmd | tail -N; echo $?` reports the
exit status of `tail`, not of the command. Use `${PIPESTATUS[0]}` or redirect
to a file. That is a caller error rather than a repo defect, but it is an easy
way to turn this false green into a confident one.

rjmurillo:
## Summary

Eliminates false-green type checks and fail-open scan paths in CI hooks and
validation scripts.

### #3966 - mypy bare invocation is a false green

`run_mypy([])` returned rc=0 regardless of repo state because an early-exit path
fired when the filtered path list was empty. A push hook passing empty
`{push_files}` would silently miss real type errors.

Fix: added a hard guard at the top of `run_mypy()` in
`scripts/validation/git_hook_policy.py`. When the paths argument is empty the
function prints an error to stderr and returns rc=2.

Isolating control (before fix): same tree, bare form rc=0, file-argument form
rc=1. After fix: bare form rc=2 (refused to run), file-argument form rc=1
(type error still caught). Both forms now surface the error.

Fixes #3966

### #3916 - portability validator blind to skill-relative script paths

The portability validator checked paths starting with `.`, `$`, or `/` but
missed skill-relative forms like `python scripts/run.py` inside a skill
directory. Such a path resolves relative to the skill root, not the repo root;
a script not found at either location is a silent dangling reference.

Also closed a fail-open path: if `scan_dangling_skill_relative_scripts()` raised
OSError (unreadable file, bad UTF-8), the original code caught it, set
`dangling=[]`, and allowed exit 0. Changed to `return 2` so I/O failures
surface as real failures.

Fix: added `find_skill_relative_scripts()`, `_scan_skill_for_dangling()`, and
`scan_dangling_skill_relative_scripts()` to
`scripts/validation/check_skill_md_exec_portability.py`. The scanner checks
both skill-relative and repo-relative resolution and exits 1 on any dangling
path. OSError exits 2 (fail-closed).

Fixes #3916

## Test counts

- #3966: 2 tests in `tests/test_lefthook_integration.py`, 2/2 mutations killed
- #3916: 12 new tests in `tests/validation/test_check_skill_md_exec_portability.py`
  including `test_dangling_scan_oserror_returns_exit_2`, 1/1 mutation killed
  on the fail-closed path

## Closed issues (unmerited or owner-action only)

Closed with measurement before this PR:

- #3960: `COPILOT_GITHUB_TOKEN` expired, secret rotation is owner action only.
- #3755: Requires GitHub ruleset `strict=true`, owner action only.
- #3963: `broken pipe` message is upstream lefthook behavior, not fixable here.
- #3882: Git push race condition, operator workaround only.
- #3846: Author-deferred design decision.
- #3848: Advisory only, no code change needed.
- #3947: Large DRY semver refactor, deferred as separate work.
- #3884: Architectural ratchet change requiring multi-PR coordination, deferred.

## References

- Issue #3966: bare mypy invocation returns false green
- Issue #3916: portability validator misses skill-relative paths

Re-ran the title check: this PR lost its pull_request events on an earlier push, so the required checks never reported. See issue #4359.


rjmurillo:
## Adversarial review: two blocking findings

### CRITICAL: any workflow timeout becomes a passing gate

`scripts/validation/run_workflow_local_test.py:791-798` assumes exit code 124 means a cold action cache and unconditionally downgrades it to `StageResult.ok=True`. A timeout cannot distinguish a slow download from a hung or broken workflow.

Trigger: a workflow action deadlocks, or runs past the 300-second budget.

Observed after injecting rc 124 into the real stage:

```text
ok=True
detail=[WARN] .github/workflows/ci.yml: dry-run exceeded 300s...
```

The workflow never completes and local validation reports success.

This is the same silent-pass shape the PR is meant to remove (see #4068), so it undercuts the PR's own subject.

**The tests require the false green** at `tests/validation/test_run_workflow_local_test.py:1620-1644`. That is a test codifying the defect: it must be inverted, not deleted.

Suggested fix: timeouts fail by default. Restrict any downgrade to a positively identified download or cache condition, and assert on `main()`'s exit code rather than the helper's return value.

### CRITICAL: plugin manifests are behind main

Both `.claude/.claude-plugin/plugin.json:4` and `src/copilot-cli/.claude-plugin/plugin.json:4` carry `0.6.5428`; main is ahead. `Validate Plugin Version Bump` reports `NOT BUMPED` in job `90931631330`.

Rebase and set both manifests to the same version strictly greater than main.

Auto-merge is disabled on this PR until the timeout finding is resolved.

Refs #4068

rjmurillo:
Thread PRRT_kwDOQoWRls6VnYqX: Fixed in commit 69f556c6a. The guard is now at the top of run_mypy() before the filtering loop. run_mypy([]) returns rc=2 with an error message. The if not checked_paths: return 0 path is now only reachable when paths is non-empty but all provided paths do not exist on disk (a legitimate skip). Test test_mypy_policy_rejects_empty_paths verifies rc=2; mutation harness (return 0 mutant) killed 1/1.

Issue #3943, opened 2026-07-30
feat(validation): instruction budget gate phase 2 (enforce)


rjmurillo:
## What

Instruction budget gate phase 1 shipped always-on in observe mode (#3419, #3427). Phase 2: enforce the budget, failing validation when instruction surface exceeds it.

## Why

Observe mode collected the data; enforcement stops instruction-surface regressions from merging. Second enforcement flip in the v0.7.0 theme.

## Acceptance criteria

- Budget thresholds ratified from phase-1 observation data
- Gate fails closed on budget breach
- Breaking-change note drafted for v0.7.0 release notes

Refs #3419, #3427.

rjmurillo:
Closing as already implemented. The instruction budget gate runs with `--ci` flag in `validate_instruction_budget` (checks_tooling.py:450) which exits 1 when any language exceeds its ceiling. pre_pr_sequence.py calls this as a blocking gate (not advisory). Current measurement: all languages PASS (py: 95.7%, cs: 95.6%, ps1: 95.4%, md: 98.3% of ceiling). Enforcement is active.

Issue #3942, opened 2026-07-30
feat(validation): flip ADR-080 model-pin check from warn to blocking


rjmurillo:
## What

ADR-080 model-pin governance check runs in warn mode with a draining baseline (scripts/validation/model_pin_baseline.json, 4 entries remaining). Drain the 4 baseline entries, then flip the check to blocking in the pre-PR runner.

## Why

A warn-mode governance check binds nobody. Flipping to blocking makes model pins a contract. This is the v0.7.0 release trigger: first enforcement flip lands, tag v0.7.0 (breaking change, contributor-facing).

## Acceptance criteria

- model_pin_baseline.json empty and deleted
- pre_pr runner fails (not warns) on unpinned model references
- Breaking-change note drafted for v0.7.0 release notes

Refs #3071, #3227, ADR-080.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #3942: feat(validation): flip ADR-080 model-pin check from warn to 
- **Type**: complexity_triage
- **Priority**: high
- **Suggested Agent**: agent-architect
- Issue labeled with high complexity.
Suggested agents: agent-architect

### Route issue #3942 to agent-architect
- **Type**: agent_routing
- **Priority**: low
- **Suggested Agent**: agent-architect
- Issue content suggests routing to: agent-architect
Complexity: high

rjmurillo:
Closing as stale. Issue premise: 4 entries remaining in model_pin_baseline.json. Current measurement: 46 entries (frozen_count: 52). The baseline grew rather than drained - the precondition for flipping to blocking mode is not met. `python scripts/validation/check_model_pins.py --mode warn` shows 38 backlog items awaiting migration. This issue should be re-opened when the baseline actually reaches zero.

Issue #3938, opened 2026-07-30
pre_pr.py: documented python3 invocation fails two validations with ModuleNotFoundError


rjmurillo:
AGENTS.md documents the pre-PR gate as:

```
python3 scripts/validation/pre_pr.py
```

That exact command fails two validations on a correctly bootstrapped clone:

```
[FAIL] Skill Markdown Portability
[FAIL] Shipped Skill Routes
  ModuleNotFoundError: No module named 'markdown_it'
```

`uv run python scripts/validation/pre_pr.py` passes both. The module is present in the project venv (`markdown_it` 4.2.0) and absent from system Python.

## Why it propagates

`checks_spec.py` shells out with `sys.executable`. Launching the orchestrator with system `python3` makes every child validation inherit system Python, so a single wrong interpreter at the entry point silently degrades the whole suite.

## Why it matters

Two failures that are pure environment noise. An agent following the documented command sees red, and the cheapest reading is "my change broke it." That burns a debugging cycle per session, and the more expensive failure mode is learning to discount pre_pr output.

This is the same class as the existing guidance in `.github/copilot-instructions.md`: prefer `uv run python` for scripts importing project deps. The gate command in AGENTS.md never got the memo.

## Fix

Either:

1. Update the documented command to `uv run python scripts/validation/pre_pr.py`, or
2. Have `pre_pr.py` detect it is running outside the project venv and either re-exec under `uv run` or fail fast with that instruction.

Option 2 is better. Option 1 leaves the trap for anyone who types the old command from memory.

Both are cheap. The current state costs a cycle every time someone hits it.

rjmurillo:
Closing as duplicate of #3903. Both issues describe the same root cause: the documented `python3 scripts/validation/pre_pr.py` command fails outside the uv venv because markdown_it is absent from system Python. The fix is tracked in #3903.

Issue #3670, opened 2026-07-28
ADR review gate has no server-side backstop, so every local narrowing is defeatable by design


rjmurillo:
An adversarial security review of the ADR review gate returned three HIGH findings. One of them is a real logic gap and is fixed in PR #3669. The other two share a root cause that no amount of local narrowing can fix, and that root cause is an architecture decision rather than a bug.

## The root cause

`check_adr_review_policy` runs only as a lefthook client-side hook (`lefthook.yml`, the `adr-review-policy` step). Nothing enforces it in CI. Every input to its decision is author-controlled:

- `refs/remotes/origin/main` is a local ref. `git update-ref` rewrites it.
- `MERGE_HEAD` is a file inside the git directory.
- The index is author-controlled by definition.

The review agent demonstrated both. Pointing `origin/main` at an attacker commit exempts arbitrary ADR content. Writing a `MERGE_HEAD` by hand manufactures a merge that never happened.

## Why the fixes in #3669 are still worth having

PR #3669 rejects a `MERGE_HEAD` that HEAD already contains, and requires that the copy being replaced is one main has actually carried. Those close the paths that a careless author reaches by accident, which is what a client-side hook is for. They do not turn the hook into a security control, and the PR body does not claim they do.

## The decision this needs

An author who can run `git update-ref` can also run `git commit --no-verify`, so the honest framing is that this hook is a guardrail against accident and never a boundary against a motivated committer. Two options, and picking between them is an owner call:

1. Add a CI job that runs the ADR review policy against the PR base SHA from the GitHub API, where the base is not author-controlled. This makes the policy enforceable and makes the local hook a fast preview of the real gate.
2. Document that the policy is advisory, drop the pretence of enforcement, and stop hardening the local checks.

Doing neither leaves the repository in the current state, where the policy reads as enforced and is not.

## Provenance

Adversarial security review, `gpt-5.6-sol`, Findings 1 and 2. Finding 3 is fixed in #3669. All three were reproduced before being accepted; Finding 3 reproduced with no write to the git directory, which is why it was treated as in-scope for a code fix and these two were not.

Refs #3666

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #3670: ADR review gate has no server-side backstop, so every local 
- **Type**: complexity_triage
- **Priority**: high
- **Suggested Agent**: agent-security
- Issue labeled with high complexity.
[...]

PR #4022, opened 2026-07-30
fix(validation): remove fail-open paths and unsatisfiable gate conditions

### LINKED ISSUE #3712: PR description gate: reference-H2 exemption silences the CRITICAL but also blocks the significant-file WARNING
## Problem

`scripts/validation/pr_description.py` has two checks that pull in opposite directions, and satisfying one can make the other unsatisfiable.

1. A file path mentioned in inline code, bold, a list item, or a link, that is not in the diff, is a CRITICAL "file mentioned but not in diff".
2. A significant changed file that is never mentioned is a WARNING.

The documented way to silence (1) for a legitimate reference is to put the path under a contextual H2 such as `## References`, `## Evidence`, `## Notes`, or `## Out of Scope`. But `extract_mentioned_files` strips those sections before extraction. So a path under a reference H2 is invisible to (1) **and** to (2). If that path is also a changed file, the author has silenced the CRITICAL and simultaneously guaranteed the WARNING, with no way to satisfy both from one mention.

## Second sharp edge

Inline-code and list-item mentions register as "mentioned" only inside an H2 whose name matches one of four exact strings: `Changes`, `Per-file changes`, `Files Changed`, `Changed Files`. A perfectly reasonable `## What changed` or `## Summary of changes` silently does not count. The author gets a WARNING for a file they described in detail, and nothing in the message says the heading is why.

## Impact

Every PR in the current campaign has hit one side of this. The workaround, once you know it, is to use the literal H2 `## Changes` for the per-file narrative and a separate reference H2 for anything else. Nothing communicates that, and the failure text points at the file rather than the heading.

## Suggested fix

Three options, cheapest first.

1. Name the cause in the message. When a mentioned file is stripped by section filtering or falls outside a recognized H2, say so: "path mentioned under `## Notes`, which is excluded from extraction; move it under `## Changes` to satisfy the significant-file check."
2. Widen the recognized H2 set to a prefix or regex match on "change", so `## What changed` and `## Summary of changes` count.
3. Let a single mention satisfy both checks: extract from reference sections for the purposes of (2) while still exempting them from (1).

Option 1 alone removes most of the cost.

### LINKED ISSUE #3730: Portability ratchet never tightens: 8 refs of slack let removed upstream paths return with a green gate
## Problem

The vendor-portability ratchet in `scripts/validation/check_skill_portability.py`
only fails on regression. It never tightens on improvement, and nothing fails when
slack accumulates. Every ref of slack is a ref that can silently come back.

## Evidence

Current run on a clean checkout:

```text
Portability improved (tighten the baseline with --update-baseline):
  [IMPROVED] skills/memory-reflexion/tests/test_memory_reflexion_structure.py: 1 refs (baseline 2)
  [IMPROVED] skills/memory/scripts/test_memory_health.py: 2 refs (baseline 3)
  [IMPROVED] skills/memory/tests/test_memory_health.py: 7 refs (baseline 12)
No vendor-portability drift. 156 grandfathered refs across 82 scripts (baseline 164).
```

Exit code is 0. The advice to tighten is printed and nothing enforces it.

`test_memory_health.py` carries the largest gap: 7 actual against a baseline of 12.
Five upstream-path refs can be reintroduced there with a green gate. That is the
exact regression the ratchet exists to prevent.

Repo-wide the slack is 8 refs (156 actual, 164 baseline).

## Why it matters

A ratchet that only clicks one way stops being a ratchet once slack exists. The
work that removed those refs is unprotected: the next edit can undo it and the
gate stays green, so the improvement has to be re-done and re-reviewed later at
full cost.

This is the same failure shape the ADR-006 run-block counter avoids by being
lowered after each merge.

## Proposed fix

Fail when measured refs are below baseline by more than a permitted slack, with
the remedy printed (`--update-baseline`). That converts the existing advisory
line into a gate and keeps the baseline honest without blocking the improving PR:
the tightening is a one-command, zero-risk follow-up.

Alternative considered: auto-tighten inside the checker. Rejected, because a gate
that rewrites its own baseline during CI hides the change from review and makes
the committed baseline depend on execution order.

## Acceptance criteria

- [ ] The baseline is tightened to current actuals (156 across 82 scripts)
- [ ] Slack beyond a documented allowance fails the check
- [ ] The failure message names `--update-baseline` as the remedy
- [ ] A test covers improved, unchanged, and regressed cases

### LINKED ISSUE #3895: Push gate: a PR that is over the commit ceiling and labeled needs-split cannot receive its own review fixes
## Summary

A pull request that is both over the 40-commit push ceiling and labeled
`needs-split` cannot receive any further commit, including the commits that fix
its own review threads. The two escapes contradict each other, so there is no
move that does not defy one of them.

## Evidence

PR #3688 is the live instance.

```
$ git rev-list --count origin/fix/3673-3683-semgrep-gate ^origin/main
52
$ gh pr view 3688 --json labels -q '.labels[].name'
bug
area-infrastructure
automation
needs-split
```

Pushing two review-fix commits onto it:

```
$ git push origin fix/3673-3683-semgrep-gate
no commit-limit-bypass label (PR #3688)
ERROR: push has 54 commits, limit is 40
error: failed to push some refs
```

## Why the obvious workarounds do not work

**Count only the new commits.** The docstring on `_unpushed_commit_count` says
the branch's own remote ref is deliberately kept in the count, because
excluding it "would retire the ceiling entirely for any branch pushed more than
once". That is a deliberate decision, not an oversight. `_check_commit_limit`
counts `update.range_spec`, which measures against main, so a two-commit push
onto a 52-commit branch reports 54.

**Add the bypass label.** `commit-limit-bypass` is the only relief the gate
offers. Applying it to a PR the maintainer labeled `needs-split` overrides the
instruction the label carries.

**Ship the fixes against main instead.** Tried. Both commits conflict, because
they depend on code that exists only on the branch:

```
$ git cherry-pick 086733afd 2e4495465
CONFLICT (content): Merge conflict in scripts/validation/git_hook_policy.py
CONFLICT (content): Merge conflict in tests/test_lefthook_integration.py
```

The second commit fixes four tests that the branch itself added, so there is
nothing on main for it to fix.

**Stack a new branch on top.** Its commit count against main is also 54, so it
hits the same ceiling.

## Impact

Review feedback on a large PR cannot be addressed. The PR can only be split,
which is a restructuring of 52 commits, or abandoned. Neither is a review
response, so the review threads stay open indefinitely and the PR cannot merge.

## Suggested directions

Any one of these breaks the deadlock:

1. Treat `needs-split` as itself granting a narrow bypass, capped at a small
   number of commits, so review fixes can land while the split is planned.
2. Count only commits that no other remote ref carries when the destination
   branch already exists and the push adds fewer than some small number.
3. Let the ceiling admit a push whose new commits are all fixes to files the
   branch already touches, which is what a review response looks like.

Option 1 is the smallest and needs no new measurement.

## Not in scope

Whether PR #3688 should be split. It should; the label is right. This issue is
about the fact that a PR awaiting a split cannot answer its reviewers in the
meantime.

### LINKED ISSUE #3901: fix(ci): avoid commit-limit error annotations when bypass is active
## Problem

The Validate PR job emits an error annotation from `pr_commit_count.py` before the enforcement step honors the `commit-limit-bypass` label.

Evidence from PR #3677, run 30511147512:

```text
PR #3677 has 53 commits (status: BLOCKED)
##[error]PR exceeds commit limit (53 > 40). Consider splitting this PR.
...
##[warning]Commit limit bypassed via commit-limit-bypass label
PR validation passed
```

The job later failed for an unrelated taste-count regression. The GitHub check log extractor therefore surfaced the bypassed commit limit as an apparent failure alongside the real root cause.

## Expected

When `commit-limit-bypass` is active, the commit-count path should emit a warning or neutral diagnostic, not an error annotation or a BLOCKED status. The final enforcement decision should own the annotation severity.

## Impact

Humans and automated CI triage can select the wrong root cause, especially when a later step fails the same job.

### LINKED ISSUE #3902: feat(ci): show new taste-lint violations when the count ratchet fails
## Problem

`taste_count_ratchet.py` reports only the aggregate count and baseline. It does not identify which findings are new.

PR #3677 failed with:

```text
taste count ratchet: REGRESSION. 617 violations > baseline 615 (+2).
```

Finding the two regressions required a second main worktree, running the JSON linter against the same changed paths at base and head, and manually comparing the reports. The new findings were file-size errors in `scripts/eval/_run_persistence.py` and `scripts/validation/spec_contradiction.py`.

## Expected

On regression, the ratchet should print the new error-severity findings, including rule, path, and line. If an exact base-to-head delta is unavailable, listing error findings for changed files would still make the failure actionable.

## Impact

The current output makes a deterministic CI failure expensive to diagnose and encourages broad suppressions or guesswork. This is distinct from the absolute-baseline drift tracked in #3884: this issue is about failure diagnostics after a real count increase.

Found while repairing PR #3677.

### LINKED ISSUE #3903: fix(ci): make the documented pre_pr command use the project environment
## Problem

The canonical CI script rule tells contributors to run `python3 scripts/validation/pre_pr.py`. In a clean linked worktree, that command uses the system interpreter and two validators crash because `markdown_it` is unavailable.

## Reproduction

```text
python3 scripts/validation/pre_pr.py
```

Observed failures:

```text
Skill Markdown Portability: ModuleNotFoundError: No module named markdown_it
Shipped Skill Routes: ModuleNotFoundError: No module named markdown_it
```

Running the same gate through the locked project environment succeeds:

```text
uv run --frozen python3 scripts/validation/pre_pr.py
Total Validations: 41
Passed: 37
Failed: 0
Skipped: 4
```

## Expected

The documented canonical invocation should be reproducible from a clean worktree. Either update the rule and contributor guidance to use `uv run --frozen`, or make `pre_pr.py` bootstrap or validate its required environment before launching dependency-bearing checks.

## Impact

The current command reports false local gate failures and can send CI repair work toward unrelated validators. Found while repairing PR #3677.

### LINKED ISSUE #3920: PR commit-count gate treats main-merge maintenance as scope growth
## Symptom

PR #3780 reached 29 commits and received `needs-split`. Its final diff contained 2 files and 171 added lines for one nested-agent validation change. Before merge, 25 commits were merges from `origin/main`; only 4 were regular commits.

## Root cause

`scripts/validation/pr_commit_count.py::fetch_commit_count` fetches the pull request commits and assigns `count = len(commits)` without checking each commit's `parents` array. Branch-maintenance merges therefore consume the same threshold as authored changes.

The thresholds then classify the maintenance history as scope growth:

- Warning: 10 commits
- Alert: 15 commits
- Block and `needs-split`: 20 commits

For PR #3780, the effective authored count was 4, below every threshold. The raw count was 29, above the block threshold.

## Impact

The label lies about scope. A small PR on an active base branch appears oversized after repeated base refreshes, then follow-up gates can treat the false classification as authoritative.

## Distinct from #3895

Issue #3895 covers a deadlock after a genuinely oversized PR already has `needs-split`. This issue covers the earlier false classification: merge-only branch maintenance creates the oversized count.

## Acceptance criteria

- Classify a PR using non-merge commits, identified by at most one parent.
- Keep the raw total available in output for audit and diagnostics.
- Add tests for all-regular, all-merge, and mixed commit payloads.
- Prove PR #3780 classifies as 4 authored commits rather than 29 total commits.
- Re-evaluate and remove `needs-split` when the corrected count falls below the threshold.

## Evidence

- `origin/main..refs/pull/3780/head` before merge: 25 merge commits and 4 regular commits.
- `fetch_commit_count` currently returns `CountResult(classify_count(len(commits)), len(commits), transient=False)`.
- Existing issue searches for `pr_commit_count`, `merge commits` plus `needs-split`, and `commit-count heuristic` found no matching defect.

### LINKED ISSUE #3951: fix(validation): two gates enforce different always-on context budgets on the same files
## Summary

Two gates enforce the always-on context budget on the same files with different limits. A file can pass one and fail the other, so an author who validates with the looser gate is still blocked at push time with no earlier warning.

## The defect

`tests/test_workspace_limits.py` and `scripts/validate_workspace_budget.py` both bound the always-on instruction corpus. They disagree on every number, and on which files count.

| Enforcer | Per-file limit | Total limit | Files counted |
| --- | --- | --- | --- |
| `tests/test_workspace_limits.py` | 3072 | 6758 | `AGENTS.md`, `CLAUDE.md` |
| `scripts/validate_workspace_budget.py` | 3000 | 6600 | `AGENTS.md`, `CLAUDE.md`, `.claude/CLAUDE.md` |

The stricter one binds in practice, so the looser constants are dead allowance that reads as real headroom.

## How it was found

While cutting `AGENTS.md` back under budget, the file landed at 2995 bytes. Anything in the 3001 to 3072 range passes the pytest limit and fails the script. The 72-byte window between them is invisible until push.

## Impact

An author reading `tests/test_workspace_limits.py` sees 3072 bytes of allowance that does not exist. The failure surfaces at push time, after the work is committed, when the fix means finding something to displace.

## Suggested fix

Give the limits one home. Have the test import the constants from `scripts/validate_workspace_budget.py` rather than restating them, and reconcile the file list so both count the same set. The specific numbers matter less than the two enforcers agreeing.

Found while applying the context-engineering doctrine in `rjmurillo/shihipar-context-engineering`.

### LINKED ISSUE #3991: chore(budget): .github/copilot-instructions.md is always-on at 6351 bytes and no gate measures it
## Summary

`.github/copilot-instructions.md` is injected into every GitHub Copilot session, but no budget gate measures it. It currently sits at **6351 bytes**, which is **2.1x** the 3000-byte per-file ceiling this repository enforces on its Claude-side equivalents.

## Evidence

Two always-on gates exist, and neither covers this file.

`scripts/validate_workspace_budget.py`:

```python
PER_FILE_BUDGET_BYTES = 3000

WORKSPACE_FILES = [
    "CLAUDE.md",
    "AGENTS.md",
    ".claude/CLAUDE.md",
]
```

`scripts/validation/instruction_budget.py` measures only `.github/instructions/*.instructions.md`, per its own module docstring:

> every ``.github/instructions/*.instructions.md`` whose ``applyTo`` matches that language

So the Copilot entry point is measured by neither.

Measured sizes on `origin/main` at `32af5363`:

| File | Bytes | Gated by | Ceiling |
|---|---|---|---|
| `AGENTS.md` | 2999 | `validate_workspace_budget.py` | 3000 |
| `CLAUDE.md` | 2482 | `validate_workspace_budget.py` | 3000 |
| `.claude/CLAUDE.md` | 174 | `validate_workspace_budget.py` | 3000 |
| `.github/copilot-instructions.md` | **6351** | **nothing** | **none** |

Reproduce:

```
git show origin/main:.github/copilot-instructions.md | wc -c
grep -n "WORKSPACE_FILES" -A5 scripts/validate_workspace_budget.py
```

## Why this matters

The always-on budget work (#3419, #1335) rests on a premise: what is injected on every turn is measured, so it cannot grow silently. That premise holds for the Claude entry points and fails for the Copilot one. `AGENTS.md` currently has **one byte** of headroom against its ceiling, so contributors are pushed toward the unmeasured file whenever they need to add always-on guidance. The gate shape actively steers growth into the blind spot.

The concrete symptom, observed while working #3952: `.github/copilot-instructions.md` accumulated a `## Gotchas (non-obvious, save cycles)` section of roughly 2100 bytes. One of its entries (use `uv run python`, not bare `python3`) now duplicates `.agents/governance/GOTCHAS.md`, which is a retrieval-tier file created precisely so that detail is not paid for on every turn. Nothing flagged the growth or the duplication, because nothing was looking.

## Proposed fix

1. Add `.github/copilot-instructions.md` to `WORKSPACE_FILES` in `scripts/validate_workspace_budget.py`.
2. Seed its ceiling as a non-regression ratchet at the measured post-cleanup size rather than at 3000, so the gate lands green and the reduction is a separate, reviewable step. Record the seeded number and the reason in the same change, matching the convention `instruction_budget.py` already documents ("never raise one without recording why in the same change").
3. Follow up by moving progressive-disclosure content out to `.agents/governance/GOTCHAS.md` until the file can meet 3000.

Step 3 is partly addressed by the PR that accompanies this issue, which deduplicates the Gotchas section. This issue tracks the missing instrument, which is the part that lets the problem recur.

## Acceptance criteria

- [ ] `.github/copilot-instructions.md` is measured by a byte gate that runs in CI.
- [ ] The seeded ceiling and its rationale are recorded in the change that introduces it.
- [ ] A test asserts the file is in the gated set, so removing it from the list fails rather than silently disabling the gate.
- [ ] `.agents/governance/GOTCHAS.md` documents the new ceiling alongside the existing `AGENTS.md` byte-gate entry.

## Out of scope

Lowering the ceiling to 3000. That is content work, not instrument work, and it should be reviewed on its own.

## References

See `scripts/validate_workspace_budget.py`, `scripts/validation/instruction_budget.py`, and `.agents/governance/GOTCHAS.md`. Related: #3419 (always-on instruction budget instrument), #1335 (AGENTS.md over budget), #3910 (user-level instruction duplication).

rjmurillo:
## Summary

Fixes 8 issues covering fail-open paths, unsatisfiable gate conditions, and
missing validation coverage across the push-hook and CI validation layer.

### fix(validation): use authored commit count; downgrade BLOCKED to warning

`pr_commit_count.py` counted all commits including merge commits, which
inflated counts for branches merged from main. Added `_authored_commit_count()`
that counts only commits with at most one parent (excluding merge commits).
Added `total_count` field to `CountResult` for diagnostics. Changed the BLOCKED
annotation from `::error::` to `::warning::` because enforce_pr_validation.py
owns the final enforcement decision.

Fixes #3920
Fixes #3901

### fix(hooks): add needs-split bypass for incremental pushes on over-limit branches

The `needs-split` label signals a PR is already scheduled for splitting.
Before this fix the push gate blocked every push on such a PR regardless of
size, creating a deadlock: the label could never be applied without
`--no-verify` because the hook fired before the label could exist. Added
`_check_needs_split_bypass()` in `scripts/validation/git_hook_policy.py` which allows pushes of up to 5 new commits when
the PR carries the `needs-split` label.

Fixes #3895

### fix(validation): unify workspace-limit constants; tighten portability ratchet

`tests/test_workspace_limits.py` imported hardcoded constants instead of
reading them from the authoritative script, so the test and the enforcer could
silently diverge. Fixed by importing `PER_FILE_BUDGET_BYTES`,
`TOTAL_BUDGET_BYTES`, and `WORKSPACE_FILES` from
`scripts/validate_workspace_budget.py`.

`scripts/validation/check_skill_portability.py` returned success when the
portability baseline had slack (current count below baseline). That allowed
regressions to grow up to the slack without detection. Fixed: the script now
exits 1 when either regressions or improvements are found, so callers always
run `--update-baseline` when the baseline drifts.

Fixes #3951
Fixes #3730

### fix(docs): replace python3 with uv run python in pre_pr invocations

Several documentation files and instruction mirrors referenced bare `python3`
for pre-PR validation steps. On machines where `python` and `python3` do not
resolve to the project venv, those commands silently ran against the wrong
interpreter. Changed to `uv run python` in all six affected files.

Fixes #3903

### fix(validation): suppress WARNING for files mentioned in reference sections

`pr_description.py` emitted a WARNING for every file path mentioned in inline
code inside a description, even when the path appeared under a contextual
heading that signals intentional reference rather than a change claim. Added
`extract_all_mentioned_files()` which strips admonition-style blocks and
citation-cue prefixes but preserves reference-section paths, and wired it into
`validate_pr_description` as the suppression set for WARNING-level findings.

Fixes #3712

### fix(ci): print violation list on ratchet regression

When the taste-count ratchet or the analogous count ratchet regressed, callers
received only a one-line error count with no file list. Added a `lister`
callback parameter to `scripts/ci/count_ratchet.py` `run()` and implemented
`list_violations()` in `taste_count_ratchet.py` so the regression message is
followed by the offending paths.

Fixes #3902

## Issues closed without code change (with measurement)

- #3670: Closed. The architecture decision to use `origin/main` as the push
  base is intentional; the worktree constraint is a design property, not a
  defect.
- #3938: Closed as duplicate of #3903.
- #3942: Closed. Measured 46 baseline entries, not 4. The stale premise in the
  issue text did not match the repository state.
- #3943: Closed. `_return_bindings` and the associated tests were already
  implemented and passing before this campaign began.

### `_check_commit_limit` with needs-split bypass

Block (over cap): push of 6 new commits on `needs-split` branch returns 1.
Permit (within cap): push of 5 new commits on `needs-split` branch returns 0.
Permit (no label): push that is over total limit but carries
`commit-limit-bypass` label returns 0. These are covered by the four tests
added to `tests/test_lefthook_integration.py`.

### `_authored_commit_count` in `pr_commit_count.py`

Block: merge commit with two parents is excluded from the authored count.
Permit: regular commit with one parent is counted. Covered by parametrized
tests in `tests/validation/test_pr_commit_count.py`.

### Portability ratchet in `check_skill_portability.py`

Block: exits 1 when current count is below baseline (slack detected).
Permit: exits 0 when current count equals baseline. Covered by
`tests/validation/test_check_skill_portability.py`.

## Test coverage

8982 tests pass (targeted run with `-k "valid or gate or hook or lefthook or pr_"`).
Full suite (18686 passed, 30 skipped) passes on push.

## Mutation harness

7/7 mutants killed across the 6 production files changed in this branch.

## References

Issue cluster: #3670 #3712 #3730 #3895 #3901 #3902 #3903 #3920 #3938
#3942 #3943 #3951

### fix(budget): gate .github/copilot-instructions.md with non-regression ratchet

`.github/copilot-instructions.md` is injected into every Copilot session but
was unmeasured by either budget gate. Gate 1
(`scripts/validate_workspace_budget.py`) covers Claude-side workspace files.
Gate 2 (`scripts/validation/instruction_budget.py`) covers
`.github/instructions/*.instructions.md` per language extension. Neither
covered the Copilot entry point.

Added `FILE_CEILING_BYTES` dict with a non-regression ratchet for
`.github/copilot-instructions.md` at 6351 bytes (measured on main at time of
fix). Ratcheted files use their own ceiling and are excluded from the shared
`TOTAL_BUDGET_BYTES` pool (which was designed for the Claude-side trio only).
Created `.agents/governance/GOTCHAS.md` documenting all byte-gate ceilings.

Block: file over its individual ceiling fails `validate_budget()`.
Permit: file at or below its ceiling passes.
Negative control: `test_workspace_files_nonempty` guards against an empty
list that would make all per-file assertions vacuously true.

5/5 mutants killed.

Fixes #3991


rjmurillo:
Verified this PR's fix and then swept the rest of the tree. The fix is correct
where it lands, and there is a measured remainder that it does not cover.

## What this PR fixes, confirmed

`AGENTS.md`, `.agents/SESSION-PROTOCOL.md`, `.github/PULL_REQUEST_TEMPLATE.md`,
`.claude/rules/ci-scripts.md` MUST-2, and the two rule mirrors. Reproduced the
underlying failure independently on this branch:

```
python3 scripts/validation/pre_pr.py --quick
  [FAIL] Skill Markdown Portability   ModuleNotFoundError: No module named 'markdown_it'
  [FAIL] Shipped Skill Routes         ModuleNotFoundError: No module named 'markdown_it'
  RESULT: 2 validation(s) failed

uv run python scripts/validation/pre_pr.py --quick
  RESULT: All validations passed
```

Root cause confirmed: `markdown-it-py>=4.2.0` is a declared dependency in
`pyproject.toml:12`, so the gate needs the project environment. `lefthook.yml`
already knew this (lines 37, 46, 370 all use `uv run --frozen python`), which
means only the human-facing docs were wrong.

## Remainder: 12 occurrences in 10 shipped-plugin files

Identical on `main` and at this PR's head, so this PR does not touch them:

```
git grep -c "python3 scripts/validation/pre_pr.py" <ref> -- '.claude/skills/*' 'src/copilot-cli/skills/*'

.claude/skills/ai-agents-change-control/SKILL.md:2
.claude/skills/ai-agents-change-control/references/gate-ladder.md:1
.claude/skills/ai-agents-debugging-playbook/SKILL.md:1
.claude/skills/ai-agents-empirical-probe-toolkit/SKILL.md:1
.claude/skills/ai-agents-validation-and-qa/SKILL.md:1
src/copilot-cli/skills/... (the 5 generated mirrors, same counts)
```

All five canonical sites are prescriptive, not descriptive. Two are checklist
items the reader is told to tick (`ai-agents-validation-and-qa/SKILL.md:136`,
`ai-agents-change-control/SKILL.md:120`), one is the rung 1 row of the gate
ladder, one is the "full shift-left sweep" step in the debugging playbook, and
one is a bisect instruction in the probe toolkit. These ship in the plugin, so
they reach users who never see the repo docs this PR fixed.

I deliberately excluded the rest. At this PR's head, 22 markdown files still
contain the string: the 10 shipped-plugin files above, 11 historical markdown
files under `.agents/analysis/`, `.agents/archive/`, and `.agents/specs/`, and
one Serena status memory. A further 16 non-markdown files under `.agents/`
(session logs and memory episodes) also contain it. Everything outside the 10
is a record of what was run at the time, not an instruction, so rewriting them
would falsify history rather than fix a defect.

## Why I am not opening a PR for the remainder

Three open PRs are already in those files, and one of them edits the exact
lines:

| PR | Overlap | Effect on those lines |
|----|---------|----------------------|
| #3993 | `ai-agents-change-control/SKILL.md`, `gate-ladder.md` | Rewrites the commit-threshold text on the same lines, leaves `python3` in place |
| #4037 | `ai-agents-change-control/SKILL.md`, `ai-agents-debugging-playbook/SKILL.md` | Does not change the interpreter |
| #3979 | `ai-agents-change-control/SKILL.md`, `ai-agents-validation-and-qa/SKILL.md` | Does not change the interpreter |

A fourth PR editing the same lines would conflict with all three. Happy to take
it as a follow-up once #3993 and #4037 land, or you can fold the sed into
whichever lands last. Byte headroom is not a concern: the largest of the four
skills is 18,817 bytes against the 24,576 limit, and the change adds 8 bytes per
occurrence.

## One judgment call, flagging rather than asserting

`.claude/rules/ci-scripts.md` MUST-1 still reads "direct `python3` invocation
for validation scripts", one line above the MUST-2 this PR fixed. Read as "run
it directly rather than through CI" it is fine. Read as naming the interpreter
it now contradicts MUST-2. I did not touch it because you own that file in this
PR and because the correct reading is genuinely ambiguous. Your call.

Refs #3903

rjmurillo:
## Adversarial review: the PR description gate will block this

The body claims `scripts/validate_workspace_budget.py` under `## Changes`, but that path is not in the diff.

The validator treats a backticked path under a change-claim heading as an assertion that the file was modified, and reports CRITICAL when it is absent. See `scripts/validation/pr_description.py`.

Fix by either removing the claim or moving it under a recognized reference heading (`## References`, `## Evidence`, `## Notes`, `## Out of Scope`), or prefixing it with a citation cue such as `see `path``.

Editing the body re-triggers the gate; no new commit is needed.

Auto-merge is disabled on this PR.

Issue #4035, opened 2026-07-30
Escaped pipes silently break re-verify commands in ai-agents skills, and four rows state facts their own command disproves


rjmurillo:
## Summary

Two defect classes in the `ai-agents-*` knowledge pack's "Fact / Source / Re-verify" tables. Both make a row look verified when it is not.

## 1. Escaped pipes silently break re-verify commands

A re-verify command lives in a Markdown table cell, so its pipe must be written `\|`. GitHub's renderer strips that backslash at table-parse time (the pipe is the cell delimiter), which changes the command's meaning.

The branch's original assumption was that a shell pipe is safe to escape because a mis-paste fails loudly. That holds only when the command stands alone:

```text
ls DIR \| wc -l                    rc 2   fails loudly
ls DIR \| wc -l; grep -c X f       rc 0   failure masked
```

In a semicolon-chained row the failure is swallowed by the last command's exit code, so the row appears to pass while producing garbage. No escaped pipe is acceptable in a re-verify command, regex or shell.

One affected row was an absence proof (`expect no matches`). On a fixture where the regression is present, the raw form matched but the rendered form reported the absence satisfied. A guard that can never fire.

Pipe-free replacements, each verified to return the same value as the piped original:

```text
count files      set -- DIR/*; echo $#
filter listing   find DIR -maxdepth 1 -name "*pat*"
grep a file set  git grep -n PAT -- '*glob*'
```

## 2. Four rows state a Fact their own command disproves

| Claim in the table | What the command prints |
|---|---|
| Investigation allowlist has 9 patterns | 8 |
| ADR-034 text lags at 5 patterns | ADR-034 lists the same 8 (reconciled in #2958) |
| ~1471 commits, retro-cited SHAs unresolvable | 2014 commits; `ddb76e0` and `01e76615a` both resolve |
| ADR-068 status is Proposed | `status: accepted` |

ADR-069 and ADR-072 really are Proposed and are correct as written.

The maintenance rule in these tables requires fixing the skill and its reference together, so prose carriers in the same skills move with the rows.

## Notes

The `~1471 commits` claim also appears as prose in five skills outside this scope. That is a separate coherent unit and will land on its own.

Worth considering: a baseline-ratchet validator modeled on the existing skill-markdown portability check would prevent both classes from recurring.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

[...]

PR #4059, opened 2026-07-30
fix(skills): replace a falsified absence proof with the merge policy that makes it permanent

rjmurillo:
## Summary

Nine skill files under `.claude/skills/` are documents of record: they tell an agent how to verify a fact by running a command. Four of them lied. This PR makes every documented command produce the output the document claims, and replaces the one claim that cannot be kept fresh with a mechanism that does not decay.

Refs #4035

## The defect that mattered

`.claude/skills/ai-agents-failure-archaeology/SKILL.md` told readers to check two retro-cited SHAs and expect `Not a valid object name` as proof the commits were gone.

Executed it. Both SHAs resolve:

```bash
git cat-file -t ddb76e0    # commit
git cat-file -t 01e76615a  # commit
```

They resolve because this clone carries extra refs (`refs/remotes/pr/43` for the first; a stale local branch and two fork remotes for the second). Neither is reachable from `main`.

So the documented expectation was the exact inverse of reality. An agent following the instruction would have concluded the provenance claim had been falsified, and would have been wrong. This is the worst failure mode in this corpus: the document does not merely go stale, it actively teaches the opposite of the truth.

The root cause is that it was an **absence proof**. A check whose expected result is "not found" silently inverts the moment the thing appears, and resolvability is clone-state dependent.

## The replacement is a policy, not a measurement

Reachability from `main` is not clone-state dependent, and there is a permanent reason for it:

```bash
gh api repos/rjmurillo/ai-agents -q '.allow_squash_merge, .allow_merge_commit, .allow_rebase_merge'
# true, false, false
```

Squash is the only merge method GitHub permits on this repo. Every PR lands as one new commit with a new SHA, so a commit that only ever existed on a PR branch is never reachable from `main`. That is a repository setting, not a snapshot, which is why it correctly replaces the `~1471 commits` anchor the doc used to carry (real count on origin/main is now 2015).

Both re-verify forms are documented, and each confirms the conclusion whether or not the object happens to be present locally:

```bash
git merge-base --is-ancestor ddb76e0 origin/main; echo $?
# 1                                    when the object is present
# fatal: Not a valid object name / 128 when it is absent
```

Deliberately **not** claiming "zero merge commits on main". That is false. There is exactly one (`0f13c85ab`, "Merge pull request #1", 2025-12-13), which predates the current setting. The claim in the diff is the setting, which is true without qualification.

## Stale counts, re-measured and date-stamped

| Claim | Was | Now | Command |
|---|---|---|---|
| retrospective files | 95 | 121 | `set -- .agents/retrospective/*.md; echo $#` |
| index data rows | 5 | 9 | `grep -c '^. 20' .agents/retrospective/INDEX.md` |
| memory-tier entries | 122 (undated) | 158 as of 2026-07-30 | `set -- .serena/memories/*; echo $#` |
| commits on main | ~1471 | removed, replaced by merge policy | see above |

The index-row figure is the load-bearing one: `.claude/skills/ai-agents-docs-of-record/SKILL.md` uses it to argue the index has fallen behind the corpus. That gap is real (9 rows against 121 files), but both recorded numbers were wrong, so a reader who re-measured would have found neither matched and discarded the argument along with the error.

Counts that had no mechanism available are now stamped with a measurement date, so the next reader can distinguish decay from error.

## One uncopyable command

`.claude/skills/ai-agents-research-methodology/SKILL.md` documented a grep whose alternation pipe was escaped as `\|` to survive a Markdown table cell. Agents read raw Markdown, so the backslash reached the shell verbatim and the command failed on paste. Rewritten to a form that needs no pipe escaping, so the copied text is the executed text.

Scope note: `\|` is only a defect inside a table cell. Two remaining instances in `ai-agents-research-frontier` sit in bullets, where the renderer preserves the backslash and it is the correct BRE alternation operator. Left alone deliberately.

## Verification

Every command that appears in the diff was executed, not cited:

```
gh api repos/rjmurillo/ai-agents -q .allow_merge_commit   -> false
git merge-base --is-ancestor ddb76e0 origin/main; echo $? -> 1
set -- .agents/retrospective/*.md; echo $#                -> 121
set -- .serena/memories/*; echo $#                        -> 158
grep -c '^. 20' .agents/retrospective/INDEX.md            -> 9
```

Also checked:

- Zero em-dash or en-dash bytes across all 20 changed files, verified at byte level rather than visually.
- All nine Copilot mirrors regenerated with the generator and confirmed byte-identical to source. No mirror was hand-edited.
- Full pre-push suite green, including 452s of python tests.
- Manifests bumped to `0.6.5440`, strictly greater than main and all three sibling PRs in flight, so the version gate holds regardless of merge order.

## Stacked PR

Base is `docs/dead-verify-commands` (#4037), not main. That PR rewrote the exact table hunk containing the stale commit-count row and deliberately left that one row to this PR, per the deferral recorded in #4035. Branching from main would conflict there with certainty. Merge #4037 first.


rjmurillo:
## Adversarial review round 2: one finding disputed, then narrowed

Round 2 (gemini-3.1-pro-preview) re-verified all nine claims in this PR. Eight survived. It falsified CLAIM 3 (relation-not-snapshot) and returned REQUEST CHANGES, proposing that every hardcoded number in the diff be replaced with a relation phrase.

I disputed the severity and the remedy rather than complying. The reviewer conceded all three objections. Recording the argument because the conclusion is a reusable rule, not a one-off.

### Objection 1: the proposed remedy destroys the sentences it targets

The reviewer wanted `9 index rows vs 121 retro files as of 2026-07-30` replaced with `the number of indexed rows falls short of the retro files count`.

The magnitude is the claim. `9 vs 121` says INDEX.md covers under 8 percent of the corpus, which is exactly why the surrounding text tells the reader not to trust it as complete history. "Falls short of" is equally satisfied by 119 vs 121, which would make the warning wrong.

Conceded.

### Objection 2: several cited instances already are the relation form

Two of the flagged lines pair a dated magnitude with the shell command that reproduces it:

```
| Retro corpus size (121 files as of 2026-07-30) | .agents/retrospective/ | set -- .agents/retrospective/*.md; echo $# |
```

That is strictly better than either a bare number or a bare relation: it establishes scale and mechanism at once.

Conceded.

### Objection 3: severity

There are three grades, not two.

1. Bare unqualified count claiming to be current. Broken, because nothing signals it has rotted.
2. Dated count. Honest. Loses usefulness over time but never lies, because the date bounds the claim.
3. Dated count plus the command that reproduces it. Durable.

This PR moves text from grade 1 to grades 2 and 3. Labelling grade 2 as REQUIRED blocks an improvement in order to demand a standard no file in the repository currently meets universally.

Conceded. Downgraded to OPTIONAL. Verdict flipped to APPROVE.

### The part of the finding that was real

I asked the reviewer to separate counts whose magnitude carries the argument from counts that are decoration, meaning the sentence survives unchanged with the number deleted. It returned four, and I verified each against the file rather than trusting the list. Its line numbers were wrong in all four cases; the content claims were right.

Removed in the follow-up commit:

- `ai-agents-architecture-contract`: the ADR-007 row argues retrieval precedence. Memory-directory size is irrelevant to that rule.
- `ai-agents-build-and-env`: the MCP row argues that memories are plain readable files rather than an LSP-gated store. Count does not support that structural claim, and the same file already carries it in a provenance row with the verify command.
- `ai-agents-failure-archaeology`: the paragraph directs the reader to two locations instead of `git log`. The instruction operates identically at any scale, and lines 169-170 already carry both counts with their commands.
- `ai-agents-research-methodology`: "richest vein" is a superlative with nothing quantified to compare against, and the provenance table below already carries the count with its command.

Kept: 15 instances where the magnitude is the argument, the date bounds a status claim that changes, or the row pairs the count with its command.

### The rule this settles

Remove a count when the surrounding argument survives its deletion. Keep it when the number is doing the work, and prefer pairing it with the command that reproduces it. A blanket "no hardcoded numbers" policy damages the text it is meant to protect.

Issue #4054, opened 2026-07-30
pr-comment-responder: comment-map status vocabulary is undefined and the two completion gates disagree


rjmurillo:
## Summary

The pr-comment-responder comment map has no defined status vocabulary. Two independent gate families enumerate statuses by hand, they enumerate different sets, and they fail in opposite directions. On real historical data they return contradictory answers about whether the same PR is complete.

## The template defines no legend

The template mentions exactly four comment statuses, and never lists them as a vocabulary:

```
$ grep -ohE '\[(NEW|ACKNOWLEDGED|COMPLETE|WONTFIX|DEFERRED|DUPLICATE)\]' templates/agents/pr-comment-responder.shared.md | sort | uniq -c
      1 [ACKNOWLEDGED]
      2 [COMPLETE]
      1 [NEW]
      1 [WONTFIX]

$ grep -c -iE 'valid status|status.*(values|legend|vocabulary)' templates/agents/pr-comment-responder.shared.md
0
```

`[DEFERRED]` and `[DUPLICATE]` do not appear in the template at all. They exist only in real comment maps, invented at authoring time.

## Two gate families, opposite failure modes

**Family A** counts pending by enumerating pending statuses. Anything not on the list is silently treated as done, so an unknown status fails open.

**Family B** counts addressed by enumerating terminal statuses. Anything not on the list is treated as outstanding, so an unknown status fails closed and can never clear.

## They disagree on real data

`.agents/pr-comments/PR-225/comments.md` at `c0c8698e2` declares 7 total comments:

```
$ git show c0c8698e2:.agents/pr-comments/PR-225/comments.md | grep -oE '^\*\*Status\*\*: .*' | sort | uniq -c
      4 **Status**: [COMPLETE]
      1 **Status**: [DEFERRED]
      1 **Status**: [DUPLICATE]
      1 **Status**: [WONTFIX]
```

Family A reports zero pending, so the blocking gate passes:

```
$ git show c0c8698e2:.agents/pr-comments/PR-225/comments.md | grep -Ec "^\*\*Status\*\*: \[ACKNOWLEDGED\]|^\*\*Status\*\*: pending|^\*\*Status\*\*: \[NEW\]"
0
```

Family B reports 5 of 7 addressed, so the same run warns that 2 comments remain:

```
ADDRESSED=4 WONTFIX=1 SUM=5   (declared total 7)
```

[...]

PR #4062, opened 2026-07-31
fix(session-log-fixer): make the diagnostic command table actually runnable

### LINKED ISSUE #4064: session-log-fixer: 3 of 6 rows in the Diagnostic Command Reference cannot be run as written
## Summary

The `Diagnostic Command Reference` table in the `session-log-fixer` skill holds six commands. Three cannot be run as written, and three are elided with `...`. The table is the skill's only quick-reference for triaging a failed CI run, so an agent that copies a row gets a command that fails.

`.claude/skills/session-log-fixer/references/ci-debugging-patterns.md` (and its generated mirror under `src/copilot-cli/`).

## Root cause

The commands live in Markdown table cells. A bare `|` would split the cell, so each pipe was written `\|`. Agents read skill files raw, not rendered, so the backslash travels with the copied command.

```
$ git show origin/main:.claude/skills/session-log-fixer/references/ci-debugging-patterns.md | sed -n '90,96p'
| Purpose | Command |
|---------|---------|
| Run overview | `gh run view $RunId` |
| All jobs status | `gh api .../actions/runs/$RunId/jobs --jq '.jobs[]...'` |
| Failed jobs only | `...--jq '.jobs[] \| select(.conclusion == "failure")'` |
| Incomplete jobs | `...--jq '.jobs[] \| select(.status != "completed")'` |
| Workflow logs | `gh run view $RunId --log 2>&1 \| head -100` |
| Specific job log | `gh run view $RunId --log --job $JobId` |

escaped-pipe rows: 3
elided rows: 3
```

## Reproduction

Each failure mode executed, not inferred.

**1. Escaped pipe inside a jq program is a syntax error.**

```
$ jq -n '.jobs[] \| select(.conclusion == "failure")'
jq: error: syntax error, unexpected INVALID_CHARACTER, expecting end of file (Unix shell quoting issues?) at <top-level>, line 1:
.jobs[] \| select(.conclusion == "failure")
jq: 1 compile error
```

**2. Escaped pipe at shell level becomes a literal argument, not a pipe.**

```
$ gh api /repos/rjmurillo/ai-agents/actions/runs/1/jobs \| jq unique
accepts 1 arg(s), received 4
```

The shell passes `|`, `jq`, and `unique` to `gh` as three extra positional arguments. The command never pipes.

**3. The failure is not always loud.** In a semicolon-chained sequence the exit code of the last command masks it, which is the mechanism already documented in #4035.

## Not a blanket rule

`\|` in a table cell is not always wrong. In a `grep` basic regular expression it is the correct alternation operator, and the same skill uses it that way correctly:

```
grep -c "Status: \[COMPLETE\]\|\[WONTFIX\]" <file>   -> 2   (works raw)
grep -c "Status: [COMPLETE]|[WONTFIX]"     <file>   -> 0   (broken)
```

Only the shell-pipe and jq-pipe uses are defects. A rule that rewrites every `\|` would break working commands.

## Suggested remedy

Move the block out of the table into a fenced block. Escaping is required only because the commands sit in table cells, so a fence removes the requirement structurally rather than by convention, and the defect cannot be reintroduced by a later edit. The file already contains eight fenced `powershell` blocks, so this matches its own convention. Expand the three `...` elisions to full commands while the rows move.

## Relationship to #4035

#4035 covers the same defect class in the `ai-agents-*` knowledge pack's `Fact / Source / Re-verify` tables, and is being addressed by #4037 and #4059. `session-log-fixer` is not an `ai-agents-*` skill and is outside that issue's scope, so this is tracked separately rather than folded in.

rjmurillo:
### What

Three of the six commands in the `session-log-fixer` skill's `Diagnostic Command Reference` could not be run as written. They sat in a Markdown table, so their pipes were escaped as `\|` to avoid splitting the cell. Agents read skill files raw, not rendered, so an agent that copies a row gets the backslash. Three rows were also elided with `...`, so even unescaped they were not copyable.

The table is replaced with a fenced `powershell` block holding six complete, runnable commands.

### Why

The table is the skill's quick reference for triaging a failed CI run. A command that fails on paste sends the agent debugging its own tooling instead of the run.

## Specification References

| **Issue** | Fixes #4064 | session-log-fixer: 3 of 6 rows in the Diagnostic Command Reference cannot be run as written |

## The defect

Executed each form to confirm it is broken rather than merely ugly.

| Raw form | Result |
| --- | --- |
| `gh api ... \| jq unique` | `accepts 1 arg(s), received 4` |
| `jq '.jobs[] \| select(...)'` | `syntax error, unexpected INVALID_CHARACTER` |

In the first case the shell hands `|`, `jq`, and `unique` to `gh` as three extra positional arguments, so the command never pipes.

## Why a fenced block, not unescaped pipes

Escaping was required only because the commands sat in table cells. A fenced block removes that requirement structurally, so the same defect cannot be reintroduced by a later edit. The file already contains eight fenced `powershell` blocks, so this matches its own convention rather than introducing a new one.

## Two more commands in the same class

Two reviewers found the first commit's sweep incomplete. Both were right that it was incomplete, and both were wrong about how much. The census below is my own, executed rather than inherited.

Two further re-verify commands could not run as written:

| File | Raw form | Result |
| --- | --- | --- |
| `.claude/skills/ai-agents-change-control/references/provenance.md` | `ls ... \| grep -E "908\|1187..."` | `ls: invalid option -- 'E'` |
| `.claude/skills/ai-agents-generation-and-release/SKILL.md` | `ls ... \| grep drift` | pipe never opens |

These get pipe-free equivalents rather than fenced blocks. Both rows sit in twelve-row provenance tables where breaking one row out costs more comprehension than the escape does. Executed in the worktree, the replacements return exactly the files their rows name: five incident retrospectives for the first, one drift retrospective for the second.

## Not a blanket rule

`\|` inside a Markdown table cell is not always wrong. In a `grep` basic regular expression it is the correct alternation operator. Verified:

```
grep -c "Status: \[COMPLETE\]\|\[WONTFIX\]" <file>   -> 2   (works raw)
grep -c "Status: [COMPLETE]|[WONTFIX]"     <file>   -> 0   (broken)
```

An earlier pass classified five sites as defects. Executing each one reduced that to three. The two `grep` sites were left alone because "fixing" them would have broken working commands.

## Scope of the sweep

This PR covers `.claude/skills/` only. That is a deliberate boundary, not an oversight, and the claim is narrow on purpose because the first version of it was too broad and got falsified.

A census across the agent, skill, rule, and memory trees, with the four stale checkout copies excluded, found every escaped pipe sitting inside a fenced code block was already correct: eleven `grep` alternations, one PowerShell `-replace` regex, one deliberate `grep -P` literal, one Python raw-string regex. Zero leaked escapes. The defect exists only in table cells, where the escape is required Markdown, so a pattern match on the escape alone overcounts badly.

## Out of Scope

Deliberately untouched, each checked by executing or reading it rather than by pattern:

- `.serena/memories/jq/jq-quick-reference.md` is an operator reference whose first column documents the pipe character itself. The examples illustrate syntax; they are not commands the reader is told to run. Fencing them would dismantle the table to fix a non-defect.
- `.serena/memories/powershell/powershell-observations.md` carries its escape in a dated learning-log row, not a command.
- `.serena/memories/quality/quality-test-criteria-patterns.md` has the same defect and is fixed in #4063, which owns the PowerShell tree. It converts cleanly because its five rows are a procedural list whose column labels survive as comments. The jq reference above does not, because a three column operator lookup has no fenced equivalent.
- `.claude/skills/pr-comment-responder/references/gates.md` was flagged by both reviewers. Its row is descriptive prose containing a literal `...` elision, so it is not runnable with or without the escape. It is a different problem from this one.

## Changes

- Replace the `Diagnostic Command Reference` table with a fenced `powershell` block, expanding the three `...` elisions to full commands.
- Project `{id: .id, name: .name}` in the two `select(...)` rows. The block ends with `gh run view $RunId --log --job $JobId`, and the pre-conversion rows emitted whole job objects, so the id was available. A first pass projected to `| .name`, which silently removed the block's only source of `$JobId`; this restores it.
- Regenerate the Copilot mirror and bump both plugin manifests.

## Type of Change

- [x] Documentation update

## Reviewer Expectations

**Review kind / scrutiny / urgency**: targeted, standard, no rush

**Focus areas**: whether each replacement command is faithful to the row it replaces, and whether any remaining `\|` in the skill is a defect rather than correct regex alternation.

## Testing

- [x] No testing required (documentation only)

Evidence:

- All three `jq` programs run against a realistic payload:
  `echo '{"jobs":[{"id":123,"name":"build","status":"completed","conclusion":"failure"}]}' | jq -c '<program>'` -> `{"id":123,"name":"build"}`.
- Copilot mirror regenerated with `build/scripts/generate_skills.py`: 612 files written, zero drift, and the two copies are byte-identical.
- Byte-checked for U+2014 and U+2013 across all changed files: 0.
- Both plugin manifests bumped to the same strictly-greater version (0.6.5442, main is at 0.6.5437).
- Ran the repo pre-PR gate (`scripts/validation/pre_pr.py`): 41 gates, all pass.
- Targeted tests: 1625 passed (plugin manifest, version bump, build scripts, dogfood), 319 passed (portability), 6 passed (session-log-fixer skill tests).

### Security Review

### Other Agent Reviews

- [x] Independent code review completed. It found no blocking defect and caught the `$JobId` projection gap now fixed in 35289acbe.

## Notes for Reviewers

The original closing-keyword linkage on this PR pointed at #4054, which was wrong: that issue tracks the `pr-comment-responder` status vocabulary and shares no scope with this change. The AI spec-coverage gate caught it and failed the PR. The correct issue, #4064, was filed for this defect after confirming no existing issue covered it. #4035 covers the same defect class in the `ai-agents-*` knowledge pack and is being addressed by #4037 and #4059; `session-log-fixer` is not an `ai-agents-*` skill and is outside that issue's scope.


rjmurillo:
Reverted 8c4c07e3 in 06ce2199. Flagging it here because it was a deliberate commit, not an accident, and the call is reversible if you disagree.

Three reasons, in order of weight.

**1. #4037 already rewrote the same two lines.** Both PRs edited the identical rows in the identical files:

```
.claude/skills/ai-agents-change-control/references/provenance.md
.claude/skills/ai-agents-generation-and-release/SKILL.md
```

Whichever landed second would have conflicted. #4037 carries `Fixes #4035`, so that scope is already spoken for.

**2. #4037 picked the portable form.** Brace expansion is a bash extension, so the reverted replacement does not run under POSIX sh:

```
$ bash -c 'ls .agents/retrospective/*{908,1187,1887,1965,2205}*'
.agents/retrospective/2026-01-15-pr-908-comprehensive-retrospective.md
... (5 files, rc 0)

$ sh -c 'ls .agents/retrospective/*{908,1187,1887,1965,2205}*'
ls: cannot access '.agents/retrospective/*{908,1187,1887,1965,2205}*': No such file or directory
```

#4037 uses `find .agents/retrospective -maxdepth 1 \( -name "*908*" -o ... \)`, which runs in both. For a PR whose premise is that a documented command must run as written, shipping a shell-specific replacement would have undercut the argument.

**3. Scope drives the spec gate.** This PR closes #4064, which covers the session-log-fixer diagnostic table only. The `ai-agents-*` rows belong to #4035. The gate scores the diff against the linked issue, and mixing the two is what produced the original `VERDICT: FAIL` on this PR when it was linked to #4054.

Nothing is lost: both rows land through #4037 in the more portable form.

The plugin version is 0.6.5444 rather than back at 0.6.5442, so the branch stays monotonic across the revert.

Issue #4079, opened 2026-07-31
Prohibited Unicode dashes remain in 33 shipped plugin files (378 occurrences)


rjmurillo:
`.claude/rules/universal.md` MUST NOT item 5 bans U+2014 and U+2013 in all authored text. The `staged-dash-policy` pre-commit guard enforces it correctly, but it only inspects staged files, so it never retroactively cleaned what predates it. PR #4067 cleared `.claude/agents/` completely. The rest of the plugin trees still carry the defect.

## Census

Text files only. Binary assets are excluded: a PNG can contain the UTF-8 byte sequence of an em dash by coincidence, and an earlier census that counted three image files was wrong for that reason.

- 33 files
- 378 occurrences

The files pair up: most `.claude/skills/X` entries have a `src/copilot-cli/skills/X` twin, because the latter is generated from the former. Fixing the source and regenerating clears both, so the real work is roughly 17 source files.

Reproduce:

```
git ls-files '.claude/*' '.serena/memories/*' 'src/claude/*' '.github/agents/*' 'src/copilot-cli/*' | while read f; do python3 - "$f" <<'PY'
import sys, pathlib
p = pathlib.Path(sys.argv[1])
try:
    t = p.read_text(encoding="utf-8")
except (UnicodeDecodeError, OSError):
    sys.exit(0)
n = sum(t.count(c) for c in ("\u2014", "\u2013"))
if n:
    print(f"{n}\t{p}")
PY
done | sort -rn
```

## Concentration

Two files hold 141 of the 378, both under `prompt-engineer/references/`. They read as vendored or externally-sourced reference material rather than repo-authored prose, which is a different judgement call than the rest: a verbatim quote of an external document arguably should not be silently rewritten. That question should be settled before the sweep runs, not during it.

## Why this is not one PR

A blind replacement damages text. In PR #4067 one em dash separated two parallel contrast pairs, and replacing it with a comma flattened them into an ambiguous four-item list. Each occurrence needs the surrounding sentence read. At 378 occurrences that is not a mechanical job, and batching it into a single change would produce a diff nobody can review.

Suggested split, one PR each:

1. `.serena/memories/` (repo-authored, lowest risk)
2. `.claude/skills/` excluding `prompt-engineer/`, plus regenerated mirrors
3. `prompt-engineer/references/`, after deciding the vendored-text question

## Guard note

[...]

PR #4088, opened 2026-07-31
fix(skills): make retro-path re-verify command runnable from source

rjmurillo:
## Summary

The `ai-agents-change-control` provenance table had one row whose re-verify
command could not work when copied from the file. Replaced it with a
pipe-free equivalent that needs no escaping at all.

## The defect

GFM renders `\|` as `|`, so a human reading the rendered table sees a working
command. Agents read raw markdown and copy the escaped form verbatim, which is
broken twice over:

1. `ls ... \| grep -E "..."` does not pipe. Bash strips the backslash during
   quote removal and hands `ls` a literal `|` argument, followed by `grep`,
   `-E`, and the pattern. `ls` rejects `-E` as an invalid option and exits 2
   having listed nothing. Stdout is empty, so an agent that reads stdout
   without checking the return code sees no retrospectives and concludes they
   are absent.
2. `grep -E "908\|1187\|..."` matches nothing. In POSIX ERE, `\|` is a literal
   pipe, not alternation. GNU and BusyBox grep do read `\|` as alternation in
   basic mode, but that is a GNU extension rather than a portable guarantee:
   POSIX defines no alternation operator for basic regular expressions.

## Evidence

Measured against the working tree:

```console
$ ls .agents/retrospective/ \| grep -E "908\|1187\|1887\|1965\|2205"
ls: invalid option -- 'E'
Try 'ls --help' for more information.
rc=2, stdout 0 bytes

$ ls .agents/retrospective/ | grep -E "908\|1187\|1887\|1965\|2205"
rc=1, no output

$ find .agents/retrospective \( -name "*908*" -o -name "*1187*" \
    -o -name "*1887*" -o -name "*1965*" -o -name "*2205*" \)
rc=0
.agents/retrospective/2026-01-15-pr-908-comprehensive-retrospective.md
.agents/retrospective/2026-02-08-session-1187-skip-prepush-abuse.md
.agents/retrospective/2026-05-05-pr-1887-iteration-paradox.md
.agents/retrospective/2026-05-10-pr-1965-review-axes-convergence.md
.agents/retrospective/2026-06-02-pr-2205-customer-wedge-incident.md
```

An agent following the old row would conclude the retrospectives do not exist.

## Why this remedy

Every other row in this table is already pipe-free. This was the only row that
needed an escape, which is why it was the only one that broke. Removing the
pipe makes the raw source and the rendered table byte-identical, so the class
of defect cannot recur in this row. `find` with `-name` and `-o` is POSIX, so
it does not depend on GNU `find` extensions.

## Verification

Extracted the command from the file itself, split the row on `|`, and executed
it: 3 cells (the table still renders) and all five retrospectives returned.
Negative control on the pre-fix line yields 8 cells, so the extraction test
discriminates rather than passing trivially.

## Scope

> [!NOTE]
> The Copilot CLI copy is generated from the Claude source by
> `build/scripts/generate_skills.py`. It was regenerated, not hand-edited.
> Both plugin manifests are bumped to the same version, as the parity gate
> requires. `src/claude/` is untouched, so its manifest is not bumped.

Refs #4079

### DIFF
--- a/.claude/.claude-plugin/plugin.json
+++ b/.claude/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "project-toolkit",
   "description": "Complete project development toolkit: agents, slash commands, lifecycle hooks, and reusable skills for Claude Code workflows",
-  "version": "0.6.5445",
+  "version": "0.6.5446",
   "author": {
     "name": "rjmurillo"
   }
--- a/.claude/skills/ai-agents-change-control/references/provenance.md
+++ b/.claude/skills/ai-agents-change-control/references/provenance.md
@@ -18,7 +18,7 @@ Verified against the working tree on 2026-07-03. Volatile facts and their re-ver
 | SHA-pin tension (Exceptions: None vs GP-006) | `.agents/governance/PROJECT-CONSTRAINTS.md:180`; `.agents/governance/golden-principles.md:63-69` | `grep -n "Exceptions" .agents/governance/PROJECT-CONSTRAINTS.md; sed -n 63,70p .agents/governance/golden-principles.md` |
 | ADR-066 proposed, ADR-071 accepted, #2230 rejected | Status sections of both ADR files | `sed -n 1,20p .agents/architecture/ADR-066-hook-fail-open-reconciliation.md; sed -n 1,10p .agents/architecture/ADR-071-plugin-hook-runtime-contract-verification.md` |
 | FM-9 and FM-10 sections; "neutral default" quote | `.agents/governance/FAILURE-MODES.md:284,315` | `grep -n "neutral default" .agents/governance/FAILURE-MODES.md` |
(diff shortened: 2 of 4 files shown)


rjmurillo:
Adversarial review (gpt-5.6-sol) found two inaccuracies in how this PR described the defect. Both are now corrected in the PR body. Recording them here because the commit message still carries the superseded wording and rewriting a pushed commit is not worth a force-push.

1. The old text said a plausible directory listing still reaches stdout. It does not. Measured: stdout is 0 bytes, stderr carries `ls: invalid option -- 'E'`, and the return code is 2. `ls` aborts on the invalid option before listing anything. The harm is the opposite of what I wrote: the command produces no evidence at all, so an agent concludes the retrospectives are absent.

2. The old text said only BRE treats an escaped pipe as alternation, implying POSIX sanction. POSIX defines no alternation operator for basic regular expressions. GNU grep 3.11 and BusyBox grep both take the alternation reading (measured), but that is a GNU extension, not a portable guarantee.

The implementation itself is unchanged and was approved. Both findings were about the explanation, not the fix.

Issue #2208, opened 2026-06-02
fix(github): deduplicate superseded check runs in get_pr_checks


rjmurillo:
## Problem
During pr-autofix for PR #2201, `.claude/skills/github/scripts/pr/get_pr_checks.py --pull-request 2201` reported `OverallState=FAILURE` because an older `Validate PR` check run failed on the same commit, even though a newer `Validate PR` check run succeeded and `test_pr_merge_ready.py` reported the PR ready.

## Impact
Automation can treat a ready PR as blocked when GitHub keeps superseded check runs with the same name on one commit.

## Expected behavior
`get_pr_checks.py` should deduplicate check runs by name or branch-protection context and prefer the latest run, matching `test_pr_merge_ready.py` behavior.

## Evidence
PR #2201 head before refresh: `c48ebb1ad081e3445211f6263b07cdc46aec667e`.
`Validate PR` had one failed run and one newer successful run. `test_pr_merge_ready.py` returned `CanMerge=true` while `get_pr_checks.py` returned failure.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #2208: fix(github): deduplicate superseded check runs in get_pr_che
- **Type**: complexity_triage
- **Priority**: medium
- **Suggested Agent**: agent-qa
- Issue labeled with medium complexity.
Suggested agents: agent-qa

### Route issue #2208 to agent-qa
- **Type**: agent_routing
- **Priority**: low
- **Suggested Agent**: agent-qa
- Issue content suggests routing to: agent-qa
Complexity: medium

PR #4112, opened 2026-07-31
fix(pr-maintenance): decide failing checks from the latest run and page past 100 contexts

### LINKED ISSUE #3978: PR maintenance misreads statusCheckRollup: stale failures never clear, and contexts truncate at 100
## Summary

`has_failing_checks` in both copies of the PR maintenance script reads
`statusCheckRollup` without deduping superseded check runs and without
paginating past 100 contexts. Two independent defects follow. Both are measured
below against live PRs.

The correct implementation already exists in this repo. `get_pr_checks.py` under
the github skill paginates, tracks `totalCount`, and dedupes by name with an
explicit rank function. The maintenance script does none of that.

## Defect A: a re-run that goes green never clears the failing flag

`statusCheckRollup` is per head commit and retains every check run for a name,
including superseded ones. `rollup.state` aggregates all of them. When a check
fails and a later run of the same check succeeds without a new commit, the
rollup keeps both and `state` stays `FAILURE`.

`has_failing_checks` returns True on `state` alone, so the PR is classified
`HAS_FAILING_CHECKS` forever until someone pushes a new commit.

This is not hypothetical. It is the normal repair path for the PR description
gate: you edit the PR body, the gate re-runs, it goes green, and no new commit
is created.

### Measurement, PR 3977 at head 86f3620a

Full pagination, 120 context nodes across 92 distinct names.

Latest run per name with a failing conclusion: none.
Names whose only failure is a superseded run: Validate PR.
`rollup.state`: FAILURE.

So every current check is green and the script reports the PR as failing.

Two control PRs measured at the same time, 3961 and 3976, have no stale failures
and correctly report `state` SUCCESS. The discriminator is the presence of a
superseded failing run, not something about PR 3977.

## Defect B: contexts are silently truncated at 100

The query uses `contexts(first: 100)` with no `pageInfo`, no `totalCount`, and
no pagination. Measured `totalCount` on three live PRs: 120, 126, and 126. So 20
to 26 nodes are dropped on every PR in this repo, and nothing reports that the
set is partial.

Today the loop over `contexts.nodes` is masked by the `state` early return, so a
failure in the dropped tail still gets caught. That masking is accidental. The
loop exists because `state` alone was not trusted, and it is scanning an
incomplete set. Narrowing the early return later, which Defect A requires, turns
the truncation into a live fail-open.

### Reproduction

Query the rollup for any open PR in this repo requesting `totalCount` alongside
`contexts(first: 100)`, then compare `totalCount` to the number of nodes
returned. The gap is the silently dropped tail.

## Impact

The maintenance dispatcher wakes on PRs that are already green, and its reason
string says `HAS_FAILING_CHECKS`, which sends a human or an agent to look for a
failure that no longer exists. Combined with Defect B, the reverse is possible
once the `state` shortcut is removed.

## Suggested fix

Reuse the logic that already works instead of writing a third one. Paginate
until `pageInfo.hasNextPage` is false, assert the fetched count equals
`totalCount` and fail loudly when it does not, group contexts by name, keep the
most recent run per name, and decide from that set rather than from
`rollup.state`.

Both copies of the script need the fix, and a test should pin that a rollup
containing a superseded failure plus a later success for the same name is
classified as passing. The inverse case, a superseded success plus a later
failure, must be classified as failing, or the fix is one directional and
proves nothing.

## References

See `get_pr_checks.py` under the github skill for the pagination and dedupe
implementation to reuse.

rjmurillo:
## Summary

The pr-maintenance dispatcher woke agents on green PRs. It decided "this PR has failing checks" from `statusCheckRollup.state` alone, and that field aggregates every check run recorded against the head commit, including runs a later re-run superseded. A check that failed and then passed on a re-run, with no new commit, left `state` at `FAILURE` forever. PR 4040 at oid 308e2094 had 127 contexts, zero failing latest runs, one superseded `Validate PR` failure, and the shipped `has_failing_checks` still returned True. That is the normal repair path for the PR description gate, so the dispatcher burned triage cycles on PRs that were already green.

Underneath that sat a second defect the first one was masking. The GraphQL query asked for `contexts(first: 100)` with no `totalCount`, no `pageInfo`, and no pagination. Measured against live PRs, the real counts were 121, 127, 119, and 122, so 19 to 27 context nodes were dropped per PR, silently. While `rollup.state` was the decider, that truncation was invisible. The moment you narrow that early return, which fixing the first defect requires, an incomplete context set becomes a live fail-open: a genuinely failing check sitting in the unread tail reads as passing and nobody is told.

Both defects are fixed once, in the shared library both scripts already import. `rollup_has_failing_checks` pages the contexts connection to completion, groups runs by check name and owning workflow, keeps only the most recent run per group, and decides from those winners. When the context set cannot be completed, because pagination failed, no commit oid was available to page with, or fewer nodes arrived than `totalCount`, it logs the gap and falls back to `rollup.state`. That is exactly the old behavior, so a truncated read fails closed instead of deciding from a partial set.

Two design points worth flagging for the reviewer. Ordering is by recency, not by the verdict precedence in `dedupe_checks` in the get_pr_checks script, which ranks passing above failing per issue #2208. Precedence would classify "superseded SUCCESS, then a later FAILURE" as passing, which is the exact inverse case this dispatcher must still wake on, so verbatim reuse of that helper was not sufficient. The grouping key is (check name, workflow), not the name alone: this repository has 9 workflows with a job named "Check Changed Paths" and 2 with "Aggregate Results", and those are concurrent siblings rather than re-runs of each other. Measured on PR 4069, three FAILURE runs of "Check Changed Paths" completing at 01:26:07Z, 01:26:08Z, and 01:26:10Z were all discarded by a SUCCESS of the same name from a different workflow at 01:26:11Z. The workflow name, not the check-suite id, is the right discriminator, because a re-run lands in a new check suite under the same workflow.

Fixes #3978

## What changed

| File | Why |
|---|---|
| `scripts/github_core/checks_rollup.py` | Canonical fix. Adds `CONTEXT_NODE_FIELDS`, `_CONTEXTS_PAGE_QUERY`, `_context_key`, `latest_contexts_by_run`, `_fetch_remaining_contexts`, `_collect_contexts`, and `rollup_has_failing_checks`. Both scripts already import from `scripts.github_core`, so no new transport was needed. |
| `.claude/lib/github_core/checks_rollup.py` | Plugin mirror of the canonical module, regenerated by the repository's plugin lib sync script. |
| `src/copilot-cli/lib/github_core/checks_rollup.py` | Same mirror for the Copilot plugin root. |
| `scripts/invoke_pr_maintenance.py` | Query now requests `oid`, `totalCount`, `pageInfo`, and the timestamp and workflow fields via the shared `CONTEXT_NODE_FIELDS`. `has_failing_checks` drops the state-first logic and delegates to the shared helper. `classify_prs` threads `owner` and `repo` through so pagination can issue the follow-up query. |
| `.github/scripts/invoke_pr_maintenance.py` | The same change to the twin. This is the copy the workflow actually wires up, and the two are hand-maintained rather than generated, so both had to move together. Also types the `pr` parameter as `dict[str, Any]`: the signature this branch rewrote carried a bare `dict`, which the pre-push mypy ratchet blocks on a changed line. |
| `tests/test_checks_rollup.py` | New suite for the shared helper: grouping, recency ordering, both superseded directions, cross-workflow masking, re-run in a new check suite, pagination, and every incomplete-set fallback path. |
| `tests/test_invoke_pr_maintenance.py` | Pins the same behavior against the `.github` copy, including the two dispatcher-level cases where a paged PR is or is not dispatched. |
| `tests/test_invoke_pr_maintenance_py.py` | Same pins against the `scripts/` copy. |
| `.agents/sessions/2026-07-31-session-4041-fix-checks-rollup-superseded-run-dedupe.json` | Session log per SESSION-PROTOCOL.md. |
| `.serena/memories/session/session-end-linked-worktree-targets-main-checkout.md` | Records a tooling defect found on the way, described under Noticed below. |

## Evidence

Targeted suite for the changed modules, re-run at the current branch tip after merging main:

```
uv run pytest tests/test_checks_rollup.py tests/test_invoke_pr_maintenance.py tests/test_invoke_pr_maintenance_py.py -q
============================= 118 passed in 1.19s ==============================
```

Full suite, run from a checkout outside `.claude/worktrees/` for the reason given under Noticed:

```
uv run pytest tests/ -q
========== 19792 passed, 29 skipped, 2 warnings in 423.18s (0:07:03) ===========
```

The full suite also ran as the lefthook pre-push gate on the same checkout, on the commit that took version slot 0.6.5495:

```
git push origin HEAD:refs/heads/fix/pr-checks-rollup-pagination
  ✔️ python-tests (579.42 seconds)
```

The reachability suite that fails only from inside `.claude/worktrees/`, run from the clean checkout:

```
uv run pytest tests/ci/test_validation_scripts_are_reachable.py -q
======================= 136 passed, 2 skipped in 12.96s ========================
```

Lint over the changed files:

```
uv run --extra dev ruff check scripts/github_core/checks_rollup.py .claude/lib/github_core/checks_rollup.py src/copilot-cli/lib/github_core/checks_rollup.py scripts/invoke_pr_maintenance.py .github/scripts/invoke_pr_maintenance.py tests/test_checks_rollup.py tests/test_invoke_pr_maintenance.py tests/test_invoke_pr_maintenance_py.py
All checks passed!
```

Repository pre-PR gate:

```
uv run python scripts/validation/pre_pr.py
RESULT: All validations passed
```

Plugin lib mirror parity:

```
uv run python scripts/sync_plugin_lib.py
All plugin lib copies are in sync.
```

## Sub-claims covered

1. **Defect A: `has_failing_checks` returns True on `rollup.state` alone.** Covered. `scripts/github_core/checks_rollup.py:332` `rollup_has_failing_checks` consults `rollup.state` only on the incomplete-set fallback path at `:370`; the complete path decides at `:372` from `latest_contexts_by_run` winners. Both call sites now delegate: `scripts/invoke_pr_maintenance.py:107` and `.github/scripts/invoke_pr_maintenance.py:189`.
2. **Defect A applies to BOTH copies of the script.** Covered. `scripts/invoke_pr_maintenance.py:107` and `.github/scripts/invoke_pr_maintenance.py:189` both lost the state-first logic and call the same shared helper, and both queries were updated in step at `scripts/invoke_pr_maintenance.py:155` and `.github/scripts/invoke_pr_maintenance.py:57`. `tests/test_invoke_pr_maintenance.py:193` and `tests/test_invoke_pr_maintenance_py.py:146` pin that both queries request the workflow name, so the twins cannot drift apart silently.
3. **Defect B: `contexts(first: 100)` with no pageInfo, no totalCount, no pagination.** Covered. Both queries now request `oid`, `totalCount`, and `pageInfo { hasNextPage endCursor }`, and `scripts/github_core/checks_rollup.py:265` `_fetch_remaining_contexts` walks the cursor chain, capped at 50 pages by `_CONTEXTS_MAX_PAGES` at `:130`. Pinned by `tests/test_checks_rollup.py:315` `TestPagination` and end to end at `tests/test_invoke_pr_maintenance.py:391` and `tests/test_invoke_pr_maintenance_py.py:580`.
4. **Defect B corollary: truncation becomes a live fail-open once the state return is narrowed.** Covered. `scripts/github_core/checks_rollup.py:305` `_collect_contexts` returns a `complete` flag that is False when pagination failed, when there was no owner, repo, or oid to page with, or when the node count is short of `totalCount` at `:326`. `rollup_has_failing_checks:361` logs the gap and falls back to `rollup.state`, so the truncated path fails closed. Pinned by `tests/test_checks_rollup.py:361` `TestIncompleteFallsBackToState`, and per copy at `tests/test_invoke_pr_maintenance.py:227` and `tests/test_invoke_pr_maintenance_py.py:183`.
5. **No test pinned that a superseded failure plus a later success is classified passing.** Covered. `tests/test_checks_rollup.py:177` `test_stale_failure_with_a_later_success_reads_as_passing`, plus the per-copy pins at `tests/test_invoke_pr_maintenance.py:207` and `tests/test_invoke_pr_maintenance_py.py:163`. The two tests that previously froze the defect are rewritten: `tests/test_invoke_pr_maintenance.py:198` and `:203` now assert that a `FAILURE` or `ERROR` rollup state with every latest run green reads as passing.
6. **No test pinned the inverse, a superseded success plus a later failure classified as failing.** Covered. `tests/test_checks_rollup.py:191` `test_stale_success_with_a_later_failure_reads_as_failing`, with the per-copy pins at `tests/test_invoke_pr_maintenance.py:217` and `tests/test_invoke_pr_maintenance_py.py:173`.
7. **Reuse the reference implementation rather than write it a third time, but recency ordering is needed because `dedupe_checks` ranks passing above failing.** Covered, and the divergence is deliberate and documented. `scripts/github_core/checks_rollup.py:108-124` names the get_pr_checks helper, states why its verdict precedence is wrong for this caller, and quotes `_FAILING_CONCLUSIONS` verbatim from it so the failing set stays as wide as `rollup.state` used to be. `latest_contexts_by_run` at `:237` orders by `(_context_time(node), index)`, which is recency with a deterministic tie-break. The transport is reused rather than rebuilt: `gh_graphql` already exported from `scripts.github_core`.

## Noticed on the way

Two tooling defects surfaced while shipping this. Neither is fixed here, because neither is in this branch's scope, and both are recorded in `.serena/memories/session/session-end-linked-worktree-targets-main-checkout.md`.

First, the session-end completion script resolves the repository root from `git rev-parse --git-common-dir`, which in a linked worktree points at the main checkout. Run with no arguments from a worktree it auto-detected and mutated a session log belonging to a different branch in the main checkout; that edit was reverted with `git checkout --`. Passing `--session-path` is then refused by its own containment check, so the correct log can never be targeted. It also writes `sessionEnd.reworkWarning.Evidence` as a list, which the schema rejects, since `Evidence` must be a string. This violates the anchoring requirement in the CI-scripts rule, MUST 7 and MUST 8.

Second, the validation-script reachability suite filters out any path containing the component `worktrees`. Agent worktrees live under `.claude/worktrees/`, so every source is filtered, the reference graph comes back empty, and 109 tests fail for the wrong reason. The Copilot skill-bundle suite parameter fails from the same location for a related reason. All 110 pass at this branch's head from a checkout outside that path, and all 110 pass at `origin/main` inside it, so the trigger is the checkout location and not this change.

> Manifest version rows removed: ADR-092 deleted the `version` field from all three packaged
> manifests, so this branch no longer touches them and the description would otherwise claim
> files that are not in the diff.


rjmurillo:
Closing as superseded by current `origin/main`.

Measured:
- The original #4112 branch fixed both duplicated script copies, not only one. It changed both `.github/scripts/invoke_pr_maintenance.py` and `scripts/invoke_pr_maintenance.py`, with test coverage for each copy.
- Current `origin/main` now fixes the same #3978 defects through `scripts/pr_maintenance_rollup.py`.
- Both current script copies import that shared helper path: `.github/scripts/invoke_pr_maintenance.py` and `scripts/invoke_pr_maintenance.py` both call `complete_status_check_rollups` and `rollup_has_failing_checks`.
- The helper requests `totalCount` and `pageInfo`, paginates up to `_CONTEXTS_MAX_PAGES`, marks incomplete rollups, dedupes by latest timestamp per check name, and treats incomplete rollups as failing.
- `uv run --frozen pytest tests/test_pr_maintenance_rollup.py tests/test_invoke_pr_maintenance.py tests/test_invoke_pr_maintenance_py.py -q` on `origin/main` passed: 99 passed.

No branch change remains required for #4112's stated fix.

PR #4122, opened 2026-07-31
fix(rules): guard the unchecked git fetch in the version-enumeration recipe

rjmurillo:
## Summary

The version-enumeration recipe in `.claude/rules/plugin-version-bump.md` runs `git fetch origin --quiet` without checking its exit status. When the fetch fails (no network, auth expired, remote unreachable), the recipe silently continues against stale remote-tracking refs and still prints a maximum version. The caller gets a plausible number that is wrong, and every downstream guard passes.

An adversarial review flagged the recipe's atomicity. This fixes the one sub-claim that was genuinely uncovered.

## What changed

Guarded the fetch:

```
git fetch origin --quiet || { echo "ERROR: git fetch failed; refs may be stale" >&2; exit 1; }
```

Added a prose paragraph explaining why the existing emptiness guard cannot catch this: `git show` reads local remote-tracking refs, so a failed fetch yields a stale but non-empty answer, which the emptiness check accepts.

## What this does NOT change

The `sort -V` prerelease-ordering caveat. The rule already documents it, with the same worked example and a "latent rather than live" judgment. Re-fixing it would be a duplicate.

## Verification

Extracted the fixed snippet verbatim from the committed markdown and ran it both ways:

| Case | Expected | Actual |
| --- | --- | --- |
| Fetch fails (unreachable remote) | non-zero exit, no version printed | exit 1 |
| Fetch succeeds (real remote) | exit 0, three version lines | exit 0, all three printed |

Gates: `build_all.py --check` exit 0. `plugin-version-bump: OK`. Negative control on the version gate run with the revert **committed** (a working-tree revert passes vacuously because the validator diffs `origin/main`): exit 1 with a precise message.

Both instruction mirrors regenerated with `build/scripts/generate_rules.py`. Parity pair bumped to `0.6.5488`, enumerated with the rule's own recipe.

Refs #4114

### DIFF
--- a/.claude/.claude-plugin/plugin.json
+++ b/.claude/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "project-toolkit",
   "description": "Complete project development toolkit: agents, slash commands, lifecycle hooks, and reusable skills for Claude Code workflows",
-  "version": "0.6.5486",
+  "version": "0.6.5488",
   "author": {
     "name": "rjmurillo"
   }
--- a/.claude/rules/plugin-version-bump.md
+++ b/.claude/rules/plugin-version-bump.md
@@ -46,7 +46,7 @@ same day, the parity line topped out at `0.6.5472` and the `src/claude` line at
 then bump the parity pair together to one value and `src/claude` to its own.
 
 ```bash
-git fetch origin --quiet
+git fetch origin --quiet || { echo "ERROR: git fetch failed; refs may be stale" >&2; exit 1; }
 for MANIFEST in .claude/.claude-plugin/plugin.json \
                 src/copilot-cli/.claude-plugin/plugin.json \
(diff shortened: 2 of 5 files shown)


rjmurillo:
Closing as obsoleted by #4179, not as rejected. The defect this PR reports was real.

The unchecked `git fetch origin --quiet` in the version-enumeration recipe did exactly what this PR says: on a failed fetch it kept going against stale remote-tracking refs and still printed a maximum version, so the caller got a plausible wrong number and every downstream guard passed. That is a genuine silent-failure bug and the guard here was the right shape for it.

#4179 removes the recipe. ADR-092 deletes the `version` field from all three packaged manifests, so there is no maximum version to enumerate, no fetch to guard, and no caller to mislead. `.claude/rules/plugin-version-bump.md` no longer contains the block this PR patches.

Nothing is lost by closing. Reopen if the version field ever returns.

Issue #4039, opened 2026-07-30
bug(validation): net-new type ignores bypass the mypy ratchet


rjmurillo:
## Problem

Issue #3761 decided to remove `type: ignore` from `SECURITY_SUPPRESSION_RE` because typing belongs outside the security suppression gate. That scope decision is correct, but its stated replacement control is not effective.

The changed-line mypy ratchet only sees errors mypy reports. A net-new `# type: ignore[...]` suppresses the error before mypy emits it, so the ratchet has nothing to block. After excluding type ignores from the security gate, no automated control challenges new typing suppressions.

## Reproduction

1. Add a typed assignment that mypy rejects.
2. Add `# type: ignore[assignment]` on the same changed line.
3. Run the changed-line mypy ratchet. Mypy emits no error, so the ratchet passes.
4. Run the narrowed security suppression gate. It also passes because `type: ignore` is intentionally excluded.

## Impact

Contributors can add typing suppressions without review or ratchet pressure. This is a type-quality gap, not a security vulnerability. The code comment and issue #3761 rationale currently overstate coverage by saying the mypy ratchet owns these suppressions.

## Suggested direction

Add a separate typing-suppression policy or count ratchet for net-new `type: ignore` comments. Keep it separate from security suppression enforcement. Include move detection so refactors do not reclassify existing suppressions as new.

## References

- #3761, FR-2.3 and its decision to exclude `type: ignore`
- PR #4023, review follow-up
- `scripts/validation/git_hook_policy.py`, changed-line mypy ratchet and suppression regex

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #4039: bug(validation): net-new type ignores bypass the mypy ratche
- **Type**: complexity_triage
- **Priority**: high
- **Suggested Agent**: agent-security
- Issue labeled with high complexity.
Suggested agents: agent-security, agent-implementer, agent-qa

### Route issue #4039 to agent-security, agent-implementer, agent-qa
- **Type**: agent_routing
- **Priority**: low
- **Suggested Agent**: agent-security
- Issue content suggests routing to: agent-security, agent-implementer, agent-qa
Complexity: high

PR #4138, opened 2026-07-31
fix(ratchet): remove the dead type-ignore that turned main red

rjmurillo:
## Problem

`origin/main` is red. Any open PR that still carries the 56th occurrence inherits the failure:

```
type-ignore count ratchet: REGRESSION. 56 violations > baseline 55 (+1)
```

Reproduced on a pristine checkout of `origin/main` with zero local changes, and observed on the CI of two unrelated PRs whose diffs contain no Python at all.

## Cause

An ordering race between two independently green PRs.

| When | PR | Effect on the count |
|------|----|---------------------|
| 2026-07-31 11:46 UTC | #4093 (`cea600edc8`) | Added the ratchet and its baseline of 55. Correct at the time: the count at that commit is exactly 55. |
| 2026-07-31 18:04 UTC | #4091 (`088292977f`) | Added one suppression, taking the count to 56. No baseline bump. |

Neither PR saw the other. Both were green on their own branch. Main went red only once both had landed. This is the same failure shape as a stale version bump: a baseline computed against `origin/main` at PR creation time goes stale when a sibling lands first.

Verified by counting at each commit:

```
git grep -o -E '#\s*type:\s*ignore' cea600edc8 -- '*.py' | wc -l   # 55
git grep -o -E '#\s*type:\s*ignore' origin/main  -- '*.py' | wc -l   # 56
```

and by diffing the per-file tallies between the two commits, which isolates exactly one added occurrence.

## Fix

Delete the added suppression in `scripts/validation/pr_commit_count.py`. It is provably dead:

```
$ uv run mypy --warn-unused-ignores scripts/validation/pr_commit_count.py
scripts/validation/pr_commit_count.py:238: error: Unused "type: ignore" comment  [unused-ignore]
```

mypy infers the surrounding value as `Any`, so there is no `arg-type` error for the comment to suppress. Removing it is behavior preserving. The explanatory comment on the line above is kept, because it still documents the invariant the helper enforces for human readers.

The baseline stays at 55 rather than rising to 56. Raising it would permanently widen the ceiling to admit a suppression that mypy says is unnecessary.

## Verification

```
type-ignore count ratchet: OK (count == baseline 55)          exit 0
tests/validation/test_pr_commit_count.py                      98 passed
tests/ci/test_type_ignore_count_ratchet.py                    17 passed
mypy + ruff on the changed file                               clean
```

Negative control: stashing this single line restores `exit 1` with the same message, and restoring it returns `exit 0`. The change is what fixes the gate, not anything environmental.

## Impact: main is currently failing this gate

The ratchet is a whole repo count compared against a baseline file that is
checked in beside it. Both inputs live on main, and main's own tree does not
satisfy its own baseline.

Direct measurement on a clean detached checkout of `origin/main` at 088292977f:

```text
$ uv run python scripts/ci/type_ignore_count_ratchet.py
type-ignore count ratchet: REGRESSION. 56 violations > baseline 55 (+1).
exit 1
```

The same command on this branch:

```text
type-ignore count ratchet: OK (count == baseline 55).
exit 0
```

The two trees differ by exactly the one line this PR deletes. Confirmed by
diffing every Python file and the baseline file between them: empty.

Two open PRs inherit the failure without introducing it. #4136 changes sixteen
markdown and JSON files and zero Python files, and is red on this check. #4131
is red for the same reason. Four other open PRs are green because their trees
either predate the suppression or already delete it.

That makes this a repair of a broken default branch rather than a cleanup, so
it wants to merge ahead of the queue.

## Notes

One further unused suppression exists and is deliberately left alone, to keep this PR to the single line that unblocks the repository:

```
scripts/github_core/bot_config.py:13: error: Unused "type: ignore" comment  [unused-ignore]
```

It predates the baseline, so it is already inside the 55 and is not causing the failure. `warn_unused_ignores` is not enabled repo-wide, which is why both survived. Enabling it would catch this class at authoring time and is worth its own PR.

### Overlap with an open PR

Adversarial review surfaced that #4124 independently carries the identical one-line deletion, bundled into a larger change to the rule-activation eval instrument. Verified with `gh pr diff 4124`.

Either PR unblocks the ratchet, so this one is not required. It is kept separate because it is reviewable in one line and #4124 carries unrelated scope. Whichever merges first makes the other's hunk a no-op; the loser resolves a one-line conflict. Close this PR instead if the preference is to let #4124 carry it.

### A separate defect found during review, deliberately not fixed here

`_external_non_first_parent_shas` reads `parent.get("sha")` twice: once inside `_is_external_parent` for validation, once at the call site for insertion. A `dict` subclass whose `get` returns different values on successive calls can insert a SHA that was never validated. Reproduced locally.

It is unreachable from the production path, because the payload arrives through `json.loads`, which emits plain `dict` and never a subclass. It also predates this diff. Fixing it means changing the helper's signature and adding a regression test, which would take this PR past the single line that unblocks the repository, so it ships separately.

Refs #4039

### DIFF
--- a/scripts/validation/pr_commit_count.py
+++ b/scripts/validation/pr_commit_count.py
@@ -198,44 +198,68 @@ def main_first_parent_shas(repo_root: Path, run_git: GitRunner | None = None) ->
     return frozenset(result.stdout.split())
 
 
-def _is_external_parent(parent: object, own_shas: set[str]) -> bool:
-    """Return True only when this parent is a readable sha the branch does not own.
+def _external_parent_sha(parent: object, own_shas: set[str]) -> str | None:
+    """Return this parent's sha only when it is readable and the branch does not own it.
 
     An unreadable parent must not buy relief: a parent that is not a mapping,
-    carries no ``sha``, or carries a non-string ``sha`` is unreadable and fails
-    closed to False.
+    carries no ``sha``, or carries a ``sha`` that is not exactly ``str`` is
+    unreadable and fails closed to None.
+
+    The sha is read once and returned, so the caller inserts the same value that
+    was validated here. Reading the key a second time at the call site would let
+    a mapping whose ``get`` is not a pure lookup return a different, unvalidated
+    value on the second read.
+
+    Two narrower guards keep that single read honest. ``dict.get`` is called
+    unbound so a subclass cannot override the lookup to raise or to answer
+    differently. The sha is then normalised through the unbound ``str.__str__``
+    to an exact ``str``, because a ``str`` subclass may define ``__hash__`` and
+    ``__eq__`` that miss on the ``own_shas`` probe and match afterwards, which
+    would let the branch's own sha pose as external and buy the exemption this
+    function gates. Normalising rather than rejecting keeps a subclass from
+    being read as an unrelated sha.
     """
     if not isinstance(parent, dict):
-        return False
-    sha = parent.get("sha")
-    if not isinstance(sha, str) or not sha:
-        return False
-    return sha not in own_shas
+        return None
+    sha = dict.get(parent, "sha")
+    if not isinstance(sha, str):
+        return None
+    sha = str.__str__(sha)
+    if not sha:
+        return None
+    return None if sha in own_shas else sha
 
 
 def _external_non_first_parent_shas(commits: list[Any]) -> set[str]:
     """Collect SHAs of non-first parents that are not in the PR's own commit list.
 
     Used by contains_main_merge (precise check verified against origin/main).
+
+    ``own_shas`` is the deny list a parent is probed against, so it is built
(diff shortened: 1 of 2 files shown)


rjmurillo:
Closing as superseded by current `origin/main` and the sibling duplicate #4139.

Measured:
- #4138 and #4139 both target the same #4091 dead `type: ignore[arg-type]` in `scripts/validation/pr_commit_count.py`.
- #4138 also adds broader hardening and tests around hostile mapping values. That is outside the ratchet fix described by the PR.
- Current `origin/main` no longer has the suppression. It uses `shas.add(parent["sha"])` after `_is_external_parent` narrows the value.
- `uv run --frozen python scripts/ci/type_ignore_count_ratchet.py --base-ref origin/main` on `origin/main` passes with `count == baseline 53`.

No branch change remains required for this PR's stated fix.

Issue #4068, opened 2026-07-31
bug(ci): ADR-006 extraction converts loud shell failures into silent Python successes


rjmurillo:
## Summary

ADR-006 `run:` block extraction systematically converts loud shell failures into silent Python successes. Adversarial review of two in-flight extraction PRs (#3992, #4014) found **six** instances, and **three of them ship tests that assert the broken behavior**.

This is a defect *class*, not six unrelated bugs. Filing it so the remaining extraction work and the eventual ratchet tightening account for it.

## The mechanism

A `run:` block executes under `set -e`. Any command exiting nonzero fails the step. A command substitution whose producer dies also fails it.

The natural Python translation returns a sentinel instead: empty string, `None`, empty list, a dataclass with a `None` field, or a bare `except: pass`. When no caller converts that sentinel into a nonzero exit, the step goes green on a failure that used to be red.

Extraction is therefore a silent-pass **generator**, not only a silent-pass detector.

## Confirmed instances

| Script | Shell original | Python replacement | Consequence |
|---|---|---|---|
| `spec_load_content.py` | recursive `find .agents/specs` | top-level-only glob | Every spec lives in a subdirectory, so ID lookup returns nothing, writes "No spec content found", exits 0. A required status check grades a placeholder. |
| `smoke_install_tarball.py` | shell resolves `npm` | `subprocess.run(["npm", ...])` | Cannot resolve `npm.cmd`; both Windows matrix rows crash before doing work. |
| `measure_npm_pack_size.py` | `set -e` on npm pack | `main()` returns `EXIT_OK` regardless | Pack size collection fails, publishing proceeds. |
| `drift_create_alert_issue.py` | `cat` on the details file | missing file becomes `""` | A content-free "Agent Drift Detected" issue is filed. |
| `artifact_create_issues.py` | `set -e` on each create | warning, then `return 0` | Every `gh issue create` can fail and the workflow is green. |
| `spec_extract_refs.py` | assignment under `set -e` | nonzero exit becomes empty scope | Validation grades the full parent spec instead of the declared incremental slice. |

Verified independently for `spec_load_content.py`:

```
glob('.agents/specs/*REQ-005*')                    -> []
glob('.agents/specs/**/*REQ-005*', recursive=True) -> ['.agents/specs/requirements/REQ-005-command-skill-bundling.md']
glob('.agents/specs/*DESIGN-011*')                    -> []
[...]

PR #4150, opened 2026-08-01
fix(ci): add CI backstop for the security suppression policy

### LINKED ISSUE #4061: Security suppression gate has no CI backstop
## Summary

The security suppression policy runs only through local Lefthook hooks. CI does
not run the same no-net-new suppression check. A contributor who does not have
Lefthook installed, or who bypasses hooks, can push a new `nosec`,
`nosemgrep`, `lgtm`, `codeql`, or security `noqa` directive without this policy
checking it.

This is a process enforcement gap, not an external attacker privilege gain.
The actor already needs repository write access. It still means the repository
states a blocking policy that the protected branch does not enforce.

## Evidence

PR #4023 wires the check into local hooks:

- `lefthook.yml`: `security-suppressions-staged` under pre-commit.
- `lefthook.yml`: `security-suppressions-staged` under pre-merge-commit.
- `lefthook.yml`: `security-suppression-policy` under pre-push.

No GitHub workflow references `security-suppressions`,
`staged-suppressions`, or `check_pushed_suppressions`.

The independent security review for PR #4023 confirmed:

- The file-scoped removal-credit behavior is intentional and Low risk.
- The larger gap is that CI never invokes the suppression gate.
- `audit-hook-bypass.yml` is non-blocking and records bypass evidence only.

Related discourse:

- #3770 fixed the same missing-backstop shape for conflict markers.
- #2808 fixed CI checks that converted their own failures into green runs.
- #3899 established the file-scoped no-net-new suppression contract.

## Impact

- Local hook installation determines whether the policy runs.
- Hook bypass reaches CI without this policy producing a failure.
- Code review must notice new suppression directives manually.
- Enforcement differs between contributors and automation.

## Proposed fix

Run the existing suppression diff scanner in PR CI. Use the same parser and
file suffix set as the local hooks. Preserve the documented file-scoped
removal credit so CI and local behavior remain identical.

## Acceptance criteria

- A PR adding a net-new security suppression fails a required CI check.
- A PR moving an equal suppression payload within one file passes.
- Cross-file removal credit remains blocked.
- One removal cannot pay for two additions.
- CI and Lefthook call the same policy implementation.
- Failure output names each path and suppression line.

rjmurillo:
## Summary

The security-suppression policy shipped as a lefthook gate only. `pre-commit` and `pre-merge-commit` ran `security-suppressions-staged`; `pre-push` ran `security-suppressions-push`. No workflow referenced either subcommand, so the policy bound exactly those contributors who had lefthook installed locally and did not pass `--no-verify`.

This adds a `security-suppressions-range --base --head` subcommand that runs the identical collector over an explicit range, and wires it into the `validate-pr` job.

Fixes #4061
Refs #4068

## Why validate-pr and not a new job

`validate-pr` reports the required context "Validate PR", one of the 17 the branch ruleset requires. A new job would create a context that is not in the required set, so the gate could be bypassed by never running it.

The step sits in the unconditional section of the job, alongside `Enforce model pin policy`. Steps guarded by `steps.should-run.outputs.skip` are skipped for `dependabot[bot]`, `github-actions[bot]`, and `renovate[bot]`. Bots author code, so a gate they are exempt from is weaker than the local hook it backstops, not equal to it. Placing it unconditionally makes the CI gate strictly stronger than the hook.

## Fails closed

A shallow clone yields no merge base. That case exits 2 (config error), never 0. A security gate that cannot read history must not report success. Both checkouts in the job now request full history for that reason.

## Verification

17 tests, all driving `main(argv)` and asserting on its exit code rather than on a helper return value, which is the #4068 failure shape.

A 7-mutant harness, all dead:

| Mutant | Result |
|---|---|
| Delete the CI step from the workflow | DEAD |
| Revert full history on the bot-fallback checkout | DEAD |
| Fail open when the range cannot be resolved | DEAD |
| Swallow the collector's `None` sentinel | DEAD |
| Ignore `--base`, merge-base against head itself | DEAD |
| Reporter always reports clean | DEAD |
| Subcommand ignores `--head` | DEAD |

The first harness run had four survivors. Two mutations had silently failed to apply, and the workflow assertion was a bare substring match that passed with the step name deleted and only the `run:` line surviving. Both were fixed; the wiring assertions now parse the workflow and check step structure.

## Two premises corrected while building this

**A test I wrote was wrong, not the policy.** `test_one_removal_cannot_pay_for_two_additions` originally paid for one removed `noqa` with one added `noqa` plus one added `type: ignore`, and expected a failure. It got a pass. Probing the shipped `security-suppressions-push` hook showed it returns 0 on that same input, so the new range check was faithful to the shipped policy. `SECURITY_SUPPRESSION_RE` matches `nosec`, `nosemgrep`, `noqa`, `lgtm[`, `codeql[`, and `cwe-suppress`. `type: ignore` is a typing suppression with no security signal, governed by its own counter. The test now uses two same-class tokens, and a second test pins the scope boundary explicitly so widening the regex cannot happen silently.

**The inline SHA form was an injection risk.** `actionlint` passed it. The repo's own `scripts/validate_workflows.py` rejected it. Both SHAs now bind through `env:`, and a test asserts the run block contains no `${{` at all.

## Out of scope, filed separately

Security-relevant steps in `validate-pr` that sit behind the bot-skip guard remain exempt for the three bot actors. This change does not alter that; it only declines to add to it.

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -238,6 +238,12 @@ jobs:
       - name: Checkout repository for workflow validation
         if: steps.should-run.outputs.skip == 'true'
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          # The unconditional steps below include `Enforce security suppression
+          # policy`, which resolves a merge base between the PR base and head
+          # (issue #4061). A default depth-1 checkout has no merge base, so the
+          # gate would exit 2 on every bot PR. Matches the primary checkout.
+          fetch-depth: 0
 
       - name: Setup uv for workflow validation
         uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
@@ -314,3 +320,24 @@ jobs:
       # changed pins, so it passes clean on today's tree.
       - name: Enforce model pin policy
         run: uv run --frozen python3 scripts/validation/check_model_pins.py --mode enforce
+
+      # Issue #4061: the security-suppression policy shipped only as a lefthook
+      # gate (`security-suppressions-staged` on pre-commit and pre-merge-commit,
+      # `security-suppressions-push` on pre-push). No workflow referenced it, so
+      # it bound exactly those contributors who had lefthook installed and did
+      # not pass --no-verify. `security-suppressions-range` runs the same
+      # collector over the PR's base..head range, so CI and the hook cannot
+      # disagree about what the policy is.
+      #
+      # Deliberately outside the `steps.should-run.outputs.skip` guard. Bots
+      # author code changes here, and a security gate they are exempt from is
+      # weaker than the local hook it backstops, not equal to it.
+      - name: Enforce security suppression policy
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          uv run --frozen python3 scripts/validation/git_hook_policy.py \
+            security-suppressions-range \
+            --base "$BASE_SHA" \
+            --head "$HEAD_SHA"
--- a/scripts/validation/git_hook_policy.py
+++ b/scripts/validation/git_hook_policy.py
@@ -2261,25 +2261,64 @@ def _push_updates(stream: TextIO, repo_root: Path) -> list[PushUpdate] | None:
     return updates
 
 
-def check_pushed_suppressions(stream: TextIO, repo_root: Path) -> int:
-    updates = _push_updates(stream, repo_root)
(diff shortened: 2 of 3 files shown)


rjmurillo:
Closing as superseded by current `origin/main`.

Measured:
- #4150 only adds the #4061 CI backstop for net-new security suppressions.
- Current `origin/main` already has `security-suppressions-diff` in `scripts/validation/git_hook_policy.py`.
- Current `origin/main` wires that gate in `.github/workflows/pytest.yml` at two required Python-test paths.
- The remaining #4110 rebase dropped its older `security-suppressions-range` conflict path and now focuses on the unsuperseded #4068 CLI exit-contract ratchet and ADR-006 exact-count gate.

No branch change remains required for #4150's stated fix.

PR #4166, opened 2026-08-01
fix(eval): keep judge failure evidence, gate on exact averages and salvage rate

### LINKED ISSUE #3975: eval: judge parse failures store only a 200-char prefix, which destroys the evidence and fabricates a cause
## Problem

`eval-rule-activation.py` stores only a 200-character prefix of the judge
response when the response fails to parse. That prefix is too short to
diagnose the failure, and it actively produces a misleading diagnosis.

Measured against the 24 archived judge failures in
`.agents/analysis/eval-artifacts/2026-07-29-unified-software-engineering/`:

- All 24 stored prefixes are exactly 200 characters.
- Re-parsing them reports `Unterminated string starting at` for 19 of 24.
  That error is caused by the truncation, not by whatever the judge emitted.
- Only 5 of 24 show a stray quote inside the stored window.

## Impact

A reviewer reading these artifacts cannot tell why the judge output failed.
Worse, the truncation supplies a plausible-looking wrong answer: the
`Unterminated string` message reads like a finding when it is an artifact of
the storage limit.

This already produced a published false claim. An earlier draft of
`.claude/skills/context-optimizer/references/rule-audit-procedure.md` asserted
that all 24 failures were unescaped quotes in the judge's `reasoning` field
and that verbose models trip it more often. Adversarial review found neither
claim was supported. Both have been retracted; the document now says the cause
is not recoverable from the archive and cites this issue.

## Proposed fix

Store the full judge response on parse failure, or raise the cap high enough
that the failure region is inside it. These are failure artifacts, so they are
rare by construction and the storage cost is bounded. Also record the model
identity on the sample rather than relying on the filename, so per-model
claims are checkable without parsing paths.

## Acceptance

- A judge parse failure retains enough of the response to identify the
  original parse error.
- Re-parsing a stored failure reproduces the same error the eval saw live.
- Model identity is readable from the sample record.

Refs #3953

### LINKED ISSUE #3988: eval: stop hand-writing a malformed-JSON recovery path; twelve defects of one class over nine review rounds
## Summary

`scripts/eval/eval-rule-activation.py` recovers scores from judge output that
`json.loads` rejects. Over nine rounds of adversarial review, that recovery path
has produced **twelve defects of a single class**: a search picking the wrong
candidate object and reporting it as the judge's answer, with `judge_failed`
false.

Each round fixed the instance and left the class standing. That is the evidence
this issue is about. The question is no longer "is this parser correct yet" but
"is writing this parser the right approach at all."

## The twelve

Each was found by adversarial review, not by the test suite, and each was a
real fabrication reachable from a real payload shape.

| Round | What the search picked instead of the verdict |
|---|---|
| 1 | The first balanced object, which is a tool trace |
| 2 | An object inside the judge's own quoted rubric |
| 3 | A score-shaped object quoted inside a malformed `reasoning` string |
| 4 | An object reached by resuming one character past an unbalanced opener |
| 5 | A field donated by a second root object |
| 6 | A second root object that *was* the verdict, outranked by the first |
| 7 | A `6` re-admitted as a clean `5` because salvage ran after the range gate |
| 8 | An object inside a top-level array, because both walkers counted braces and ignored brackets |
| 9 | An exemplar at offset 0 followed by an explicit English refusal |
| 10 | A second verdict serialized inside a string, whose escaped keys evaded the duplicate-name count |
| 11 | A verdict found past offset 0 on the *success* path, returned with no marker at all |
| 12 | A partial object outranked by a later complete one |

The current state anchors both paths at offset 0, refuses any payload naming a
score field twice, and marks every prefix recovery `judge_salvaged`. Round 10
of review is running now. Convergence is not assumed.

## Why this is structural, not a run of bad luck

Three independent arguments, all raised by the reviewer (GPT-5.6 Sol) across
rounds 7, 8, and 9:

1. **The safety argument is self-undermining.** Every version justified its
   choice by reading structure out of the payload. The payload is known
   malformed. When a safety argument depends on parsing what you already know
   is broken, the argument is the defect, not the implementation.

2. **Position cannot establish provenance.** Anchoring at offset 0 replaced
   "search for the best candidate" with "take the first." That is cheaper and
   more auditable, but an offset still cannot say whether the object is the
   judge's answer or something the judge was quoting. Round 9's finding 1 is
   exactly this and remains open by design: an exemplar at offset 0 followed by
   "I cannot score the response" still grades 5/5/5.

3. **The asymmetry is unbounded on one side.** A refused sample costs one of
   three judge samples and is recorded as a failure. A fabricated sample
   silently corrupts a published number and is recorded as a success. Twelve
   times, the fabrication side was the one that fired.

## What the archive says about the value being defended

Measured across all eight archived runs of the published experiment:

| Quantity | Count |
|---|---|
| Judge samples | 288 |
| Graded, fed the medians | 264 |
| Failed, excluded from every median | 24 |
| Carrying a salvage marker | 0 |

`_reduce_samples` drops failed samples rather than folding them in as zeros, so
the 24 failures cost nothing. **Salvage never produced a published number.** The
entire twelve-defect history is in service of a path that, on the only real
dataset available, has never once changed a result.

That is the strongest argument for deleting it: the measured benefit is zero and
the measured cost is twelve fabrication defects.

## Options

1. **Delete the recovery path. Require strict JSON.** A sample whose payload
   does not parse is a failed sample. Three samples per cell already tolerate
   one loss, and the archive says the loss rate would have been 24 of 288, or
   8.3%, none of which reached a cell anyway. Simplest, and it eliminates the
   class rather than the instance.

2. **Constrain the judge instead of the parser.** Use structured output or a
   tool-call schema so the provider guarantees a parseable object. This moves
   the problem to a layer that can actually enforce it. Availability differs
   per provider; the Copilot CLI path would need checking.

3. **Keep the anchored recovery, gate it on the marker.** Recoveries are now
   marked. Fail the run when the salvaged fraction exceeds a threshold, so a
   provider that starts emitting preambles is caught rather than silently
   accommodated. Weakest, but it preserves samples if option 1 proves too
   lossy in practice.

Option 1 is what the measurement supports. Option 3 is what to do if a future
provider makes option 1 expensive, and the marker added in round 9 is what
makes that decision measurable rather than speculative.

## Why an ADR

This is a decision about what the instrument is allowed to infer, not a bug
fix. It has been re-litigated implicitly twelve times because it was never
decided explicitly. An ADR would record the asymmetry, the measured benefit,
and the chosen bound, so round 13 does not start from scratch.

## Verification is not blind

Worth stating plainly: the same author wrote every one of the parsers and every
one of the tests. "It survived nine rounds of review" is weaker evidence than
the round count suggests, because each round hardened the same author's mental
model. Falsifying the current version properly needs one of: blinded manual
transcription of the failed payloads, a second parser written by an author who
has not seen them, or held-out malformed output from a different provider.
None has been done.

## Related

- #3983 (PR): the round 8 and 9 hardening, and the twelve-defect history in
  `.claude/skills/context-optimizer/references/rule-audit-procedure.md`
- #3975: only 200 characters of a failed payload are archived, so failures
  cannot be fully re-examined after the fact
- #3934: the judge is drawn from the same model family being evaluated
- #3956: the artifact does not record which model produced it

### LINKED ISSUE #4031: _judge_parse_failure returns judge_failed: False, inverting what its name states
Found while verifying a published claim against the archive. The name cost me four measurement rounds and nearly caused me to retract a correct finding.

## What it does

`_judge_parse_failure(text, message)` in `scripts/eval/eval-rule-activation.py` is the **salvage** path. On an unparseable payload it extracts the three top-level integers and returns:

```
judge_failed:   False
judge_salvaged: True
reasoning:      "scores salvaged from unparseable judge output: ..."
```

The sample counts as graded.

## Why the name is wrong

Read the entry branch and the control flow says the opposite of what happens:

```python
embedded = _recover_verdict(text)
if embedded is None:
    return _judge_parse_failure(text, f"judge parse error: {text[:200]}")
```

A reader traces "recovery returned nothing, so return parse failure" and concludes the sample failed. It did not. `_recover_verdict` returning `None` means no *embedded complete object* was found; the three scores are still extractable, and `_judge_parse_failure` extracts them.

## Measured consequence

Verifying the claim "all 24 archived failures salvage", I probed the helper in isolation:

- `_recover_verdict(raw)` -> `None` for **24 of 24**
- `_strict_json_loads(raw)` -> raises for **24 of 24**

Both true, and both irrelevant to the question. The entry point gives `judge_failed=False, judge_salvaged=True` for **24 of 24**. Two correct helper measurements pointed the opposite way from the real behavior, purely because the name of the third function asserts an outcome it does not produce.

## Suggested fix

Rename to state the action rather than the trigger, for example `_salvage_scores_from_unparseable` or `_grade_from_partial_payload`. The call site then reads "no embedded object, so grade from the partial payload", which is what happens.

The docstring should also state the return contract explicitly, since `judge_failed: False` from a function with `failure` in its name is the surprising part.

## Scope

Naming and docstring only. No behavior change. The behavior is correct and is covered by tests; the archived-triple end-to-end test exercises this path.

Refs #3975

### LINKED ISSUE #4070: eval: the restraint gate compares a display-rounded average against the floor
Found by independent adversarial review of PR #4056. Confirmed by construction, not observed on the live suite.

## Summary

`_mechanism_summary` in `scripts/eval/eval-rule-activation.py` rounds the mechanism average to two decimals:

```python
avg = round(sum(scores) / len(scores), 2) if scores else None
```

That rounded value is what the gates then compare against their thresholds:

```python
if worst_neg_avg is not None and worst_neg_avg < MIN_RESTRAINT_SCORE:   # 3.5
    return "FAIL_OVER_ACTIVATION"
...
passes_threshold = desc_avg >= MIN_ACTIVATION_SCORE                      # 3.5
```

Rounding is a presentation concern. Gating on the presented number rather than the measured one means a true average of `3.49758` reads as `3.50` and clears a floor it is below. The instrument publishes a verdict the measurement does not support, which is the class of defect PR #4056 exists to close.

## Reachability

Cell scores are `k/3` for integer `k` in `[3,15]` (the mean of three rubric integers), or `k/6` when an even sample count puts the median at a midpoint. So a pool of `N` cells has average granularity `1/(3N)` or `1/(6N)`.

To land inside `[3.495, 3.5)` without hitting `3.5` exactly, the granularity must be finer than the `0.005` rounding window:

| Reduction path | Minimum negative-pool size | Witness |
|---|---|---|
| legacy triple-mean, `1/3` | **35** | `K=367`, avg `3.49524` |
| `cell_score` even-sample midpoint, `1/6` | **34** | `K=713`, avg `3.49510` |

Every scenario file in `tests/evals/rule-scenarios/` currently holds at most 3 negative scenarios, so this is not reachable on the shipped suite today. It becomes reachable the moment the negative pool is grown, which issue #3933 and the PR #4056 session log both name as required follow-up work.

## Scope

This is not introduced by PR #4056. The `round(..., 2)` and the `desc_avg >= MIN_ACTIVATION_SCORE` comparison both predate it on `origin/main`; the new restraint gate inherits the existing pattern. It was left unfixed there deliberately, because the narrow fix touches both pools and PR #4056 guarantees all 8 archived artifacts replay unchanged.

## Suggested fix

Separate the published value from the gating value. Keep the rounded average in the summary for the table, and compare the unrounded average against `MIN_RESTRAINT_SCORE` and `MIN_ACTIVATION_SCORE`. Preserve the negated `>=` form in `_positive_verdict`, since every comparison against NaN is False and a `<` form falls through to PASS.

Add a regression pinning an average inside the rounding window just below each floor, plus a check that the archived replay is unaffected.

Refs #3933, #3989, PR #4056.

rjmurillo:
## Summary

The rule-activation eval stored a 200-character prefix of every judge payload that failed to parse. Measured failure offsets across the 24 archived failures run 162 to 421, so the window cut the damage out: re-parsing the stored prefixes yields "Unterminated string starting at" 19 times while the real payloads yield "Expecting ',' delimiter" 24 times. The archive recorded the truncation, not the judge. Operators reading a failed run got a fabricated cause.

That one storage decision hid three more defects in the same file, each filed separately. The gates compared display-rounded averages against their floors, so a negative pool measuring 3.49524 published 3.5 and returned PASS instead of FAIL_OVER_ACTIVATION. The `judge_salvaged` marker was written twice and read nowhere, so no gate could bound how much of a verdict came from recovery. And `_judge_parse_failure` returned `judge_failed: False` on its primary path, which is the opposite of what the name says.

All four are now fixed. Every failed and salvaged judge record carries the whole payload under `judge_raw`, on all three salvage paths and in both sibling scorers. Every judge sample carries `judge_model`, including the RuntimeError path, so a per-model claim is checkable from the record instead of the artifact filename. Every gate reads an exact average while the table keeps publishing the rounded one, so archived runs replay field for field. The salvage marker survives reduction, reaches the run summary, and `--max-salvaged-fraction` fails the run when recovery carried more of the verdict than the bound allows. ADR-091 records the decision to keep the hand-written recovery path rather than delete it, the round-9 residual it accepts, and provider-enforced structured output as the exit path.

Fixes #3975
Fixes #4070
Fixes #4031
Fixes #3988

## What changed

| File | Why |
|---|---|
| `scripts/eval/eval-rule-activation.py` | All four defects live here. Stores the whole payload under `judge_raw` on every failure and salvage path, stamps `judge_model` on every sample, emits `avg_score_exact` beside the rounded `avg_score` and points every gate at it, propagates `judge_salvaged` through reduction as `salvaged_sample_count` and `salvaged_sample_fraction`, adds the `--max-salvaged-fraction` bound, and renames `_judge_parse_failure` to `_salvaged_or_failed_judge` with the return contract stated in the docstring. |
| `scripts/eval/eval-agents.py` | Same truncation shape at the zero-score failure record. Now stores `judge_raw` alongside the truncated reasoning line. `_avg_scores` reads a fixed dimension list, so the extra key reaches no average. |
| `scripts/eval/eval-knowledge-integration.py` | Same truncation shape, same fix, same reason the extra key is inert. |
| `tests/eval/test_eval_rule_activation.py` | Payload retention on all three paths, the cross-path invariant that every salvaged sample the gate counts carries its payload, model stamping on four sample kinds, rounding-window boundary tests for all three gates, rounded-tie ranking, and the salvage disclosure with its negative controls. Also revives `test_dry_run_counts_repeated_judge_calls`, which skipped on every run since its pinned rule file left the repository. |
| `tests/eval/test_eval_agents_dry_run.py` | New. Whole payload stored, the stored payload reproduces the live parse error while the 200-character prefix does not, a clean payload stores no raw, and the extra key reaches no average. |
| `tests/eval/test_eval_knowledge_integration.py` | New, same four cases for the second sibling scorer. |
| `.agents/architecture/ADR-091-judge-verdict-recovery-bounds.md` | New. Four files on this branch cited ADR-091 while the file itself was untracked. Records the decision, the accepted residual, and the exit path. |
| `.agents/analysis/ADR-091-debate-log.md` | New. Records the review that actually happened and states plainly that the adr-review six-agent debate did not run, because this harness exposes no Task tool. |
| `.agents/analysis/eval-artifacts/2026-07-29-unified-software-engineering/README.md` | Told readers only 200 characters of each payload survive. `recovered-judge-payloads.json` sits in the same directory with the full raw for all 288 samples and the README never mentioned it. |
| `.claude/skills/context-optimizer/references/rule-audit-measurement-discipline.md` | Named `_judge_parse_failure` in prose and called the name an outcome it does not produce. Now records the old name as history and points at the new one. |
| `.claude/skills/context-optimizer/references/rule-audit-parser-forensics.md` | Pointed readers at the issue for the argument. Now also names ADR-091, which records what was decided. |
| `src/copilot-cli/skills/context-optimizer/references/*.md` | Copilot mirrors, regenerated with `build/scripts/generate_skills.py`. |
| `.claude/.claude-plugin/plugin.json`, `src/copilot-cli/.claude-plugin/plugin.json` | Both plugin roots ship the changed reference files, so the parity pair moves together to 0.6.5482. |
| `.agents/sessions/2026-07-31-session-3975-eval-recovery-evidence-and-gating.json` | Session log for this work. |
| `.agents/memory/episodes/episode-2026-07-31-session-3975-*.json` | Episode extracted from the session log by the commit hook. |
| `.agents/retrospective/2026-08-01-eval-judge-recovery-evidence.md` | Records the shared root across the four issues, the two tests that passed for the wrong reason on the first pass, and two tooling defects found on the path. |

## Evidence

```
uv run pytest tests/eval/test_eval_rule_activation.py tests/eval/test_eval_agents_dry_run.py tests/eval/test_eval_knowledge_integration.py -q
```

```
396 passed in 4.35s
```

Full suite, run by the pre-push hook against the pushed head:

```
uv run pytest tests/ -q
```

```
19717 passed, 28 skipped, 50 deselected, 2 warnings in 467.07s (0:07:47)
```

Also green: `uv run --extra dev ruff check` on all six changed Python files ("All checks passed!"), `uv run mypy` on the three changed eval scripts ("Success: no issues found in 1 source file" for `eval-rule-activation.py`; the one remaining `eval-agents.py:732` operator error is pre-existing on `main` at line 725 and untouched by this branch), and `uv run python scripts/validation/pre_pr.py` ("RESULT: All validations passed").

Every new regression test was run against pre-fix code and required to fail there.

### #3975

- 200-character prefix is too short to diagnose a judge parse failure: covered, `scripts/eval/eval-rule-activation.py:1466` states the contract and `:1481`, `:1501`, `:569` store the whole payload.
- The truncation fabricates a cause: covered, `tests/eval/test_eval_rule_activation.py:5065` asserts the stored payload no longer re-parses to a different error than the live one.
- Acceptance 1, a parse failure retains enough of the response to identify the original parse error: covered, `scripts/eval/eval-rule-activation.py:1481` and `:1501`, tested at `tests/eval/test_eval_rule_activation.py:5050`.
- Acceptance 2, re-parsing a stored failure reproduces the error the eval saw live: covered, `tests/eval/test_eval_rule_activation.py:5065` for the primary scorer, `tests/eval/test_eval_agents_dry_run.py:397` and `tests/eval/test_eval_knowledge_integration.py:127` for the siblings.
- Acceptance 3, model identity readable from the sample record: covered, `scripts/eval/eval-rule-activation.py:552`, `:1482`, `:1502`, `:1629`.
- Sibling truncation sites in the same function (the ambiguous two-verdict refusal and the non-object JSON refusal): covered, both route through `_salvaged_or_failed_judge` at `scripts/eval/eval-rule-activation.py:511` and `:545`, tested at `tests/eval/test_eval_rule_activation.py:5105`.
- The RuntimeError judge sample carries no model: covered, `scripts/eval/eval-rule-activation.py:1629`, tested at `tests/eval/test_eval_rule_activation.py:5224`.
- Discovered during triage, the archive README says the evidence is gone while it sits next to the reader: covered, `.agents/analysis/eval-artifacts/2026-07-29-unified-software-engineering/README.md:78` now names `recovered-judge-payloads.json`.

### #4070

- Restraint gate compares a display-rounded average against `MIN_RESTRAINT_SCORE`: covered, the worst negative average is selected off `avg_score_exact` at `scripts/eval/eval-rule-activation.py:1968` and compared at `:2076`. Test at `tests/eval/test_eval_rule_activation.py:5363`.
- Positive activation gate compares the same rounded average: covered, `scripts/eval/eval-rule-activation.py:1836` feeds `_positive_verdict` at `:2050`. Test at `tests/eval/test_eval_rule_activation.py:5384`.
- `round(..., 2)` in `_mechanism_summary` conflates the published value with the gated value: covered, `scripts/eval/eval-rule-activation.py:1795` emits `avg_score_exact` beside the rounded `avg_score`, and the published headline stays rounded (test at `tests/eval/test_eval_rule_activation.py:5354`).
- The fix must preserve the negated `>=` form in `_positive_verdict` so a NaN average fails closed: covered as a non-regression, `scripts/eval/eval-rule-activation.py:2050-2055` keeps the negated form with the reason in the docstring; `tests/eval/test_eval_rule_activation.py:658` still pins NaN failing closed.
- Regression test pinning an average inside the rounding window just below each floor: covered, `tests/eval/test_eval_rule_activation.py:5363`, `:5384`, `:5392`, one per gate.
- Check that the 8 archived artifacts replay unaffected: covered, `tests/eval/test_eval_rule_activation.py:5354` pins the published headline as rounded so archived summaries replay field for field.
- Adjacent, not claimed in the issue: `beats_baseline` subtracts two rounded averages: covered, `scripts/eval/eval-rule-activation.py:2051` now subtracts the exact averages. Test at `tests/eval/test_eval_rule_activation.py:5392`.

### #4031

- `_judge_parse_failure` returns `judge_failed: False` on the salvage path, inverting what the name states: covered, renamed to `_salvaged_or_failed_judge` at `scripts/eval/eval-rule-activation.py:1456`.
- The call site reads as if the sample failed when it did not: covered, `scripts/eval/eval-rule-activation.py:511` and `:545`.
- The docstring does not state the return contract: covered, `scripts/eval/eval-rule-activation.py:1457-1470` states both branches and what each returns.
- The remedy is a rename that states the action rather than the trigger, without colliding with `_salvage_scores`: covered, the new name is `_salvaged_or_failed_judge`; `_salvage_scores` is untouched.
- Consequential, not claimed in the issue: the skill reference naming `_judge_parse_failure` goes stale on rename: covered, `.claude/skills/context-optimizer/references/rule-audit-measurement-discipline.md:111-113` and its Copilot mirror.

### #3988

- The hand-written recovery path still exists, Option 1 (delete) not taken: covered as a decision rather than a code change. ADR-091 records why deleting is wrong: the measurement that motivated deletion was retracted in-repo at `.claude/skills/context-optimizer/references/rule-audit-parser-forensics.md:59-66`.
- Round 9 finding 1, an exemplar object at offset 0 followed by an explicit English refusal still grades 5/5/5: covered as accepted-by-design and now written down. ADR-091 records the residual, and the refusal text is no longer discarded: `tests/eval/test_eval_rule_activation.py:5134` asserts the salvaged record keeps the refusal its scores contradict.
- Option 3, gate the run on the salvage marker: covered, `scripts/eval/eval-rule-activation.py:634` (reduction), `:1726-1728` (run summary), `:2325` (flag), `:2700` (gate). Tests at `tests/eval/test_eval_rule_activation.py:5522` and the exit-code cases at `:5669-5682`.
- Option 2, provider-enforced structured output or tool-call schema: NOT covered by code. ADR-091 names it as the durable exit path and the condition under which it gets built. Implementing it means changing how every provider adapter in `scripts/eval/` requests a verdict, which is a separate change with its own blast radius.
- No ADR records the decision, so it keeps being re-litigated: covered, `.agents/architecture/ADR-091-judge-verdict-recovery-bounds.md`, status proposed.
- Verification is not blind, the same author wrote every parser and every test: NOT covered. This branch does not fix it and cannot: the same constraint applies to this change. `.agents/analysis/ADR-091-debate-log.md` states plainly that the adr-review six-agent debate did not run because this harness exposes no Task tool, and records the three-reviewer review that did happen instead. An independent reviewer on this PR is the mitigation available today.

## Notes for the reviewer

Two defects found on the path, neither fixed here:

- `.claude/skills/session-end/scripts/complete_session_log.py:76` resolves the repository root with `git rev-parse --git-common-dir`, which in a linked worktree points at the main checkout. It auto-detected and wrote to an unrelated session log belonging to another branch; the stray write had to be reverted by hand. Passing `--session-path` with a path inside the worktree then failed with "Session path must be inside '<main checkout>/.agents/sessions'". This is the issue #3402 class, and `.claude/rules/ci-scripts.md` MUST 7 already requires the check this script skips.
- `tests/ci/test_validation_scripts_are_reachable.py:110` skips any path containing a part named `worktrees`. A worktree created under `.claude/worktrees/` therefore has every source skipped, the reference graph comes back empty, and 109 of that file's tests fail with "the probe is not walking the tree". `tests/test_skill_bundle_suites_run.py` fails from the same location for a related reason. Both pass from a worktree outside that path, which is where this branch was pushed from.

### DIFF
--- a/.agents/analysis/ADR-091-debate-log.md
+++ b/.agents/analysis/ADR-091-debate-log.md
@@ -1,0 +1,38 @@
+# ADR-091 Debate Log
+
+## What ran, and what did not
+
+The `adr-review` skill's six-agent debate did not run against `ADR-091: Judge Verdict Recovery Bounds`. The sessions that authored and reviewed it ran as isolated worktree subagents with no Task tool, so no subagent could be spawned. This file records the review that did happen and names the gap, rather than reporting a consensus nobody reached.
+
+Read every verdict below as coming from a reviewer that examined the branch, not from the six-role panel the skill would have convened. Nothing here should be read as a 6/6 accept.
+
+## Reviewers
+
+| Reviewer | Basis | Verdict on the decision | Verdict on the artifact |
+|----------|-------|-------------------------|--------------------------|
+| Reviewer 1 (independent, branch diff) | Read the whole diff plus `git status`, `git ls-files`, `git diff main HEAD --stat` | Not contested | BLOCKING: four committed references to a file that does not exist in the repo |
+| Reviewer 3 (independent, branch diff) | Same diff, plus the orphan-ref-validator and `evals/architect-spike/fixtures/A001.json` | Not contested | MAJOR: three #3988 sub-claims stay open without this file; the repo's own eval plants `ADR-091` as the canonical fabricated citation |
+| Reviewer 5 (independent, branch diff) | Same diff, plus `pre_pr.py` run to confirm no gate catches it | Not contested | MAJOR: the 0.15 default's rationale is unrecoverable while the ADR is untracked |
+| Round 16 measurement (`.claude/skills/context-optimizer/references/rule-audit-parser-forensics.md`) | 288 archived judge payloads replayed | Supports keeping the recovery path | Retracts the zero-marker inference issue #3988 rested on |
+
+All three independent reviewers reached the same conclusion from the same evidence and none disputed the decision the ADR records. That is agreement about a missing file, not a debate about the architecture.
+
+## Findings and resolution
+
+| Finding | Resolution |
+|---------|------------|
+| ADR-091 cited four times, tracked nowhere | Committed alongside the four citing files. |
+| Decision item 6 claimed every salvaged record keeps `judge_raw`, and the prefix recovery in `score_response` kept none | Fixed in `scripts/eval/eval-rule-activation.py`; item 6 now names all three salvage paths. |
+| The `--max-salvaged-fraction` bound moved the exit code and the rendered report disclosed nothing | Salvage rate added to `_render_caveats`; decision item 5 records it. |
+
+## What the decision rests on, and what would overturn it
+
+The ADR keeps the recovery path because deleting it drops 24 of 288 archived samples, 8.3%, measured after issue #3988 was filed and against a retracted zero. Two facts would overturn it:
+
+- A provider-enforced structured-output schema (`response_format` or a tool-call schema) lands anywhere in `scripts/eval/`. Recovery becomes dead code and deletion is free. Named as the exit path in the ADR and out of scope there.
+- A re-measurement showing the 8.3% figure is itself an artifact of the replay rather than of the run. The 288 payloads are archived at `.agents/analysis/eval-artifacts/2026-07-29-unified-software-engineering/recovered-judge-payloads.json`, so this is checkable.
+
+## Accepted risks
+
+- Two residuals ship live. An exemplar object at offset 0 followed by a prose refusal still grades 5/5/5, and adjacent string literals are undetected. Both are recorded in the ADR and pinned by tests. Both now retain the payload under `judge_raw`, so a reader can find them in the artifact.
+- The six-agent debate is unrun. An agent that can spawn subagents should run `adr-review` against this ADR and replace this file with its output. Until then the ADR's status stays `proposed`.
--- a/.agents/analysis/eval-artifacts/2026-07-29-unified-software-engineering/README.md
+++ b/.agents/analysis/eval-artifacts/2026-07-29-unified-software-engineering/README.md
@@ -72,12 +72,16 @@ Two limits on what this archive can support:
   marked until rounds 9 and 10 added `judge_salvaged` to them. That gap is the
   reason the marker exists, and the reason no claim about tightening the
   parser can be validated here.
-- **Re-walking the 24 failures is weaker than it looks.** Only 200 characters
-  of each payload survive (#3975), so a re-parse exercises the stored prefix,
-  not the payload the judge actually emitted. A duplicate score field in the
-  discarded tail would refuse under the current guard and is invisible here.
-  All 24 prefixes recover; that shows the hardening did not make the parser
-  worse on what was kept, not that it matches on what was not.
+- **Re-walking the 24 failures needs the recovered payloads, not the run
+  artifacts.** The run stored only a 200-character prefix of each payload
+  (#3975), so a re-parse against these files exercises the prefix, not what the
+  judge emitted. `recovered-judge-payloads.json` in this directory carries the
+  full raw for all 288 samples, recovered from the Copilot CLI session
+  transcripts rather than from the eval, so the 24 failures can be re-parsed
+  against the real thing. The eval now stores the whole payload under
+  `judge_raw`, so a future run needs no recovery step. All 24 prefixes recover;
+  that shows the hardening did not make the parser worse on what was kept, not
+  that it matches on what was not.
 
 A third figure that was published as free is not. Of the 264 graded samples,
 none has a `reasoning` string naming a score field. That was read as evidence
--- a/.agents/architecture/ADR-091-judge-verdict-recovery-bounds.md
+++ b/.agents/architecture/ADR-091-judge-verdict-recovery-bounds.md
@@ -1,0 +1,104 @@
+---
+id: ADR-091
+status: proposed
+date: 2026-07-30
+decision-makers: [rjmurillo]
+supersedes: []
+superseded-by: null
+explainer: null
+implemented: true
+---
+
+# ADR-091: Judge Verdict Recovery Bounds
+
+## Status
+
+Proposed. Implemented in the same change for issue #3988. The recovery path and its bound both ship in `scripts/eval/eval-rule-activation.py`.
+
+## Date
+
+2026-07-30
+
+## Context
+
+`scripts/eval/eval-rule-activation.py` scores model responses with an LLM judge that is told to answer with a JSON object holding three integers and one `reasoning` string. Some payloads do not parse. The judge quotes the response it is grading, an unescaped quote inside that prose invalidates the object, and the three numbers the eval actually needs sit before the damage.
+
+The script recovers those numbers. Twenty-three rounds of adversarial review found sixteen defects of one class in that recovery: a search picking the wrong candidate object and reporting it as the judge's answer with `judge_failed` false. Each round fixed the instance. Issue #3988 argued the class is the problem and proposed deleting the path outright.
+
+Two facts decided against deletion, and both were measured after the issue was filed.
+
+The zero the deletion argument rested on is a serialization artifact. Issue #3988 reported 0 of 288 archived samples carrying a salvage marker and inferred that recovery had never moved a published number. `.claude/skills/context-optimizer/references/rule-audit-parser-forensics.md` retracts that inference: the artifacts store the state before recovery was applied, so the marker count says when the files were written, not what recovery did. Recovery moved one published cell. Round 16 recovered all 288 raw payloads from the Copilot CLI session transcripts and measured that 24 of 288 cells, 8.3%, would change if the run were re-scored today.
+
+Deleting the path therefore drops 8.3% of samples for a benefit the measurement no longer supports. At three judge samples per cell, that takes 17 Opus cells down to one or two graded samples.
+
+What issue #3988 was right about is that the decision has never been written down. There is no ADR. `judge_salvaged` was written at two sites and read at none, so option 3 of the issue, bounding the run on the marker, did not exist as code. And the round-9 residual is still live: an exemplar object at offset 0 followed by an explicit English refusal still grades 5/5/5.
+
+## Decision
+
+Keep the recovery path. Bound it, mark it, and record the residuals.
+
+1. **Anchor at offset 0.** A payload's verdict is its leading object, ignoring leading whitespace. Position does not establish provenance, but it is cheap, auditable, and it replaced the search that produced every one of the sixteen defects.
+2. **Refuse on a second named verdict.** Any payload naming a score field twice is refused rather than resolved. Picking one of two candidates is a guess.
+3. **Unwrap a fence only when nothing but whitespace sits outside it.** One fence is not a selection.
+4. **Mark every recovery.** A recovered sample carries `judge_salvaged: True`. The marker now survives `_reduce_score_samples` as `salvaged_sample_count` and reaches the run summary as `salvaged_sample_count`, `graded_sample_count`, and `salvaged_sample_fraction`.
+5. **Bound the run on the marker.** `--max-salvaged-fraction` defaults to 0.15 and fails the run with exit code 1 when the observed fraction exceeds it. 0.15 sits above the 8.3% the reference run would salvage today, so that run still reproduces, and below the rate a provider that started wrapping verdicts in prose would produce. The rate is disclosed in the rendered report whenever recovery moved anything, so the exit code never contradicts a table that says nothing.
+6. **Store the whole payload.** Every failed and every salvaged record keeps `judge_raw` and `judge_model` (issue #3975), so a recovery decision can be re-examined from the artifact instead of from a transcript that may not survive. All three salvage paths do this, not only the two inside `_salvaged_or_failed_judge`: the prefix recovery in `score_response` marked the sample and kept nothing, so the gate counted evidence it had discarded.
+
+### Accepted residuals
+
+Both are recorded rather than fixed, because the fix for each is a search, and the search is the defect class this ADR exists to bound.
+
+- **Exemplar at offset 0 followed by a prose refusal.** `{"activation_score": 5, ...}\nI cannot score the response.` grades 5/5/5. Reproduced at HEAD. Distinguishing a quoted exemplar from an answer requires reading meaning out of a payload already known to be malformed. The refusal text is now retained under `judge_raw`, so the residual is auditable from the artifact even though it is not detected.
+- **Adjacent string literals.** Pinned by `test_adjacent_string_literals_are_a_known_undetected_shape` in `tests/eval/test_eval_rule_activation.py`.
+
+### The asymmetry this rests on
+
+A refused sample costs one of three judge samples and is recorded as a failure, so it moves the median visibly through `judge_failed` and the graded sample count. A fabricated sample corrupts a published number and is recorded as a success. The bound above is calibrated to that asymmetry: recovery is allowed, but a run that leans on it more than its reference did fails rather than absorbing the change.
+
+### Exit path
+
+Provider-enforced structured output, a `response_format` or tool-call schema that makes the payload parseable by construction, is the durable answer. No such schema exists anywhere in `scripts/eval/` today, and the Copilot CLI path needs checking. That is out of scope here and belongs in its own issue.
+
+## Rationale
+
+The alternative considered and rejected was option 1 of issue #3988: delete the recovery path and require strict JSON. It is the simplest option and it eliminates the class rather than the instance. It was rejected because its supporting measurement was retracted in-repo before this decision was made, and the replacement measurement (24 of 288 cells would change) says deletion costs 8.3% of the sample pool.
+
+Option 2, provider-enforced structured output, is the right long-term answer and is named above as the exit path. It is not available today without a provider change.
+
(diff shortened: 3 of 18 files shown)


rjmurillo:
Closing as superseded by PR #4101 plus merged PR #4124.

Measured on current origin/main and the rebased #4101 branch:
- #4124 already landed the exact-average gate path. `scripts/eval/eval-rule-activation.py` on origin/main has `avg_score_exact` in `_mechanism_summary` and `_gate_avg()` reads it for threshold checks.
- #4101 deletes `_recover_verdict`, `_salvage_scores`, and `_judge_parse_failure` entirely, which is the narrower resolution for #3988 and also removes the #4031 inversion where `_judge_parse_failure` returned `judge_failed: False`.
- #4101 preserves the full raw provider payload as `raw_judge_response` on success and all failure paths. Targeted tests `TestUnifiedParsePath` and `TestRawJudgeResponseNoTruncation` pass after rebase.
- #4166 keeps the hand-written recovery path and adds salvage-rate gates around it. That conflicts with issue #3988 and duplicates the evidence/gating fixes now covered by #4101/#4124.

This PR is useful provenance, but landing it now would reintroduce the parser shape #4101 removes.

Issue #4060, opened 2026-07-30
Ruff F811 is blind to underscore-prefixed names, so duplicate test helpers shadow silently


rjmurillo:
## Symptom

Ruff's `F811` (redefinition of unused name) never fires on a name that starts
with an underscore. Two module-level `def _helper(...)` in the same file are
accepted silently. Python keeps the last definition, so every caller in the
file gets it, including the ones written above it.

Test helpers are underscore-prefixed by convention, so in practice `F811`
gives this repository no coverage at all on the files where a 4000-line
module makes an accidental collision most likely.

## Reproduction

`ruff 0.15.16`, repo config, `--select F811`:

```
def helper(a):        -> F811 fires
    return a
def helper(a, b): ...

def _helper(a):       -> silent
    return a
def _helper(a, b): ...
```

Only the leading underscore differs. `__helper` is silent too.

## Cause

`F811` skips any name matching ruff's `lint.dummy-variable-rgx`. The default
is `^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$`, which matches every
underscore-prefixed identifier. This repo does not set the key, so the default
applies (`pyproject.toml` has no `dummy-variable-rgx`).

## How it surfaced

While adding tests to `tests/eval/test_eval_rule_activation.py` (PR #4056) I
appended a module-level `_scenario_from` that collided with an existing one at
line 3557 with a different signature. Ruff stayed silent through several
runs. Three unrelated tests failing was the only signal. Renaming the new
helper to `_scenario_of_mechs` fixed it.

## Blast radius

`tests/eval/test_eval_rule_activation.py` has 16 underscore-prefixed
module-level defs; `tests/test_subprocess_text_encoding.py` has 60. Neither
file gets any `F811` protection on them.

A scan of all 621 files under `tests/` finds **no duplicate module-level defs
today**, so this is a regression guard rather than a cleanup.

## Proposed fix

Add a test that walks `tests/**/*.py` with `ast`, collects module-level
`FunctionDef` and `AsyncFunctionDef` names per file, and fails on any
duplicate, reporting the file and both line numbers. It is currently green,
runs in about a second, and closes the gap without touching ruff's config.

Narrowing `dummy-variable-rgx` was considered and rejected: the same regex
drives unused-argument and unused-variable reporting, so tightening it to fix
`F811` would change the meaning of `_` throughout the codebase.

rjmurillo:
[...]

Issue #3941, opened 2026-07-30
eval-model-sweep carries dead S404/S603 noqa comments that trap the next editor


rjmurillo:
## Summary

`scripts/eval/eval-model-sweep.py` carries two `noqa` comments for ruff rules the project does not select:

```
scripts/eval/eval-model-sweep.py:40:import subprocess  # noqa: S404 - invokes a sibling eval script with a fixed, validated argv
scripts/eval/eval-model-sweep.py:332:            completed = subprocess.run(  # noqa: S603 - argv is fixed + validated, shell=False
```

`pyproject.toml` sets `select = ["E", "F", "W", "I", "N", "UP", "B", "ANN"]`. `S` (flake8-bandit) is absent, so `S404` and `S603` never fire and the suppressions suppress nothing.

## Why this is a trap, not just noise

`git_hook_policy.py security-suppressions-push` blocks a push when a commit **adds** a line matching `SECURITY_SUPPRESSION_RE`, which matches any bare `noqa`. Existing lines are grandfathered because the check reads added diff lines only.

That creates a delayed trap. The lines sit there looking like sanctioned precedent. The next person editing that file copies the pattern, or touches those lines in a way that re-adds them, and their push is rejected for a rule that does not exist. That is what happened on `rjmurillo/shihipar-context-engineering`: the pattern was copied into a new provider, and the push was blocked seven comments later.

## Suggested fix

Delete both. Keep the rationale as a plain comment at the call site, where the argv is constructed and a reader can check the claim. The branch above did exactly that for its own copies; see commit `4fa3af92`.

## Broader check worth running

A one-time sweep for `noqa` codes outside the selected rule set would find any others. Every one is either a dead comment or a signal that `select` should include the rule.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #3941: eval-model-sweep carries dead S404/S603 noqa comments that t
- **Type**: complexity_triage
- **Priority**: high
- **Suggested Agent**: agent-security
- Issue labeled with high complexity.
Suggested agents: agent-security, agent-architect

### Route issue #3941 to agent-security, agent-architect
- **Type**: agent_routing
- **Priority**: low
- **Suggested Agent**: agent-security
- Issue content suggests routing to: agent-security, agent-architect
Complexity: high

rjmurillo:
[...]

Issue #3874, opened 2026-07-29
No gate catches a statement placed after return or raise; the repo is at zero today


rjmurillo:
## Summary

A statement placed after a `return` or `raise` in the same block never runs.
Nothing in the current gate set catches that. Ruff, mypy, pytest, the push gate,
review and merge all passed a change that carried one, and it was found by hand
afterwards.

The repository has zero of them today, so the gate can be added without a
backfill.

## Evidence

Scanned every first-party `.py` file, excluding `.venv`, `node_modules`,
`.git` and `site-packages`:

```
first-party files scanned: 1551
unreachable statements: 0
```

The same scan over the virtualenv finds 21, all in third-party packages
(`httpx`, `pycparser`, `ruamel.yaml`, `anthropic`), which is why the exclusion
matters and why a naive substring filter is not enough: those paths are
relative and do not start with a slash.

## The check

```python
import ast, pathlib

SKIP = (".venv", "node_modules", ".git", "site-packages")
for path in pathlib.Path(".").rglob("*.py"):
    if any(part in SKIP for part in path.parts):
        continue
    tree = ast.parse(path.read_text(encoding="utf-8"))
    for func in ast.walk(tree):
        if not isinstance(func, (ast.FunctionDef, ast.AsyncFunctionDef)):
            continue
        for index, statement in enumerate(func.body[:-1]):
            if isinstance(statement, (ast.Return, ast.Raise, ast.Continue, ast.Break)):
                print(path, func.name, func.body[index + 1].lineno)
```

It reads one block at a time and asks only whether a terminator is followed by
a sibling. That is a deliberate floor rather than a reaching-definitions
analysis: it has no false positives to argue about, and it catches the shape
that actually shipped.

## Open questions for the owner

1. Does this belong in `scripts/validation/` behind the pre-push gate, or in CI
   only? The scan takes under a second on 1551 files, so the push gate is
   affordable.
2. Should `ast.Continue` and `ast.Break` count, or only `Return` and `Raise`?
   Including them costs nothing today because the count is zero either way.
3. Ruff has `unreachable` behind a preview flag. Enabling that instead would be
   less code to own. Worth checking whether its rule covers the same shape
   before writing a bespoke script.

rjmurillo:
Question 3 answered by probe rather than left open.

Ruff does not cover this shape. Probed with every rule enabled and preview on:

```
$ printf 'def f():\n    return 1\n    x = 2\n' > /tmp/unreach_probe.py
[...]

PR #4164, opened 2026-08-01
fix(merge-resolver): make the leftover-marker gate merge-aware

rjmurillo:
## Summary

`verify_no_conflict_markers.py` is the BLOCKING Phase 3 gate for the merge-resolver skill and agent. It ran `git diff HEAD --check`. During a merge, `HEAD` is the topic tip, so every line arriving from the incoming branch reads as newly added. A committed fenced `<<<<<<< HEAD` example on the incoming side therefore reported as a leftover conflict marker and the gate exited 1. This repository ships two such files, `.claude/skills/merge-resolver/references/strategies.md` and `.serena/memories/patterns/pattern-handoff-merge-session-histories.md`, so any topic branch that predates them tripped the gate on merge. The script's own docstring claimed those files were "intentionally ignored", which was false during a merge.

The fix makes the check merge-aware. It computes the `file:line` intersection of the `HEAD` diff with the diff of every `MERGE_HEAD` parent. Content already committed on either side of the merge is excluded, and only markers the resolution itself introduced are reported. Switching to `--cached` was rejected because it would drop the unstaged working-tree coverage the script documents.

`MERGE_HEAD` is read line by line through `git rev-parse --git-path`, not through `git rev-parse --verify`. The latter resolves only the first line, so an octopus merge's later parents would go unfiltered.

Coverage narrowing worth naming: a genuine leftover marker that is already committed on a merge parent is now excluded too. `--check` cannot distinguish it from a fenced documentation example, and the resolution did not introduce it. That case belongs to the branch that committed it, and every branch is gated by this same check on its own PR.

The prose gates moved with the code. `SKILL.md` and both copies of the agent prompt pointed readers at bare `git diff --check`, which is vacuous once files are staged, and at a `grep -r "<<<<<<"` sweep that #2424 and #2445 already removed from the skill for false-failing on fenced examples. Both now invoke the script.

Fixes #4058

## What changed

| File | Why |
|------|-----|
| `.claude/skills/merge-resolver/scripts/verify_no_conflict_markers.py` | The fix. `_markers_against` holds the single-ref `--check` call, `_merge_parents` reads every `MERGE_HEAD` line, `_marker_location` splits the `file:line` key from the right so a path containing `": "` is not truncated, and `find_leftover_markers` intersects. Docstring rewritten; the old "committed historical content is intentionally ignored" claim was false mid-merge. |
[...]

PR #4208, opened 2026-08-01
fix(lint): enable RUF100, add unreachable-code gate, add duplicate-def gate, rescue misplaced tests

### LINKED ISSUE #3792: 490 ruff noqa directives suppress nothing; RUF100 is not selected so the count only grows
## Summary

`ruff` never checks whether a `# noqa` directive still does anything. Turning on `RUF100` reports **490 findings** across first-party Python. Two different problems are hiding in that number, and they want different fixes.

## Measurement

Run against `HEAD` of `main` today:

```
uv run ruff check --extend-select RUF100 --output-format=concise --quiet .
```

| Kind | Count | Meaning |
|---|---|---|
| `non-enabled` | 413 | Directive names a rule the repo does not select (for example `# noqa: S603` while `S` is absent from `select`). |
| `unused` | 77 | Directive names a selected rule that does not fire on that line. Dead. |
| **total** | **490** | |

`select` in `pyproject.toml:101` is `["E", "F", "W", "I", "N", "UP", "B", "ANN"]`.

Top files for the unambiguous `unused` half:

```
7  scripts/github_core/__init__.py
7  .claude/lib/github_core/__init__.py         (mirror of the above)
7  src/copilot-cli/lib/github_core/__init__.py (mirror of the above)
4  .claude/skills/memory/memory_core/__init__.py
3  tests/eval/test_providers.py
3  tests/eval/test_anthropic_model_default.py
3  .claude/skills/security-scan/scripts/scan_vulnerabilities.py
```

The `__init__.py` cluster is the clearest case. Every re-export in `scripts/github_core/__init__.py` carries `# noqa: F401`, and every one of those names is also listed in `__all__` a few lines below. `__all__` already satisfies `F401` for re-exports, so stripping all seven directives leaves `ruff check` passing. Verified by deleting them and re-running.

## Why this matters

Two costs, both small per instance and both compounding.

A dead directive is indistinguishable from a live one at a glance. `.claude/rules/code-quality.md` requires every suppression to be a narrowly scoped last resort carrying a short rationale. When most of the directives a reader encounters turn out to suppress nothing, the reader stops treating any of them as load-bearing, which is exactly the "train the next reader to distrust the checker" failure that rule names.

Nothing prevents accumulation. `RUF100` is not selected, so a directive that becomes dead when a rule is removed from `select`, or when the underlying violation is fixed, stays in the tree forever with no signal.

This surfaced during work on #3781. Expanding a single-line import into a parenthesized one carried its `# noqa: F401` onto the new opening line, and `security-suppression-policy` correctly blocked the push for adding a suppression. The directive turned out to be dead, so the fix was to delete it rather than justify it. Without `RUF100` there is no way to know that without testing each directive by hand.

## Proposed fix

Two independent steps; the second is optional and the harder call.

1. Delete the 77 `unused` directives and select `RUF100` so the count cannot grow again. Mechanical, and `ruff check --fix` handles most of it. Note that the three `github_core/__init__.py` entries are one edit plus a `scripts/sync_plugin_lib.py` run, not three.

2. Decide what the 413 `non-enabled` directives are for. They plausibly document intent against linters this repo does not run (`bandit` codes, `flake8` plugin codes). If that is deliberate, `external = [...]` in the ruff config declares those prefixes and `RUF100` stops reporting them, which preserves the documentation and still catches the dead ones. If it is not deliberate, they are noise and should go. This needs an owner decision, so step 1 should not wait on it.

## Out of scope

Not proposing any change to which rules are in `select`. That is a separate argument with a separate blast radius.

## References

- `.claude/rules/code-quality.md`, section "Suppressions Are a Last Resort"
- see `scripts/validation/git_hook_policy.py`, `check_pushed_suppressions`, the gate that surfaced this
- Refs #3781

rjmurillo:
## Summary

Four lint-gate blind spots fixed: RUF100 (unused noqa) was not selected so suppression debt grew invisibly; ruff F811 silently allows duplicate module-level definitions when names begin with underscore; four test methods in a causal-restore test class were trapped after a return statement and never ran; and eval-model-sweep.py carried two dead noqa directives for rules that no longer exist.

Fixes #3792
Refs #3874
Refs #3941
Refs #4060

## Root cause

Issue #3792: RUF100 was absent from the ruff select list so every unused noqa comment accumulated silently. The count reached 464 (409 in non-selected rules plus 55 genuinely unused). Enabling RUF100 and running the auto-fix removed 406 directives and dropped the total violation count from 331 to 58.

Issue #3874: Commit 54449b351 added four test methods to TestAdrReviewPolicyRenameAndBlobScope with an indentation error that placed them inside the body of the helper function _repo_where_a_rename_repadded_the_number, after its return statement. The class header was never written. pytest never collected the four methods; they were dead code.

Issue #3941: Two noqa directives on lines 40 and 365 of the eval script suppressed S404 and S603. S404 was removed from bandit/ruff in a past release. S603 is not a ruff rule at all. Both were no-ops trapping future editors.

Issue #4060: ruff F811 exempts underscore-prefixed names from duplicate-definition detection (treating them as intentional rebinds). A test helper named _foo can shadow an earlier _foo with no warning. The new gate walks module-level scope only, using ast.iter_child_nodes rather than ast.walk to avoid catching inner-class or inner-function duplicates.

## Changes

- tests/validation/test_git_hook_policy_causal_restore.py
- pyproject.toml
- scripts/ci/ruff_count_baseline.txt
- scripts/validation/check_unreachable_after_terminator.py
- tests/validation/test_check_unreachable_after_terminator.py
- lefthook.yml
- tests/test_no_duplicate_module_level_defs.py

(Plus 255 additional Python files with mechanical noqa-removal changes from the RUF100 auto-fix.)

## Verification

Re-measured all premises before writing a line:

- #3792: counted 464 noqa directives (down from 490 in the issue; PR #4023 had already landed). Enabling RUF100 and running the fix dropped the violation count from 331 to 58, a net improvement of 273.
- #3874: ran check_unreachable_after_terminator.py against the pre-fix corpus. Found exactly 1 violation: the 4 test methods inside _repo_where_a_rename_repadded_the_number. The issue claimed "zero today" but measured 1.
- #3941: confirmed both noqa directives still present at lines 40 and 365 in the unmodified file before touching it.
- #4060: confirmed zero live duplicates today. The gate is a regression guard.

Tests:

```
uv run --frozen python -m pytest tests/validation/test_check_unreachable_after_terminator.py tests/test_no_duplicate_module_level_defs.py -q -p no:cacheprovider
44 passed in 2.96s
```

Mutation harness (8/8 killed):

```
uv run --frozen python /tmp/mut_lintgate.py
8 mutants killed, 0 survived
```

Ruff count ratchet:

```
uv run --frozen python scripts/ci/ruff_count_ratchet.py --base-ref origin/main
ruff count ratchet: OK (count == baseline 58).
```

Taste count ratchet:

```
uv run --frozen python scripts/ci/taste_count_ratchet.py --base-ref origin/main
taste count ratchet: OK (count == baseline 612).
```

Type check:

```
uv run --frozen mypy scripts/validation/check_unreachable_after_terminator.py tests/validation/test_check_unreachable_after_terminator.py tests/test_no_duplicate_module_level_defs.py scripts/eval/eval-model-sweep.py
Success: no issues found in 4 source files
```

## References

The bulk noqa-removal commit used SKIP_SCOPE_CHECK=1, the documented bypass in scripts/detect_scope_explosion.py for justified large mechanical auto-fixes. No logic was changed in those 255 files.

The src/copilot-cli mirror trees are generated wholesale from the .claude sources by build/scripts/build_all.py. Adding RUF100 to the per-file-ignore list for those trees prevents the generated mirrors from surfacing false-positive RUF100 violations caused by rules already suppressed at the directory level.

## Closing claim audit

These references do not close their linked items:

- Refs #3874. Issue #3874 is already fixed on main by PR #4074.
- Refs #3941. Issue #3941 is already fixed on main by PR #4106.
- Refs #4060. Issue #4060 is already fixed on main by PR #4186.


rjmurillo:
## Triage: two of the three linked issues are already fixed on `main`

I screened this PR before spending a rebase on it. It is not superseded, but it is no longer the change its description says it is.

| Linked issue | State now | Consequence for this PR |
|---|---|---|
| #3792 (490 `noqa` directives suppress nothing; enable RUF100) | **open** | Live. Verified absent from `main`. |
| #3874 (no gate catches a statement after `return`) | **closed** | Fixed elsewhere. |
| #3941 (`eval-model-sweep` carries dead S404/S603 `noqa`) | **closed** | Fixed elsewhere. |

Evidence for the one that is still live, read off `main` rather than inferred from the issue state:

```
$ git show origin/main:pyproject.toml | grep -nA2 '^\[tool.ruff.lint\]'
101:[tool.ruff.lint]
102-select = ["E", "F", "W", "I", "N", "UP", "B", "ANN"]
103-ignore = []
```

`RUF100` is not in that select list, so the noqa-audit work this PR exists for has not landed by another route. I checked the detector before trusting the negative: the same `git show` returns 257 lines and locates the `[tool.ruff.lint]` block, so the empty match is a real absence and not a broken command.

## Why I am not just rebasing it

Three things compound badly here:

1. **259 files.** The blast radius is large enough that a rebase across a moved `main` is likely to conflict in many places, and each resolution is a chance to silently revert someone else's newer fix.
2. **20 commits, exactly at the repository cap.** There is no headroom to add resolution commits without tripping the limit.
3. **Two thirds of the change is now redundant**, so most of that conflict-resolution cost would be spent re-landing work `main` already has.

There is a fourth risk that no textual conflict will warn about: merging a stale base can silently overwrite a newer state with an older one when both sides parse cleanly. I hit exactly that on #4164, where a ratchet baseline fixture from 27 commits back would have merged without a single conflict marker.

## Recommendation

Split this down to the RUF100 change alone and reopen it against current `main`. That PR is small, has one live issue behind it, and can be reviewed on its merits. The unreachable-code gate and the `eval-model-sweep` cleanup should be dropped rather than rebased, because `main` already has them.

I have left this open rather than closing it, because unlike #4164, #4102, and #3979, a real piece of work still sits inside it. Closing it would lose the RUF100 change; merging it as-is would re-litigate two settled fixes across 259 files.

Filed #4355 on the general pattern: nothing in this repository surfaces that a PR has been overtaken by `main` until somebody spends the rebase to find out.

Issue #4283, opened 2026-08-02
Global push lock serializes the fleet: the race is per-ref, not per-repo


rjmurillo:
## Summary

The fleet serializes every `git push` behind one global `flock /tmp/aiagents-push.lock`. The race that lock defends against is per-ref, not per-repo, so the lock is scoped one level too wide. It costs throughput and buys no safety that a per-branch lock does not already buy.

## What the lock currently is

Recorded as accepted practice in the test-infrastructure retrospective:

> The `flock /tmp/aiagents-push.lock` serialization prevents concurrent push races but the pre-push hook takes 7+ minutes, so main can advance during that window.

It is an operator convention, not repo tooling. Nothing in `lefthook.yml` or `scripts/` references it.

## The race it defends against

Issue #3882 documents the failure it exists to prevent:

```
! [remote rejected] <branch> -> <branch> (cannot lock ref 'refs/heads/<branch>': is at 89d667b76 but expected 4454aa7d4)
```

That message names one ref. Git takes its lock per ref. Two pushes to two different branches cannot produce it, because they never contend for the same lock file on the server. Concurrent object writes are not a hazard either: git writes them to temp files and renames.

So the global lock protects branch A's push from branch B's push, which were never in conflict.

## Measured cost

Observed five queued pushes to five distinct branches, each behind an approximately 11 minute pre-push hook:

```
41:27  flock /tmp/aiagents-push.lock git push --force origin fix/test-locale-codec
28:15  flock /tmp/aiagents-push.lock git push origin fix/pre-pr-runs-ratchets
21:02  flock /tmp/aiagents-push.lock git push -u origin fix/test-infrastructure-v2
06:06  flock /tmp/aiagents-push.lock git push -u origin fix/eval-record-correctness
```

Four of them were waiting on a lock protecting nothing they could collide with. The one at 28 minutes had no child process at all: it had not started work, it was queued. Serialized, `n` pushes cost `n * 11` minutes. In parallel they cost 11.

## Measured safety

Ran one push under a per-branch lock while three others held the global lock:

```
ba4ad6a367..e7a0e00b77  chore/measurement-validity-rule -> chore/measurement-validity-rule
REAL_EXIT=0
```

662 seconds, concurrent with three global-lock holders, no ref conflict.

## Proposed change

Replace the global lock with a per-branch one, which still serializes the only contention that exists (two pushes to the same branch, the #3882 case) while letting distinct branches proceed in parallel:

```bash
[...]

PR #4296, opened 2026-08-02
docs(memory): record six fleet session traps, including stale PR merge refs

rjmurillo:
### What

Adds five Serena memories and extends a sixth, all recording traps measured
during a long multi agent fleet session. Documentation only. No code, no
workflow, no gate behaviour changes.

### Why

Each entry cost real debugging time in session, and each is re derivable only by
repeating the same wrong turn. Six files:

| File | Records |
|---|---|
| `ci/ci-a-red-check-on-your-pr-may-be-inherited-from-main.md` | A red check on your PR is not evidence the fault is in your PR, plus the merge ref mechanism below |
| `ci/ci-file-size-ratchet-line-numbers-shift.md` | Why a file size ratchet violation moves when unrelated lines are added above it |
| `ci/ci-which-parity-gate-missed-an-agent-change.md` | Which parity gate catches a one copy change to a mirrored tree, and which does not |
| `bash/bash-anchor-grep-patterns-on-test-logs.md` | Anchoring grep patterns on test logs so a substring cannot match the wrong field |
| `coding/coding-size-ceiling-on-a-registration-list.md` | A size ceiling on a registration list measures the domain, not the code |
| `git/git-detached-push-dies-if-log-dir-vanishes.md` | A detached push dies silently when its log directory is removed underneath it |

The largest entry carries the finding that cost the most: **fixing `main` does
not fix your PR until the merge ref moves.** `refs/pull/N/merge` is recomputed
on a branch push, not on a rerun, not on a PR body edit, and not on a
`gh api .../pulls/N` read. Measured after the taste ratchet fix landed on main:
the one PR that had pushed a merge commit afterwards was fresh, and three that
had not were still stale sixteen minutes later, all four reporting the same
already fixed failure.

The same entry warns that `gh run view --json headSha` on a `pull_request` run is
the branch tip rather than the merge commit, so it cannot distinguish a stale
merge ref from a fresh one. That misreading is what produced the wrong diagnosis
the memory now corrects.

One commit is a retraction rather than an addition: `9eff6720f0` removes a
superseded global push lock recipe. A single global lock serialised every agent
in the fleet behind one push. The per branch lock that replaced it is recorded
in its place.

## Specification References

| **Issue** | Refs #4283 | Per branch push lock, whose retraction this documents |

Documentation PR, so no spec is required per the template guidance.

## Testing

Markdown only. Verified no em dashes or en dashes across all six files, and
`taste_lints.py` exits 0 on the changed set. `git merge origin/main` reports
already up to date, so the branch carries no conflict.

## Notes for Reviewers

The value here is negative results. Three of the entries exist to stop a future
agent from repeating a probe that looks authoritative and proves nothing. If a
claim in any entry reads as over general, say so: these were written from a
single session and are stated with the population they were measured on.

### DIFF
--- a/.serena/memories/bash/bash-anchor-grep-patterns-on-test-logs.md
+++ b/.serena/memories/bash/bash-anchor-grep-patterns-on-test-logs.md
@@ -112,16 +112,46 @@ git merge-base --is-ancestor origin/<branch> HEAD && echo "local is ahead or equ
 git rev-parse HEAD origin/<branch>     # equal means the work is on the remote
 ```
 
+Do not read the pre-push hook's success banner as push completion. This repo's
+hook prints
+
+```
+RESULT: All validations passed
+
+Ready to create pull request!
+```
+
+and that text appears in the push log roughly two minutes in, while the ref
+transfer has not started. Misread once on 2026-08-02: `ls-remote` came back
+empty right after that banner, which looked like a failed push and was actually
+a push still running. The only completion signals are a non empty
+`git ls-remote origin <branch>` matching local `HEAD`, or the `REAL_EXIT=` line
+the wrapper appends after `git push` returns.
+
 Better than detecting the race is preventing it. Serialize pushes through a
-lockfile, which also keeps concurrent agents from running several full pytest
-suites at once on the same machine:
+lockfile, keyed on the branch:
 
 ```bash
-flock /tmp/aiagents-push.lock git push -u origin <branch>
+BR=<branch>
+SLUG=$(printf '%s' "$BR" | tr '/' '-')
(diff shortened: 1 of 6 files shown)


rjmurillo:
owner: local:pr-autofix
session: session-4264
acquired_at: 2026-08-02T19:31:22Z
expires_at: 2026-08-02T19:46:22Z
base_sha: ede9fd1fef9d28d2a96f8071ddbeea9075e064d2

rjmurillo:
owner: local:pr-autofix
session: session-4264
acquired_at: 2026-08-02T19:57:09Z
expires_at: 2026-08-02T20:12:09Z
base_sha: e3877c18f8c6804594f6ab4379f52d056e637949

Issue #4057, opened 2026-07-30
fix(ci): prevent concurrent ratchet improvements from leaving main stale


rjmurillo:
## Problem

Exact count ratchets fail when the measured count falls below the committed baseline. Two concurrent PRs can each remove one violation and lower the same one-line baseline by one. Both branches can pass. Their identical baseline edits merge without conflict, but the combined tree removes two violations while the baseline falls once. `main` then fails as stale.

## Expected

Required checks must evaluate the actual merge tree against the baseline that will land, or branch policy must require a current base before merging ratchet changes.

## Evidence

PR #4023 changed stale baselines from success to regression. Independent review identified the concurrent merge case. Current pull request workflows do not prove that a previously green ratchet check reruns after another baseline-changing PR lands.

## Acceptance criteria

- Add a two-branch regression test for identical baseline edits and cumulative count reductions.
- Choose one enforcement point that blocks the second stale merge.
- Keep baseline increases blocked against the base ref.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #4057: fix(ci): prevent concurrent ratchet improvements from leavin
- **Type**: complexity_triage
- **Priority**: medium
- **Suggested Agent**: agent-qa
- Issue labeled with low complexity.
Suggested agents: agent-qa

### Route issue #4057 to agent-qa
- **Type**: agent_routing
- **Priority**: low
- **Suggested Agent**: agent-qa
- Issue content suggests routing to: agent-qa
Complexity: low

rjmurillo:
Live instance on `main` right now, and it runs in the opposite direction from the one this issue describes.

The body covers two PRs each *removing* a violation, dropping the shared baseline once, leaving `main` stale. The same race also runs the other way, and the other way is worse.

## What happened

`cea600ed` (#4093) introduced the type-ignore count ratchet with baseline 55. That was correct for the tree at that commit. Counting `#\s*type:\s*ignore(?:\[[^\]]*\])?` across tracked `*.py` at `cea600ed` gives exactly 55.

`08829297` (#4091) landed separately and added one suppression to `scripts/validation/pr_commit_count.py:238`. That PR's base predated the gate, so nothing in its pre-push run could have objected.

Both merged. `origin/main` now counts 56 against a committed baseline of 55:

```
$ uv run python scripts/ci/type_ignore_count_ratchet.py
[...]

Issue #4285, opened 2026-08-02
pre_pr_sequence.py is a registration list measured by a complexity ceiling; extraction only resets the counter


rjmurillo:
## Summary

`scripts/validation/pre_pr_sequence.py` holds one function whose body is 48 ordered `run_validation(...)` calls. Its line count tracks how many gates the project has, not how hard the module is to read, so the 500 line `file-size` ceiling fires on it for a reason the rule's remediation cannot address.

## The pattern, from the file's own docstring

> Ordered pre-PR validation sequence (extracted from `pre_pr.py`, Issue #3073). [...] Extracted so `pre_pr.py` stays under the module size ceiling while a new governance gate is added.

So this file exists because `pre_pr.py` hit the same ceiling. The extraction reset the counter without reducing anything, and the extracted file has now hit the ceiling itself. A second extraction would do the same thing a third time.

## Measured

Adding the count ratchets to the sequence (PR #4272) added 12 lines and crossed the line:

```
taste count ratchet: REGRESSION. 602 violations > baseline 601 (+1).
```

Diffing the full error-severity violation sets between `origin/main` and the branch isolated it to one file:

```
NEW on branch:  scripts/validation/pre_pr_sequence.py:512:file-size
```

The other five deltas were the same violations at shifted line numbers, because `file-size` reports at the line where the file crosses the ceiling and that line moves as a file grows.

Trimming the added comment to its minimum (the rationale was duplicated from the `checks_ratchet` module docstring, so removing it was a DRY fix worth making anyway) brought the file to 509. Still over. Getting to 500 requires touching code the PR does not own.

## Why the rule's remediation does not apply

The `file-size` remediation text proposes extracting helper functions. For a linear registration sequence, extraction moves entries to a second file and renames the problem. It also hides the one property the file exists to express: ordering. The sequence is deliberately ordered so the cheapest push-blocking signal arrives first, and splitting it across modules makes that ordering non-local.

## Proposed fix

Replace the 48 call sites with a table and one loop:

```python
_SEQUENCE: tuple[tuple[str, Callable[[Path], object]], ...] = (
    ("Python Syntax (compile gate)", validate_python_syntax),
    ("Count Ratchets", validate_count_ratchets),
    ...
)

for name, check in _SEQUENCE:
    run_validation(name, state, lambda c=check: c(repo_root))
```

[...]

PR #4302, opened 2026-08-02
fix(validation): unbreak main by lowering the taste count, not the bar

rjmurillo:
## Problem

Pristine `origin/main` fails the taste count ratchet at 602 violations against a baseline of 601. The taste ratchet runs in the pre-push lefthook group, so every branch in the repository was push-blocked, and four PRs (\#4256, \#4271, \#4273, \#4284) were failing `Run Python Tests` on an inherited failure none of their authors caused.

Two independent investigations reached this root cause separately, which is the corroboration this claim rests on rather than a single measurement.

## Root cause

The baseline is a bare count, so a direct diff cannot say which file moved. Bisected per-file instead: the baseline was last set by `aea3a49cd`, and diffing per-file taste errors between that commit and main shows `scripts/validation/pre_pr_sequence.py` going from 0 errors to 1. It had grown past the taste-lint 500-line file ceiling.

PR \#4272, titled "run the count ratchets in the pre-PR gate", is the PR that broke the count ratchet.

It merged green because the default-branch ruleset has `strict_required_status_checks_policy` set to false, so a PR merges without re-testing against the main it will land on. That gap is already tracked as issue \#4057; this PR does not attempt to close it.

## Changes

Fix by extraction, not by raising the baseline. The ratchet's own rule is that a baseline may only fall, and a suppression comment was considered and rejected: the repo's own code-quality rule treats suppressions as a last resort, and nothing here made the idiomatic fix impossible.

- `scripts/validation/pre_pr_sequence.py` sheds the six whole-tree sub-second gates, going from 512 lines to 455. Emission order is unchanged.
- `scripts/validation/pre_pr_sequence_fast.py` is new and holds `run_fast_gates` plus the `ValidationStateLike` protocol.
- `tests/ci/test_pre_pr_runs_lefthook_ratchets.py` gains two tests that guard the property the new module name claims.

The grouping is cost and blast radius, not convenience. These six are the gates cheap enough to always run, and their value is reporting before anything slower. `run_fast_gates` deliberately takes no `args` because none of them is skippable; threading the namespace through would invite one to start.

## Testing

Both new tests are guards against different failure modes, because one alone would be a tautology.

The prefix test reads its expectation from production source, so it cannot assert a stale list, but that also makes it blind to a reorder inside the group. That blindness is covered by the pre-existing `test_gate_runs_second`. What source-reading cannot catch is a gate moved out of the group into the slower tail, so the second test pins group membership against a hard-coded list.

Negative control: a mutant that moves `Test Working Tree Writes` out of `run_fast_gates` into the caller, leaving the emitted gate sequence byte-identical and changing only the grouping, is caught by both new tests while all 16 pre-existing tests in that file pass it. A test that a plausible mutant survives is not evidence, so this is the check that makes the new tests load-bearing.

Ratchet trio, run against `origin/main`:

```
taste count ratchet: OK (count == baseline 601).
type-ignore count ratchet: OK (count == baseline 44).
ruff count ratchet: OK. 323 violations <= baseline 326 (-3 slack).
```

The same taste command reports `REGRESSION` at 602 on pristine main, which is the before half of that pair.

Full pre-push suite passed on this branch, including `python-tests` at 839 seconds and `taste-count-ratchet` at 2.26 seconds. 124 tests pass across the two pre-PR suites. `ruff check` and `ruff format --check` are clean.

Refs \#4272
Refs \#4057

### DIFF
--- a/scripts/validation/pre_pr_sequence.py
+++ b/scripts/validation/pre_pr_sequence.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 import os
 import sys
 from pathlib import Path
-from typing import TYPE_CHECKING, Protocol
+from typing import TYPE_CHECKING
 
 _SCRIPT_DIR = Path(__file__).resolve().parent
 if str(_SCRIPT_DIR) not in sys.path:
@@ -27,10 +27,6 @@ from active_plan_closeout import validate_active_plan_closeout
 from check_doc_interpreter_portability import (  # noqa: E402
     validate_doc_interpreter_portability,
 )
-from check_duplicate_test_helpers import validate_duplicate_test_helpers
-from check_nested_tests import validate_no_nested_tests
-from check_test_tree_writes import validate_test_tree_writes
-from check_unreachable_code import validate_unreachable_code
 from checks_coverage import (  # noqa: E402
     validate_review_marker,
 )
@@ -44,7 +40,6 @@ from checks_plugin import (  # noqa: E402
     validate_shipped_skill_routes,
     validate_workflow_local_run,
 )
-from checks_ratchet import validate_count_ratchets  # noqa: E402
 from checks_spec import (  # noqa: E402
     validate_agent_catalog,
     validate_build_gates,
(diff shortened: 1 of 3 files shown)


rjmurillo:
Closing this. It is the wrong fix, and issue \#4285 says so in its title: "pre_pr_sequence.py is a registration list measured by a complexity ceiling; extraction only resets the counter."

That is precisely what this PR does. It moves 57 lines into a new module and the counter drops below the ceiling. Nothing got simpler. `run_all_validations` is still a 408-line function afterward, just 350 lines instead. I would have rejected this reasoning in someone else's PR, so I am rejecting it in mine.

Two things also overtook it:

PR \#4290 already unblocked main by suppressing the violation with a written rationale, and I verified main green on a pristine probe: `taste count ratchet: OK (count == baseline 601)`. So the P0 this PR was opened for no longer exists.

The suppression's rationale is correct on the merits, not just expedient. The module holds one function whose body is 48 ordered `run_validation` calls. Its line count tracks how many gates the project has, not how hard it is to read, so a file-size ceiling is measuring the wrong thing here.

What the ceiling is a proxy for is real, though: a 408-line function violates this repo's own "methods 60 lines or fewer" rule by roughly 7x. The fix that addresses that, rather than relocating it, is the table-driven registry in \#4285. This repo's own code-quality rule prescribes exactly that shape: replace a long sequence with a table. Picking that up next.

Salvaged from this branch: nothing. Both new tests assert against `run_fast_gates`, which will not exist under a registry.

Refs \#4285
Refs \#4290

PR #4314, opened 2026-08-02
docs(rules): name the stale-pycache trap that survives byte-level mutation checks

rjmurillo:
## Summary

Adds one SHOULD to `.claude/rules/testing.md` (plus its two generated mirrors) naming a
mutation-testing failure that survives every byte-level "did the patch apply" check.

CPython validates a cached `.pyc` against the source's modification time **in whole
seconds** and the source's **size**. A mutation that preserves size and is reverted
inside the same second therefore leaves the cache valid, and the restored file still
executes as the mutant. Every byte-level check passes, because the bytes really did
change. The interpreter simply never read them.

Refs #4244.

## Why this is not already covered by MUST 7

MUST 7 governs how a harness detects that a patch applied. This is the separate case
where detection succeeds and execution still uses the old code. The two are independent:
you can satisfy MUST 7 completely and still measure the wrong binary.

## Evidence

Probe with paired positive and negative controls, CPython 3.14.6. Source strings are
the same byte length, so size is preserved; `os.utime` pins mtime back to simulate a
same-second revert.

| Case | Cache | Source | Executed |
|---|---|---|---|
| baseline | (written) | `AAAA` | `AAAA` |
| mutated, mtime+size preserved, cache kept | `AAAA` | `BBBB` | **`AAAA`** |
| same mutation, `__pycache__` deleted | (none) | `BBBB` | **`BBBB`** |

Negative control (row 2) shows the mutation is invisible. Positive control (row 3)
shows the same mutation is visible once the cache is gone. The remedy is deleting the
cache directory.

Second probe, for the sub-claim that the usual bytecode switches do not help. In every
row the cache holds `AAAA` while the source holds `BBBB`, so the two answers differ and
the control discriminates:

| Invocation | Executed |
|---|---|
| plain `python` | `AAAA` |
| `python -B` | `AAAA` |
| `PYTHONDONTWRITEBYTECODE=1` | `AAAA` |
| both | `AAAA` |

All four read the stale entry. Those switches suppress *writing* new bytecode; they do
not stop the interpreter from *reading* a cache entry it considers valid. `pytest`'s
`-p no:cacheprovider` is unrelated: it governs `.pytest_cache`.

## A note on how this was verified

The first version of the probe had a fourth row that ran `python -B` in a state where
the stale cache and the fresh source held the *same* value. It returned the mutant
string and appeared to falsify the sub-claim. It could not discriminate: both answers
were identical, so the row passed for the wrong reason and nearly caused a correct
claim to be retracted.

That is precisely the failure mode SHOULD 5 and SHOULD 6 in this same file describe. It
seemed worth recording that the rule caught its own author.

## Scope

Documentation only. One added line to the canonical rule and the two generated mirrors,
regenerated rather than hand-edited. No behavior change, no script change.


rjmurillo:
Pushed `7419a67de` plus a merge of `main`.

Adversarial review on a different model family found that the supporting probe tested the
mutate direction while the rule text described the revert direction. Both are real and the
consequences differ, so the rule now names each:

- **Mutate direction.** A mutant appears to survive a test that would have killed it.
- **Revert direction.** A reverted file still runs as the mutant.

Re-ran the revert direction on CPython 3.14.6 with real writes and no `utime` call, so the
timing is a genuine same-second edit rather than a synthesized mtime. It reproduced 10 of 10.
An independent reviewer reached 10 of 10 on a separate implementation.

Also added that PEP 552 `CHECKED_HASH` invalidation does detect the change, verified, with the
caveat that it is not the default. Without that line a reader could conclude no invalidation
mode catches this.

Instruction budget unchanged at 82603 of 83000. The rule is scoped to `tests/**`, so it does
not touch the always-on ceiling.

Issue #3989, opened 2026-07-30
eval: coordinate-wise median can publish a score triple no judge gave


rjmurillo:
## Summary

`_reduce_score_samples` in `scripts/eval/eval-rule-activation.py` takes the median of each score field independently across judge samples. That is a coordinate-wise median. The resulting triple need not correspond to any sample the judge actually produced, and it can be strictly better than every one of them.

Found by GPT-5.6 Sol in adversarial review round 10 of the salvage/parse path. Independently reproduced.

## Reproduction

Three judge samples, each rating the same cell:

| sample | activation | citation | behavior | sample mean |
|---|---|---|---|---|
| 1 | 5 | 5 | 1 | 3.67 |
| 2 | 5 | 1 | 5 | 3.67 |
| 3 | 1 | 5 | 5 | 3.67 |

Per-field medians: activation 5, citation 5, behavior 5. Cell score `5.0`.

Every judge rated this cell 3.67. The instrument publishes 5.0. No judge ever gave a 5/5/5.

The direction is not symmetric in practice: disagreement among judges on *which* axis is weak produces a cell that looks unanimous and strong.

## Measured effect on the published numbers

Walked all 96 cells in `.agents/analysis/eval-artifacts/2026-07-29-unified-software-engineering/`. Comparing the coordinate-wise median against reducing each sample to its own three-field mean and taking the median of those scalars:

- **3 of 96 cells diverge.**
- Worst divergence: **0.333 on a cell**, **0.111 on a run average**.
- Divergent cells: `t-sol56` S2/description (4.333 to 4.0), `t-sol56` S3/baseline (2.333 to 2.667), `var-sol-2` S3/full (4.0 to 3.667).

Recomputed end to end under the sample-scalar reduction, two of eight published rows move:

- `t-sol56`: `4.00 / 3.67 / 3.56`, delta-full **-0.44** (was -0.33)
- `var-sol-2`: full **3.89**, delta-full **+0.67** (was +0.78)

**Sign test is unchanged: 7 positive, 1 negative, two-tailed p = 0.0703125.** The published conclusion does not move. This is a correctness defect in the instrument, not a retraction of the result.

## Why it matters anyway

[...]

PR #4354, opened 2026-08-02
fix(rules): recover 990 bytes of always-on headroom by deduplicating pragmatic-programmer

rjmurillo:
## Summary

Recovers 990 bytes of always-on instruction headroom by deduplicating
`pragmatic-programmer.md` against itself. Main currently has **8 bytes** of
headroom. This takes it to **998**.

Refs #3419

## This PR changed shape while it was open, and the title was wrong

It was opened as a two-axis unblock: a pytest marker-citation fix plus this
budget work, at a time when `main` was red and over ceiling. While it sat in
review, #4351 landed the same combined fix. The marker-citation half of this
branch converged to a zero-byte diff against `main`, and so did two skill
citation fixes that were in the original 7-file diff.

What survives is 3 files and one job. The old title claimed to unblock `main`.
`main` is green. Retitled and rewritten rather than left to mislead a reviewer.

## The number that matters

| Tree | Bytes | Ceiling | Headroom |
| --- | --- | --- | --- |
| `main` at `160d1670a` | 82992 | 83000 | **8** |
| This branch | 82002 | 83000 | **998** |

Measured with `scripts/validation/instruction_budget.py` in two worktrees, one
detached at `FETCH_HEAD` for `main` and one on this branch. Eight bytes is
roughly one word. The next contributor who adds a sentence to any always-on
rule breaks every push in the repo, because `pre_pr.py` runs this check in the
pre-push hook, not only in CI.

## What was removed, and why it is safe

Three sections of `pragmatic-programmer.md`, all restatements of content the
same file already carries:

**`## Final Instruction`** is a near-verbatim duplicate of `## Primary
Directive`. Both open with the identical sentence "When uncertain, choose the
option that:" and then give a five-item list:

| `## Primary Directive` | `## Final Instruction` |
| --- | --- |
| reduces knowledge duplication | removes duplicated knowledge |
| keeps concerns independent | keeps concerns orthogonal |
| shortens feedback loops | shortens feedback loops |
| leaves the system easier to change | leaves the codebase easier to change tomorrow |
| makes intent clearer to future maintainers | improves automation |

**`## Review Rules for Claude`** is a nine-bullet index into the file's own
earlier sections: "duplicated knowledge" is `## DRY Rules`, "hidden couplings"
is `## Orthogonality Rules`, "missing automation opportunities" is `## Automation Rules`, "long feedback loops" is `## Feedback Loop Rules`, "unclear
contracts" is `## Design by Contract and Assertions`. The file keeps `## Review
Checklist`, which is the actionable form of the same idea, phrased as questions
you can answer yes or no.

**`## Code Generation Rules`** restates the same seven principles a third time.

No principle is lost. Every one of them still appears in the section that
defines it, which is where a reader looking it up would go.

## The part worth noticing

The rule file that teaches "DRY at the knowledge level, not just duplicated
lines" stated its own primary directive three times. That is the failure mode
it warns about, in the artifact that warns about it. Worth saying out loud,
because the same shape is likely elsewhere in the always-on set.

## What this deliberately does not do

- **Does not raise `DEFAULT_CEILINGS_BYTES`.** Moving the ceiling to fit the
  content inverts the gate. The budget exists because always-on bytes cost
  instruction-following accuracy on every turn.
- **Does not cut repo-specific content.** Everything removed is a restatement.
  Facts a model cannot know, local conventions, and arbitration between
  competing principles all stay.
- **Does not touch the other always-on rules.** `builder-ethos`,
  `claude-model-patches`, `lsp-first`, `search-before-building`, `universal`,
  and `voice` are the other five. Some probably carry the same redundancy, but
  that is a separate measurement and a separate PR.

## Verification

- `scripts/validation/instruction_budget.py`: `.md 82002/83000 98.8% PASS`
- Both mirrors regenerated with `build/scripts/generate_rules.py`; no drift.
- `build/scripts/validate_plugin_version_bump.py`: OK.
- Green `main` merged in, not rebased. No force-push.

## Reviewer note

The three files are one logical edit: the canonical rule under `.claude/rules/`
and its two generated mirrors. Do not hand-edit the mirrors; regenerate.


rjmurillo:
## This PR proposes the one change this repo has actually measured, in the direction the measurement went against

The premise here is that cutting an always-on rule body buys headroom worth more than the text costs. That exact question was measured on this repo in July, on Opus 5 and Sol 5.6, and the result is published on `main`:

`.claude/skills/context-optimizer/references/rule-audit-evidence.md:104`

> The headline survives all of it. The sign test is 7 positive to 1 negative, p = 0.0703125, identically under the pre-recovery table, the post-recovery table, and either reduction in issue #3989.

"Positive" is a positive `delta-full`, meaning the **full rule body scored higher** than the cut variant. So the full body won 7 of 8 runs.

Two honest limits, because they matter more than the headline:

- **p = 0.0703125 is not significant at 0.05.** This is not proof that cutting harms. It is the only measurement that exists, and it points the opposite way from this PR's premise.
- The result carries a stated confound: all 24 judge failures sat in the four Opus artifacts, and the Opus rows carry the large positive deltas the headline leans on. That is documented at lines 23 to 41 of the same file.

`pragmatic-programmer.md` is a book-derived always-on rule (Hunt and Thomas), which is the class that was measured. The measured benefit of cutting is not established. The stated benefit here is 990 bytes.

## The deletion is not a deduplication

The diff removes two sections whole:

- **"Review Rules for Claude"**, 9 review criteria (duplicated knowledge rather than duplicated lines, hidden couplings, missing automation opportunities, long feedback loops, local fixes that worsen changeability, unclear contracts, non-repeatable manual processes, brittle integration points, code that communicates poorly).
- **"Code Generation Rules"**, 7 defaults and 4 avoid-by-default items.

The compensation is one added line to the primary directive ("automates what is repeated") and a reworded Review Checklist header ("When reviewing code, and before finalizing any change, verify:"). The retained checklist does not carry the 9 review criteria or the 11 generation defaults. Twenty items are replaced by one line and a header edit, so this is a cut, not a dedup, and the title's "deduplicating" is inaccurate.

## What would change my read

Run the eval harness on this specific diff. The instrument exists, it is wired for Opus 5 and Sol 5.6, and it is what produced the number above. If this cut scores neutral or positive across runs, that is a real result and it should land on the evidence of it rather than on a byte count.

Absent that, my recommendation is to close this, or narrow it to whatever text is genuinely redundant with a sibling rule, quoted side by side so a reviewer can check the redundancy claim directly.

## Also worth checking before any rework

This overlaps merged PR #4351. Whatever remains unique here should be rebased onto current `main` first, so the diff shows only the delta.

Issue #4458, opened 2026-08-03
fix(governance): .md instruction ceiling was raised to absorb a breach in an unrelated PR, reversing the recover-bytes precedent set on #4394


rjmurillo:
## Problem

The always-on `.md` instruction ceiling was raised from 83000 to 84000 in PR #4107, a PR whose title and body are entirely about extending the portability ratchet. The raise is not mentioned anywhere in that PR body. It absorbed a pre-existing 516 byte breach on `main` instead of recovering the bytes.

This reverses the resolution chosen on #4394 three days earlier. That issue was closed with an explicit rationale: "PR #4351 plus commit 751393343 recovered 990 bytes. No code change needed." The precedent was recover, not raise. #4107 raised.

## Evidence

Measured on a clean worktree at each commit, using `uv run --frozen python scripts/validation/instruction_budget.py`.

| Commit | Corpus bytes | Ceiling | Usage | Status |
|---|---|---|---|---|
| `5051160e2` (#4107 parent) | 83516 | 83000 | 100.6% | FAIL |
| `501150e2f` (#4107 squash) | 83516 | 84000 | 99.4% | PASS |

The numerator does not move. The full diff of the constants file in that PR:

```
-    ".md": 83_000,
+    ".md": 84_000,
```

PR #4107 file list includes `scripts/validation/instruction_budget_constants.py`. A regex over its body for `budget|ceiling|83.?000|84.?000` returns no match.

## How the corpus got there

Walking `main` backward, the corpus climbed 82002 to 82630 to 82942 to 82992 to 83201 to 83516. It crossed 83000 more than once. #4394 caught one crossing and it was fixed by cutting content. The next crossing was absorbed by the ceiling.

## Why this matters

The repository's own always-on guidance states the rule this violates:

> A budget that ratchets upward to track measured size only proves the file has not grown since someone last raised the ceiling. Delete on the test, not on the byte count.

Current headroom is 484 bytes. At the observed growth rate the ceiling will be crossed again within days, and the cheapest available action will again be a one line constant edit.

## Not a duplicate

- #4394 (closed) was one specific breach, resolved by recovering bytes. This is about the later breach being resolved the opposite way.
- #4345 (open) is about the CI path filter reporting success without measuring. That is a detection defect. This is a remediation defect. They compound: the filter hides the breach, and when it does surface, the ceiling absorbs it.

## Suggested direction, not a prescription

Two options, and the choice is a judgment call for the maintainer:

[...]

PR #4459, opened 2026-08-03
docs(memory): record how to read PR checks and which side of the budget ratio moved

rjmurillo:
## Summary

Two Serena memories recording measured failure modes hit while draining the PR queue. Both cost real time this session and both are cheap to avoid once written down.

The first records that reading PR checks through the raw status rollup misses severity and re-run supersession. Three hand-rolled readers were built before finding that the repository already ships the correct one, and none of the three reported required status at all.

The second extends an existing stale-branch memory with the instruction-budget case, where the gate reports a percentage rather than a file list, so a branch gets blamed for a corpus it never touched.

## Changes

- `.serena/memories/pr-review/use-get-pr-checks-not-raw-rollup.md`: new memory. Severity and re-run supersession outrank pagination when reading PR checks. Names the canonical reader and the nesting under the `Data` key that silently returns `None` for every merge-decision field when read off the top level.
- `.serena/memories/git/git-stale-branch-fails-repo-state-tests.md`: extend the existing memory with the instruction-budget instance. Records the measurement that the corpus was identical at 83516 bytes on both sides of the merge while the ceiling moved from 83000 to 84000, so the branch was never the cause. Adds the fleet-load staleness window.
- `.serena/memories/memory-index.md`: regenerated index.

## Verification

`uv run --frozen python scripts/validation/memory_index.py --path .serena/memories --ci --format console` passes: 305 files indexed, 0 missing.

`uv run --frozen python scripts/validation/pre_pr.py` reports all validations passed. Zero em-dashes or en-dashes.

The corpus-versus-ceiling claim is measured, not inferred: the same command was run at the merge commit and its parent and the numerator was byte-identical.

## References

Refs #4458


.serena/memories/pr-review/use-get-pr-checks-not-raw-rollup.md
rjmurillo, line 1: Fixed in d94896e71. The paragraph now reads: a CANCELLED result with at least one sibling SUCCESS is a superseded run, not a failure. A CANCELLED result with no sibling SUCCESS still surfaces as a failure. The blanket-rule reading is corrected.

Issue #4416, opened 2026-08-03
metrics.files_changed sums file counts from work-log prose the session did not author


rjmurillo:
## Summary

`metrics.files_changed` sums every "N files changed" phrase it finds in work-log prose. Prose legitimately cites file counts the session did not author: a diff against a stale base, a test-suite summary, an upstream merge. Issue #3363 established that work-log prose is not a trustworthy source for `metrics.commits` and demoted it to a fallback. The same reasoning applies to `files_changed`, but that field still reads prose unconditionally.

## Evidence

Session `2026-08-02-session-4257-testing-discriminating-input` made exactly one commit, `1f5b71b0a`, touching **3** files:

```
$ git show --stat --format="" 1f5b71b0a | tail -1
 3 files changed, 3 insertions(+)
```

The extracted episode recorded **35**. The source is a single work-log sentence:

> "git diff against the origin/main tip showed 35 files and 3001 deletions, which reads as a mass ..."

That sentence describes a stale-base artifact the session was diagnosing. It is not a count of what the session changed. `_FILES_RE` matched it and `metrics["files_changed"] += 35` took it.

Reproduced directly:

```python
for e in _as_list(data["workLog"]):
    f = _FILES_RE.search(_entry_text(e))
    if f: print(f.group(0))
# MATCH: 35 files
```

## Why it matters

`files_changed` is one half of the condition the episode-store ratchet tests (`commits == 0 and files_changed > 0`). An inflated `files_changed` makes that condition fire on sessions that would otherwise be clean, so a prose artifact can block a push.

It also lands in a committed artifact that reads as a record of what happened. The number is wrong by an order of magnitude here.

## Contrast with the commits fix

`_collect_shas` docstring already states the principle:

> "Prose is not a primary source (issue #3363). A work-log entry legitimately cites SHAs the session did not author ..."

The accumulation at `_json_metrics` never got the same treatment.

## Suggested direction

Not proposing an implementation. The structural parallel to #3363 is to prefer a computed count (`git diff --name-only <startingCommit>..<endingCommit>`) and fall back to prose only when no commit range resolves. That also fixes the case where prose names no count at all.

Refs #3363

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #4416: metrics.files_changed sums file counts from work-log prose t
- **Type**: `complexity_triage`
- **Priority**: medium
- **Suggested Agent**: agent-qa
[...]

Issue #3617, opened 2026-07-28
extract_session_episode: metrics.files_changed is scraped from any 'N files' prose in the work log


rjmurillo:
## What

`_FILES_RE = re.compile(r"\b(\d+)\s+files?\b", re.IGNORECASE)` (see `.claude/skills/memory/scripts/extract_session_episode.py` line 394) matches any "N files" phrase in a work-log entry, and `json_metrics` sums every match into `metrics.files_changed`.

## Why it bites

Work-log prose legitimately contains file counts that have nothing to do with the commit. A session that wrote "markdownlint reported Linting: 2 files, 0 issues" set `files_changed` to 2.

Worse, the staged-diff backfill only runs when the value is falsy:

```python
if not metrics.get("files_changed"):
    staged = _staged_files_changed(session_log_path.parent)
```

So a wrong prose-derived value suppresses the correct staged-diff value. The error is sticky: `_merge_metrics` takes a per-key max, so a preserve regeneration cannot lower it.

## Prior art

This is the same class as the SHA-in-prose problem already fixed for `metrics.commits`, where prose was demoted to a fallback behind `endingCommit` and the `changesCommitted` evidence. `files_changed` never got the same treatment.

## Proposed fix

Prefer the staged diff, and treat work-log prose as a fallback only when the staged diff yields nothing. Mirror the primary/fallback split `_collect_shas` already uses.

## Repro

Add a work-log entry containing the phrase "3 files" to any session log, regenerate the episode, and observe `files_changed` become 3 regardless of the actual diff.

Found by adversarial review during PR #3606.

PR #4467, opened 2026-08-03
fix(session-log): enforce declared protocol items, reject zero-commit evidence, count files from the commit range

### LINKED ISSUE #4405: validate_session_json: four MUST session-start items pass when the key is deleted, fail when left incomplete
## Problem

`validate_session_json.py` enforces a `sessionStart` MUST item only when the key is present. Four items that `new_session_log_json.py` writes as `"level": "MUST"` are absent from `SESSION_START_REQUIRED_ITEMS`, so deleting the key passes the gate while leaving it incomplete fails.

This is the deletion sibling of #3747, which fixed the demotion route (rewriting an item's own `level` from MUST to SHOULD). #3747 pinned the level of required items. It did not change which items are required, so an item outside that set can still be removed outright.

## Mechanism

`scripts/validate_session_json.py:116-125`:

```python
SESSION_START_REQUIRED_ITEMS = frozenset(
    {
        "serenaActivated",
        "serenaInstructions",
        "handoffRead",
        "sessionLogCreated",
        "branchVerified",
        "notOnMain",
    }
)
```

Six entries. `new_session_log_json.py` emits twelve `sessionStart` items, ten of them at `"level": "MUST"`. The four MUST items outside the frozenset are unenforced when absent.

## Measurement

I generated an untouched log with `new_session_log_json.py`, then deleted one `sessionStart` key at a time and re-ran the validator, checking whether the output named that key. One run per key, twelve keys:

| Key | Generator level | Deleted -> validator |
|---|---|---|
| `serenaActivated` | MUST | caught |
| `serenaInstructions` | MUST | caught |
| `handoffRead` | MUST | caught |
| `sessionLogCreated` | MUST | caught |
| `branchVerified` | MUST | caught |
| `notOnMain` | MUST | caught |
| `skillScriptsListed` | **MUST** | **silent** |
| `usageMandatoryRead` | **MUST** | **silent** |
| `constraintsRead` | **MUST** | **silent** |
| `memoriesLoaded` | **MUST** | **silent** |
| `gitStatusVerified` | SHOULD | silent (correct) |
| `startingCommitNoted` | SHOULD | silent (correct) |

The two SHOULD items behaving this way is right. The four MUST items are the defect.

The same log with those four keys present and `complete: false` fails with `Incomplete MUST: sessionStart.skillScriptsListed` and three siblings. So the gate is strictly easier to satisfy by deleting the checklist item than by filling it in. That is backwards, and it is the cheapest path in front of an agent that cannot complete the step.

## This is not theoretical

Across the 1106 committed session logs that have a `sessionStart` block:

| Key | Logs where the key is absent |
|---|---|
| `skillScriptsListed` | 244 |
| `usageMandatoryRead` | 220 |
| `memoriesLoaded` | 219 |
| `constraintsRead` | 182 |

277 logs, 25 percent of the corpus, omit at least one of the four. Every one of those passed the gate. I cannot tell from the artifact which omissions were deliberate and which were the path of least resistance, and that indistinguishability is the same objection #3747 raised about demotion: the honest case and the convenient case produce identical diffs.

I hit this by hand. The generator produced a log, the gate rejected it for five incomplete MUST items, and the two most recently merged logs on the branch I was working (`2026-08-03-session-4369-gotcha-spec-coverage-gate.json` and `2026-08-03-session-4231-episode-corpus-migration.json`) both simply lack `skillScriptsListed` and `usageMandatoryRead`. Deleting the keys was the visibly sanctioned-looking fix. I filled them in instead, but only because I stopped to measure why the two shapes differed.

## Proposed fix

Reconcile the two lists rather than patching one. Either:

**A. Promote.** Add `skillScriptsListed`, `usageMandatoryRead`, `constraintsRead`, and `memoriesLoaded` to `SESSION_START_REQUIRED_ITEMS`. This makes the validator agree with the generator. It will fail the 277 existing logs if they are ever revalidated, so it needs the same grandfathering treatment #3385 describes for historical logs.

**B. Demote.** Change the generator to emit those four at `"level": "SHOULD"`. This makes the generator agree with the validator, and admits that four items the protocol calls mandatory are in practice optional.

A is the honest reading of `SESSION-PROTOCOL.md`. B is the honest reading of what the gate actually enforces today. Picking neither leaves a MUST that is not a must, which is worse than either.

Whichever is chosen, add a consistency test that asserts every generator item at `"level": "MUST"` appears in the validator's required set for its section. Without that test this drifts again the next time an item is added to either side, which is how it drifted here.

## Repro

```bash
uv run --frozen python .claude/skills/session-init/scripts/new_session_log_json.py \
  --session-number 9011 --objective probe
uv run --frozen python scripts/validate_session_json.py \
  .agents/sessions/$(date +%F)-session-9011.json   # fails, names skillScriptsListed

# delete the key rather than completing it
uv run --frozen python -c "
import json,pathlib,sys
p=pathlib.Path(sys.argv[1]); d=json.loads(p.read_text())
del d['protocolCompliance']['sessionStart']['skillScriptsListed']
p.write_text(json.dumps(d,indent=2)+chr(10))" .agents/sessions/$(date +%F)-session-9011.json

uv run --frozen python scripts/validate_session_json.py \
  .agents/sessions/$(date +%F)-session-9011.json   # no longer mentions it
```

## Notes

Related and already closed: #3747 (demotion route), #3763 (protocol documents six fields, schema enforces two), #3346 (validator claimed schema validation it did not perform). This is the same family: the checklist and its enforcer are maintained separately and disagree.

### LINKED ISSUE #4415: startingCommit equal to endingCommit silently zeroes metrics.commits and blocks the push
## Summary

When a session log is created *after* the session's work commit, `startingCommit` and `endingCommit` hold the same SHA. `_collect_shas` correctly refuses to count the base commit, so the session's only commit disappears and the episode records `metrics.commits: 0`. The episode-store ratchet then blocks the push, and the failure message points at the episode, not at the session log that caused it.

## Evidence

`.agents/sessions/2026-08-02-session-4257-testing-discriminating-input.json` on branch `docs/testing-discriminating-input`:

```
startingCommit = 1f5b71b0a73f9e5d9f9baa94babb54db901eac51
endingCommit   = 1f5b71b0a73f9e5d9f9baa94babb54db901eac51
changesCommitted evidence = "Commit 1f5b71b0a carries the rule and both mirrors."
```

`_collect_shas(data)` returns `[]`. The exclusion is deliberate and correct (`extract_session_episode.py`, `_take`: `if starting and _same_commit(sha, starting): continue`), so the extractor is behaving as designed. The wrong input is `startingCommit`.

Setting `startingCommit` to the parent (`9c6bc856d`) and re-running `--preserve` yields `metrics.commits: 1`, and `test_the_committed_episode_store_is_clean` goes from fail to pass.

## Why it recurs

`new_session_log_json.py` stamps `startingCommit` from `git rev-parse --short HEAD` at creation time. Creating the log after committing the work is a normal order of operations, and nothing warns that the two fields now agree.

Two independent branches hit this in one afternoon (#4404, and the same shape on #4069). The ratchet ceiling is 21 and saturated, so any new episode with `commits == 0` and `files_changed > 0` now blocks a push.

## Suggested fix

Not proposing an implementation. Two candidates worth weighing:

1. Validate in `validate_session_json.py`: if `startingCommit == endingCommit` and `changesCommitted` is complete, that is a contradiction; warn or fail there, where the message can name the session log.
2. Have `new_session_log_json.py` stamp the parent when HEAD is already the session's own commit. This is harder to detect reliably at creation time.

Option 1 puts the error next to the file the author can fix. The current failure surfaces two layers away.

## Repro

```bash
git checkout docs/testing-discriminating-input   # before the fix commit
uv run --frozen python -m pytest \
  tests/skills/memory/test_extract_session_episode.py::TestValidateModeRejectsUnusableEventIds::test_the_committed_episode_store_is_clean
```

Refs #4404

rjmurillo:
## Summary

Closes the session-log metrics cluster. Three issues, one coherent surface: the
session log's own claims about what a session did, and the episode extractor
that reads them.

Fixes #4405
Fixes #4415
Refs #4416
## Changes

`scripts/validate_session_json.py`

- Session-start required items grew from 6 to 10. Added `skillScriptsListed`,
  `usageMandatoryRead`, `constraintsRead`, `memoriesLoaded`. These are declared
  by the generator and were never checked, so a log could claim protocol
  compliance while skipping four gates (issue \#4405).
- A session that declares an ending commit equal to its starting commit now
  fails validation. A zero-commit session that claims commit evidence is
  reporting work it did not do (issue \#4415).

`.claude/skills/memory/scripts/extract_session_episode.py` and
`src/copilot-cli/skills/memory/scripts/extract_session_episode.py`

- Added a middle tier to the `files_changed` metric. The primary source is the
  staged diff (from issue \#3617). When the staged count is zero, the extractor
  now computes the real count from the declared commit range with `git diff
  --name-only start..end` before falling back to counting prose mentions
  (issue \#4416).

`tests/test_session_log_metrics_integrity.py` and
`tests/skills/memory/test_episode_commit_range_files.py`

- 36 tests. The second file builds a real temporary git repository so the two
  end to end tests are actual proof the new tier is wired in, not a mock
  agreeing with itself.

## What measurement changed about the issues

Two of the three issue texts were imprecise and the fix does not match them
literally.

Issue \#4416 asks for the commit range to be used "unconditionally." That is
wrong. A staged-diff primary already exists and is more accurate, because it
sees uncommitted work. The real gap is the `staged == 0` case, which silently
fell through to prose counting. The fix inserts a tier rather than replacing
the primary.

Issue \#4415 reads as an extractor defect. It is not. The extractor already
returns False when either commit is empty and requires 7 characters of SHA.
The permissive comparison lives in the validator, whose `_same_commit` is a
mutual prefix test, so comparing any commit against an empty string returns
True. The fix is a length guard in the validator.

## Verification

Mutation harness, 7 mutations, 7 killed, cosmetic control survived, restore
byte identical.

Two of the seven mutations found real defects in my own first draft.

- An empty `startingCommit` passed the new equality check, because
  `_same_commit(x, "")` is True under a mutual prefix test and
  `isinstance(starting, str)` does not exclude the empty string. Fixed with a
  minimum length guard.
- The environment scrub test asserted `GIT_DIR` was absent without ever
  setting it, so it measured the ambient environment rather than the scrub.
  Fixed by setting the variable first.

Suites run green: 728 tests across `tests/test_validate_session_json.py`,
`tests/test_session_ending_commit_diagnostic.py`, `tests/skills/memory/`, and
the two new files. Ruff clean on all five changed files. Mypy clean on the four
inside the type-check job's scope.

All three count ratchets green. Type-ignore 44, ruff 308, taste 597.

## Notes

The stricter required-items set breaks no existing session log in the tree.
That was measured, not assumed.

## Closing claim re-audit

- Refs #4416. The zero-file range still falls through to prose, so fallback is not limited to unresolved commit ranges.


rjmurillo:
## Run Python Tests failure: shared red diagnosed

**Root cause**: `test_check_doc_interpreter_portability_identity.py::test_same_count_swap_is_detected_with_identity_baseline` was red across all PRs on CI. This is not caused by the changes in this PR.

**What happened**: PR #4397 (merged Aug 3, 7:23 AM) added a new test that calls `git commit` inside a temporary repo, but `make_repo` did not set a local git identity. Developer machines pass because they have a global gitconfig. CI runners have no global identity, so the commit exits 128 with `fatal: empty ident name`. This made `Run Python Tests` fail for every PR that branched before the fix landed.

**Evidence from run 30830624080 (the failing run for this PR)**:
- Exactly 1 test failed out of 22,555 passed + 32 skipped
- Failed step: "Run pytest" (step 14), 986s wall time
- Error: `subprocess.CalledProcessError: Command ['git', 'commit', '-m', 'swap doc'] returned non-zero exit status 128`
- stderr: `Author identity unknown / fatal: empty ident name (for <runner@...>) not allowed`

**Fix**: PR #4471 (merged Aug 3, 12:45 PM today) fixed `make_repo` to set `user.email` and `user.name` locally and added hermetic guards. All 7 tests in that file now pass on current `origin/main`.

**What you need to do**: merge `main` into this branch (`git fetch origin main && git merge origin/main --no-edit`). Once commit `89a5cc81a9` is included, the shared red will be gone.

**Hypotheses ruled out**:
- Cancellation: concurrency group is per-ref (`github.workflow`-`github.ref`), not shared across PRs.
- Timeout: job ran 17.2 min, test step 16.4 min. No timeout hit.
- Rate limit / OOM / disk: no evidence in the log.
- Changes in this PR causing the failure: the failing test file does not exist on this branch -- it was added to main after this PR branched.

Issue #4292, opened 2026-08-02
check_doc_interpreter_portability: count-only ratchet misses same-count offense replacement in a single file


rjmurillo:
## Summary

`diff_against_baseline` in `scripts/validation/check_doc_interpreter_portability.py` compares per-file **counts**, not **identities**. If a contributor removes one bare-`python3` invocation and adds a different one in the same file, the count stays equal to the baseline, no regression fires, and a new unportable invocation ships undetected.

## Reproducer

```
# doc_interpreter_baseline.json
{ "files": { "docs/contributing.md": 1 } }
```

Before the PR, `docs/contributing.md` contains:
```markdown
python3 scripts/sync_adr_protocol.py   # offense A (yaml import)
```

After the PR it contains:
```markdown
python3 scripts/detect_agent_drift.py  # offense B (different yaml-importing script)
```

`scan()` returns `{"docs/contributing.md": 1}` in both cases. `diff_against_baseline` evaluates `1 > 1` which is `False`, so no regression is emitted. Offense B ships without notice.

## Relevant code

```python
# scripts/validation/check_doc_interpreter_portability.py
def diff_against_baseline(
    current: dict[str, int], baseline: dict[str, int]
) -> tuple[list[str], list[str]]:
    regressions = [
        ...
        for rel, count in sorted(current.items())
        if count > baseline.get(rel, 0)   # only count, no identity
    ]
```

The baseline stores `{file: int}`, so there is no per-offense identity to compare. Swapping one offense for another at the same count is invisible to the guard.

## Same defect class

Issue #4152 documents the identical class in the nosec-suppression ratchet: "suppression credit ignores the rule ID, so removing a harmless nosec authorizes adding a dangerous one." This issue is the same failure mode on the doc-interpreter ratchet.

## Why the baseline is documented as empty

The script docstring says:

> `doc_interpreter_baseline.json` grandfathers per-file counts, and it is **empty**. Every in-scope file was migrated in the same change that added this guard...

An empty baseline makes the reproducer above less likely today, but does not eliminate it. Any file that acquires a baseline entry (for example through a temporary fix to a different offense, or via `--update-baseline` to grandfather a known exception) immediately becomes vulnerable. The guard logic has the defect regardless of the current baseline state.

## Impact

An unportable invocation (`python3 <script>` where the script has third-party imports) can enter any baselined file without triggering the gate, as long as a different offense in the same file is removed in the same change.

## Fix direction (not requesting code changes in this issue)

[...]

Issue #2818, opened 2026-07-02
chore(ci): AI review jobs fail as red check runs on Copilot CLI infrastructure failure (10 failed checks + WARN noise per push)


rjmurillo:
Process friction observed on PR #2817 (both pushes, runs 28562651272 and 28562747241), filed per the remediation campaign's instruction to record tool/process issues.

## Symptom

All ten AI review jobs (Security, QA, Analyst, Architect, DevOps, Roadmap, Reliability, Observability, Agent Safety, Decision Rigor) fail as red check runs on every push. Job logs show the shape:

```
##[warning]Verdict was empty, defaulting to NEEDS_REVIEW
##[warning]Both verdict and findings are empty, treating as infrastructure failure
##[warning][security] Infrastructure failure (verdict: NEEDS_REVIEW). Not blocking PR.
security review had infrastructure failure (Copilot CLI unavailable)
```

The aggregate comment then posts "Final Verdict: WARN" with ten "No findings available (empty file)" sections and an "Action Required" list naming every agent, even though zero findings exist.

## Why it matters

1. The job says "Not blocking PR" but still exits non-zero, so the PR shows 10 failed check runs. A reviewer cannot distinguish "AI reviewer found problems" from "AI reviewer never ran" without opening job logs.
2. The Action Required section tells the owner every agent "flagged issues" when every agent produced nothing. That trains reviewers to ignore the gate (the same alarm-fatigue failure mode the 2026-07-02 audit flagged for suppressed scanners).
3. PRs opened from Claude remote sessions apparently never get Copilot CLI auth in this workflow context, so this repeats on every such PR.

## Recommended direction

- On infrastructure failure, emit a NEUTRAL check conclusion (or exit 0 with a distinct warning annotation), not failure, and say "did not run: Copilot CLI unavailable" in the aggregate table instead of "flagged issues".
- Related: #2814 tracks extracting the ai-review action's inline infra-failure handling into tested Python; this issue is the observable symptom of that same code path plus the check-conclusion choice.

Refs #2814. Observed on PR #2817.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #2818: chore(ci): AI review jobs fail as red check runs on Copilot 
- **Type**: complexity_triage
- **Priority**: high
- **Suggested Agent**: agent-security
- Issue labeled with high complexity.
Suggested agents: agent-security, agent-devops, agent-implementer, agent-milestone-planner

[...]

PR #4471, opened 2026-08-03
fix(tests): make make_repo git identity hermetic so Python Tests passes on CI

rjmurillo:
## Summary

**Scope changed while this PR was open.** It opened as the fix for red `main`.
PR #4397 landed an equivalent production fix first, so the `make_repo` change
here is now redundant and has been dropped in favour of `main`'s version. What
this PR still adds is the regression guard, which `main` does not have.

The bug: `make_repo` in `tests/test_check_doc_interpreter_portability.py`
builds a temp git repo with `git init` and `git add -A` but never sets a user
identity. `test_same_count_swap_is_detected_with_identity_baseline` then runs
`git commit`. On a developer machine that commit silently picks up the global
gitconfig, so the suite is green locally. CI runners have no global identity,
so the commit exits 128 with "unable to auto-detect email address" and the job
fails. The failure is invisible to every local run, which is why it reached
`main` and stayed there from `db5aab393`.

`main` now sets the identity. Nothing stops the next contributor from removing
those two lines and shipping a green local suite again.

## Change

Two regression guards in
`tests/test_check_doc_interpreter_portability_identity.py`:

- `test_make_repo_commits_without_ambient_git_identity` commits with
  `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM` pointed at `os.devnull`, which
  reproduces the CI environment on a developer machine.
- `test_make_repo_identity_is_local_not_global` reads the value back with
  `git config --local`, which never falls back to global, so a non-empty
  result proves the identity is set on the temp repo itself.

## Verification

- **Re-proved against `main`'s implementation, not the one this branch
  originally shipped.** Deleting `main`'s two `git config` lines fails both
  guards; restoring them passes all 7 tests in the file.
- 55 tests pass across both files with a normal environment and with
  `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`.
- `scripts/validation/pre_pr.py`: all validations pass.
- `ruff check` clean.

## Notes for the reviewer

- **Merge hazard worth knowing about.** This branch and `main` independently
  created a file at the same path,
  `tests/test_check_doc_interpreter_portability_identity.py`, for unrelated
  reasons: `main` for ratchet identity (#4292), this branch for git identity.
  Git union-merged the two without reporting a conflict. The merged file was
  read in full to confirm the outcome; the 2 guards here sit alongside `main`'s
  5 ratchet tests and nothing was lost. A silent union merge is not guaranteed
  to be correct, so it was checked rather than assumed.
- The second guard originally pinned the literal `test@example.invalid`. That
  would have broken the moment `main` renamed the fixture identity, which it
  did (`tests@example.invalid`). It now asserts the structural invariant
  instead of the value.
- The guards live in a separate file rather than next to the helper because the
  helper's file was already at 495 of the 500 lines allowed by the taste-lint
  `file-size` rule. Adding `# taste-lint: ignore file-size` was rejected:
  suppressing a limit you just crossed is the same move as raising it.

Refs #4292

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


rjmurillo:
The red `Aggregate Results` on this PR is not the diff. Traced to root cause:

1. `##[warning]PR fetch failed: gh: API rate limit exceeded for user ID 6811113`
   at 18:15:49 set `CONTEXT_INFRA_FAILURE: true`.
2. All ten AI review agents skipped and wrote `DID_NOT_RUN`.
3. `gh pr checks` showed every agent as `pass`, because a clean skip is a
   successful job. Only the aggregate was red.
4. The aggregate exited 2 with `error: unknown verdict 'DID_NOT_RUN' for signal
   'qa'`, because `gate_aggregator.py:37` omits that verdict from `_KNOWN`.

Step 4 is a real defect, now filed as #4487. `DID_NOT_RUN` is first class in
`verdict.py:84`, `check_critical_failures.py:55`, and
`parse_ai_review_output.py:25`, added by #2818/#2821, so one agent's infra
failure should degrade the gate, not crash it.

Worth recording for the next person: `gh run rerun <id> --failed` does not
recover this. The agent jobs succeeded, so only the aggregate re-runs and
re-reads the same stale artifacts. Timestamps prove it, aggregation at 18:44:56
against verdicts still stamped 18:15:49, with quota recovered to 4894/5000. A
full `gh run rerun <id>` flipped 9 of 10 agents to PASS.

Issue #3275, opened 2026-07-20
nightly-cli-smoke: Copilot dogfood token is present but rejected; coverage remains zero


rjmurillo:
## Summary

`nightly-cli-smoke.yml` has failed on `main` for at least four consecutive nights (2026-07-17 through 2026-07-20, all three OS jobs red). Root cause: the Copilot auth secret the workflow references resolves to an empty string, so Copilot CLI refuses to start and the shipped-base e2e cannot run. The dogfood safety net (ADR-083 phase 4, the mechanism meant to catch shipped-base form-factor bugs before customers do) is currently exercising nothing on the Copilot side.

## Evidence (run 29724508519, 2026-07-20)

Both Copilot smoke tests fail solely on missing auth; the Claude equivalents pass:

- `tests/e2e/test_cli_hook_e2e.py::test_copilot_vendor_install_hook_resolves` FAILED, `stderr="Error: No authentication information found. ... Set the COPILOT_GITHUB_TOKEN, GH_TOKEN, or GITHUB_TOKEN environment variable"`.
- `tests/e2e/test_plugin_load_smoke.py::test_copilot_plugin_loads_expected_skills` FAILED, same auth stderr (`copilot --plugin-dir ... -p ...` returned rc=1).
- `test_claude_plugin_dir_hook_resolves` and `test_claude_plugin_loads_expected_skills` PASSED.

The job log prints `COPILOT_GITHUB_TOKEN: ` (empty) at every step.

## Why the secret is empty

`.github/workflows/nightly-cli-smoke.yml:98-99` wires:

```yaml
ANTHROPIC_API_KEY: ${{ secrets.SMOKE_ANTHROPIC_API_KEY }}
COPILOT_GITHUB_TOKEN: ${{ secrets.SMOKE_COPILOT_CLI_TOKEN }}
```

Neither `SMOKE_COPILOT_CLI_TOKEN` nor `SMOKE_ANTHROPIC_API_KEY` exists:

- Repo secrets are `ANTHROPIC_API_KEY`, `BOT_PAT`, `CLAUDE_CODE_OAUTH_TOKEN`, `COPILOT_GITHUB_TOKEN`, `FACTORY_API_KEY` (no `SMOKE_*`).
- The `copilot` environment lists no secrets, and the smoke job does not declare `environment:`, so environment secrets would not apply anyway.

[...]

PR #4523, opened 2026-08-04
fix(e2e): skip on Copilot auth failure instead of failing pre-push

### LINKED ISSUE #4483: pre-push real-CLI smokes block CI fixes without Copilot auth
Pre-push now forces hook-anchoring-e2e and plugin-load-e2e on branches that merge current main. In this non-interactive repair session, the full Python suite passed, but push was blocked because the local environment has no valid Copilot CLI auth token.

Evidence:
- PR #4110 push reached 22,489 passed tests, then failed plugin-load-e2e and hook-anchoring-e2e with: Copilot auth token was rejected (expired or revoked); rotate COPILOT_GITHUB_TOKEN for the smoke job.
- PR #4336 push reached 22,489 passed tests, then failed the same real-CLI smoke gates.
- PR #4373 push reached 22,489 passed tests, then failed the same real-CLI smoke gates.

This blocks unrelated CI repair branches after syncing with main, because the local pre-push hook requires an external auth secret that this autonomous environment does not have. Existing issue #3275 covered an empty nightly secret and is closed. This is the local pre-push variant.

rjmurillo:
Pre-push hook runs copilot/plugin-load e2e smokes whenever pushed files match the lefthook glob patterns. In environments without a valid Copilot auth token (local dev, autonomous agents), the smokes called `pytest.fail` on auth failure, blocking every push even when all 22k+ unit tests passed.

Changes:
- Add `_skip_on_auth_failure` helper in both e2e test files; callers replace inline `pytest.fail`/`pytest.skip` pairs with a single call site.
- Four call sites changed: two in `test_cli_hook_e2e.py`, two in `test_plugin_load_smoke.py`.
- Add 4 unit tests per file (8 total) covering absent token, rejected token, success no-op, and non-empty skip message.
- Add `# taste-lint: ignore file-size` to `test_plugin_load_smoke.py` (file was 658 lines before).
- Update `taste_count_baseline.txt` from 592 to 590 (branch reduces taste violations by 2).

The nightly workflow provisions `COPILOT_GITHUB_TOKEN` and uses `assert_smoke_ran.py` to catch skipped smokes, so the nightly still fails red when the secret is missing or revoked. Pre-push now succeeds without a token.

Mutation table:
- M1 `copilot_auth_failed` always False: 8 killed
- M2 skip to fail in `test_cli_hook_e2e`: 3 killed
- M3 skip to fail in `test_plugin_load_smoke`: 3 killed
- M4 cosmetic rename local var: survived

Fixes #4483
Refs #3275

## Closing claim audit

- Refs #3275. The nightly authenticated smoke and execution-proof acceptance remain outside this diff.


rjmurillo:
Superseded by PR #4555 which is based on a clean branch with the ruff baseline fixed.

PR #4533, opened 2026-08-04
fix(eval): retain bounded judge failure evidence

### LINKED ISSUE #3975: eval: judge parse failures store only a 200-char prefix, which destroys the evidence and fabricates a cause

rjmurillo:
## Summary

- Record `judge_model` on every success and failure sample.
- Preserve exact raw judge output and the live parse exception through 16,384 characters.
- Reject oversized responses before JSON parsing. Store a bounded head and tail excerpt, original character count, and SHA-256 without inventing a parse cause.
- Remove the unrelated issue #4068 CI changes from the final diff.

## Evidence

- Current main stored a 100,000-character response in full, recorded no live parse exception, and omitted per-sample judge identity.
- `tests/eval/test_eval_rule_activation.py`: 274 passed, 1 skipped.
- Negative controls failed when the evidence bound and empty-response retention were removed.
- `scripts/validation/pre_pr.py`: exit 0.
- CWE-78 scan: 2 files scanned, 0 findings.

Fixes #3975


rjmurillo:
owner: local:pr-autofix
session: refresh-behind-20260807
acquired_at: 2026-08-07T02:45:38Z
expires_at: 2026-08-07T03:00:38Z
base_sha: 0f83ad2fccc5f56382e7ac5a0b454ee1c7a08760

Issue #4489, opened 2026-08-03
new_pr conflates local validation base with GitHub base branch


rjmurillo:
## Summary

`.claude/skills/github/scripts/pr/new_pr.py --base main` uses the same value for two different namespaces:

1. local validation via `git diff main...HEAD`, and
2. GitHub PR creation via `gh pr create --base main`.

When the coordinator checkout intentionally remains on a stale local `main`, validation sees unrelated `.agents/` history and fails Session End validation even though the branch diff against `origin/main` contains only seven Python files. Passing `--base origin/main` fixes local validation but GitHub rejects it because `origin/main` is not a branch on GitHub.

## Reproduction

Observed while creating follow-up PR #4482 from an isolated worktree:

```text
local main: ede9fd1fe
origin/main: cab024221
branch: four commits on current origin/main

new_pr.py --base main
-> git diff main...HEAD includes many unrelated .agents/sessions files
-> Session End validation failed

new_pr.py --base origin/main
-> all six local validations pass
-> GraphQL: Base ref must be a branch
```

The PR was created only after separately validating the body and branch, then using GitHub's real `main` base.

## Expected

The skill should separate the local validation base from the remote PR base, for example `--validation-base origin/main` plus `--base main`, or resolve the configured remote tracking ref internally while retaining `main` for `gh pr create`.

## Impact

The validated PR creation skill is unusable from an isolated worktree when the shared coordinator checkout preserves a stale local `main`, encouraging unsafe raw fallbacks or mutation of the coordinator ref/index.

## Duplicate search

No open or closed issue matched:

- `new_pr local main stale base validation origin/main`
- `new_pr.py stale local base`

Refs #4482.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #4489: new_pr conflates local validation base with GitHub base bran
- **Type**: `complexity_triage`
- **Priority**: medium
- **Suggested Agent**: agent-implementer
- Issue labeled with medium complexity.
Suggested agents: agent-implementer

### Route issue #4489 to agent-implementer
- **Type**: `agent_routing`
- **Priority**: low
- **Suggested Agent**: agent-implementer
- Issue content suggests routing to: agent-implementer
Complexity: medium

rjmurillo:
Fixed. Same root cause and same fix as #4461. PR #4417 (commit `169120576`) introduced `resolve_comparison_base()` which separates the local validation base from the GitHub PR target.

[...]

Issue #4461, opened 2026-08-03
fix(github): new_pr.py diffs against the stale local base ref, so a worktree PR validates hundreds of untouched files and fails on someone else's session log


rjmurillo:
## Problem

`new_pr.py` computes its changed-file set with `git diff --name-only {base}...{head}` where `base` is the raw `--base` value, which defaults to `main`. In a git worktree, the local `main` ref is whatever it was when the worktree was created and is never advanced, because worktrees are branched off `origin/main` and nobody checks `main` out again. The three-dot diff therefore uses a merge base hundreds of commits old.

The consequence is not a wrong number in a report. It is a hard failure on an unrelated file. The script sees `.agents/**` in the inflated set, picks the newest session log out of it, validates that log, and exits. The author gets `Session End validation failed` for a session log their branch never touched.

## Evidence

Measured in a worktree on branch `docs/check-reader-severity`, whose real diff is three files under `.serena/memories/`.

```
local main  = 9933b7dbbba5d9d09a52e8605483a7a911e3bfbf
origin/main = db5aab39393f5cdfa17eda2ee275460b217b9da0

git diff --name-only main...docs/check-reader-severity  | wc -l   ->  797
```

Among those 797 were `.agents/SESSION-PROTOCOL.md`, `.agents/governance/GOTCHAS.md`, and roughly 60 `.agents/memory/episodes/episode-*.json` files. None are in the branch.

The failure output is opaque. Everything before the failure is buffered stdout and the failure itself is unbuffered stderr, so the terminal shows the error first and then stops at `[1/6] Checking Session End protocol...` with no indication of which file failed or why it was in scope:

```
Session End validation failed
Preparing to create PR: docs/check-reader-severity -> main
...
[1/6] Checking Session End protocol...
```

One command fixes it:

```
git fetch origin main:main
git diff --name-only main...docs/check-reader-severity | wc -l   ->  3
```

The PR then created with no other change.

## Why the current code does not catch this

`_session_log_for_validation` already has a docstring about issue #2387, where a session log in the changed set did not exist on disk in the current worktree. That fix reads the log from the branch ref instead. It makes the failure survivable but does not question why an untouched `.agents/` path was in the changed set at all. The stale base is upstream of that.

## Impact

Every PR opened from a worktree is affected, which under parallel agent operation is nearly all of them. The failure is silent about its cause, so the natural response is to go debug a session log that has nothing to do with the branch.

## Suggested fix

[...]

PR #4534, opened 2026-08-04
fix(github): resolve stale local base ref in new_pr.py worktree diffs

rjmurillo:
## Summary

In a linked worktree, the local branch ref for `main` is frozen at worktree
creation time and never advances. `new_pr.py` used `args.base` (e.g. `"main"`)
for both `git diff <base>...HEAD` validation AND `gh pr create --base`. The
stale local ref caused `git diff` to include every commit that landed on main
since the worktree was created, resulting in hundreds of unrelated files
appearing changed, failing validation on session logs written by other agents.

This is one defect with two issue descriptions. `args.base` was doing two jobs
with incompatible semantics: a git namespace (branch name for GitHub) and a
commit-ish (revision for diff). They must be independent.

## Changes

- `.claude/skills/github/scripts/pr/new_pr.py`: Added `_resolve_validation_base(pr_base, explicit)` that prefers `origin/<base>` when that remote-tracking ref exists; falls back to the bare name only when the remote ref is absent. Added `--validation-base` argument for explicit override. `gh pr create --base` still receives the bare branch name (GitHub rejects `origin/main` there).
- `src/copilot-cli/skills/github/scripts/pr/new_pr.py`: Mirror; identical to `.claude/` copy. Regenerated via `build_all.py`.
- `tests/test_new_pr.py`: 11 new tests across `TestResolveValidationBase` and `TestMainUsesResolvedValidationBase`; 4 existing tests updated for the new subprocess call sequence.

## Reproduction

Measured in this worktree before the fix:

| diff base | files in range |
|---|---|
| `main` (local, stale) | 1918 |
| `origin/main` | 0 |

## Mutation proof

| Mutation | Killing test |
|---|---|
| M1: always return pr_base | `test_returns_origin_ref_when_it_exists` |
| M2: invert returncode check | `test_returns_origin_ref_when_it_exists` |
| M3: ignore explicit override | `test_returns_explicit_when_provided` |
| M4: gh receives validation_base | `test_gh_pr_create_still_receives_bare_base` |
| COSMETIC: docstring word change | SURVIVED (control) |

Refs #4461
Refs #4489

## Closing claim audit

- Refs #4461. The issue is closed by PR #4417, and this diff removes its file-specific diagnostic.
- Refs #4489. The issue is closed as fixed by PR #4417, and this diff weakens the remote-base selection.


rjmurillo:
This PR deletes three pieces of hardening that landed in PR #4417. Two of the deletions are measurable regressions; I reproduced both rather than argue from the docstrings.

## R1. `origin/{base}` shorthand is shadowable; `refs/remotes/origin/{base}` is not

`_resolve_validation_base` returns the shorthand `origin/{pr_base}`. The function it replaces returned the fully qualified `refs/remotes/{remote}/{base}`, and its docstring said why. Shorthand resolution follows git's rev-parse search order, so `refs/heads/origin/main` is found before `refs/remotes/origin/main`.

Reproduction in a clean repo:

```
$ git branch origin/main <local-commit>
$ git rev-parse --verify --quiet 'origin/main'
acb74874842118cc8ae3603d186aca1a279450b7   # the local branch
$ git rev-parse --verify --quiet 'refs/remotes/origin/main'
5aeae8ee089e82f687cfd41f44128462073928ed   # the real remote-tracking ref
```

Different commits. When a local ref named `origin/main` exists, this PR's version diffs against the wrong base and silently reintroduces the exact stale-base bug the function exists to prevent. The `--verify` call in `_resolve_validation_base` returns rc=0 for the shadow, so the fallback never fires.

Fix: keep the fully qualified form.

## R2. Deleting `_tracking_remote()` drops single-non-origin-remote repos

`_resolve_validation_base` hardcodes `origin/`. A clone with one remote named something else loses remote-base resolution entirely and falls back to the stale local branch:

```
$ git clone -o upstream ... work2
$ git remote
upstream
$ git rev-parse --verify origin/main; echo $?
128                                     # this PR falls back to bare 'main'
$ git rev-parse --verify --quiet 'refs/remotes/upstream/main'; echo $?
0                                       # the deleted _tracking_remote() found this
$ git rev-parse --verify --quiet 'refs/heads/main'; echo $?
0                                       # ...and the stale local ref exists to be picked
```

`_tracking_remote()` handled this by name: origin wins when present, a single differently named remote is used, several remotes with no origin fall back. That coverage is gone.

## R3. The Session End failure message loses its diagnostic

The diff replaces

```
Session End validation failed for {session_log}
  (selected as the newest log in 'git diff {base}...{head}'. If that log is not
   yours, the base is behind: run 'git fetch origin' and retry.)
<validator stdout/stderr>
```

with a bare `Session End validation failed`. That drops which log was selected, how it was selected, the actionable remedy, and the validator's own output. This is the message a contributor sees at the moment the tool blocks their PR. Strictly worse.

## Note on the premise

The added docstring cites issues #4461 and #4489 as the motivation. PR #4417's `closingIssuesReferences` is `[4409]`, so #4417 was not filed as the fix for #4461/#4489. Worth confirming those two issues are still open and still describe behavior the merged code does not already handle, since the merged code resolves the same remote-tracking ref this PR resolves, only more safely.

## Recommendation

The `--validation-base` escape hatch is a genuine addition; keep it. Wire it into the existing `resolve_comparison_base` (explicit wins, otherwise the current logic) instead of replacing that function, and restore the diagnostic block in the Session End failure path.

Issue #4315, opened 2026-08-02
Advisory review-marker check prints [FAIL] and returns True, so a passing check looks like a blocking one


rjmurillo:
## Summary

`validate_review_marker` is advisory by default and returns `True` on a missing
or stale marker. Before returning, it prints the wrapped script's output
verbatim, and that output begins with `[FAIL]`. So a check that passes emits a
line that is byte-identical in shape to a blocking failure.

## Evidence

From a real pre-push run on branch `docs/retro-seam-before-suppress`:

```
[FAIL] HEAD (b4dce1639470) changes files; a review marker must be an empty commit whose Reviewed-By trailer names its parent. Re-run /review after the code tip is ready.
...
[FAIL] Instruction Budget (always-on) completed in 0.11s
...
RESULT: 1 validation(s) failed
```

Two `[FAIL]` lines, one actual failure. The first is advisory and the run would
have passed on it alone. I spent a diagnosis cycle on the review marker before
noticing the summary counted only one failure.

## Root cause

`scripts/validation/checks_coverage.py:53` prints the subprocess output
unmodified:

```python
output = (stdout or "") + (stderr or "")
if output.strip():
    for line in output.strip().splitlines()[:20]:
        print(line)
```

Then at line 62 it prints the advisory note and returns `True`.

The same function already knows the right convention. Its other advisory branch
at line 45 prints `[WARN]`:

```python
print("[WARN] validate_review_marker.py not found (advisory skip)")
```

So the inconsistency is internal to one function: one advisory path says
`[WARN]`, the other passes `[FAIL]` straight through.

## Why this is not cosmetic

The docstring says the check is advisory precisely because "most pre-PR pushes
are mid-development and have not run `/review` yet." So the confusing output is
not an edge case. It is the common path. Every such push prints a `[FAIL]` that
means nothing.

This is the same class as "a check name is not a cause": a `[FAIL]` token is
not a failure, and nothing in the line says so. The reader has to reconcile it
against the `RESULT:` count at the bottom of a long log.

## Suggested fix

In the non-enforced branch, rewrite a leading `[FAIL]` to `[WARN]` before
printing, or prefix every forwarded line with `advisory:`. Keep the passthrough
verbatim when `REVIEW_MARKER_ENFORCED=1`, since there the token is accurate.

Tests should cover all three: enforced failure keeps `[FAIL]`, advisory failure
emits no `[FAIL]`, and the missing-script advisory path keeps its existing
`[WARN]`.

[...]

Issue #4317, opened 2026-08-02
generate_rules universalizes internal-only rule scopes in the shipped plugin


rjmurillo:
## Summary

`build/scripts/generate_rules.py` converts the most narrowly scoped rules into the most
broadly applied ones in the shipped plugin. A rule whose `paths:` are all internal-only
loses every glob to the internal filter, and the generator then synthesizes universal
scope so the rule "still ships". The result is scope inversion: the narrower the source
scope, the more likely the plugin consumer loads it on every file.

Three rules are affected today. Two are on `main`.

## Measurement

Audited all 26 canonical rules with frontmatter, comparing the source scope against both
generated mirrors.

| Rule | Source `paths:` | `.github/instructions` | plugin `src/copilot-cli/instructions` |
|---|---|---|---|
| `governance` | `.agents/governance/**` | `.agents/governance/**` | **`'**'`** |
| `secret-redaction` | 6 globs, all internal | `.claude/commands/spec.md,...` | **`'**'`** |
| `session-logs` | `.agents/**` | `.agents/**` | **`'**'`** |

The first two are present on `origin/main` today:

```
$ git show origin/main:src/copilot-cli/instructions/governance.instructions.md | grep -m1 '^applyTo:'
applyTo: '**'
$ git show origin/main:src/copilot-cli/instructions/secret-redaction.instructions.md | grep -m1 '^applyTo:'
applyTo: '**'
```

Every other scoped rule keeps a real scope in the plugin, because at least one of its
globs survives the internal filter.

## Mechanism

`build/scripts/generate_rules.py:321-337`:

```python
if not keep_internal:
    value, dropped = _filter_internal_globs(value)
    ...
# If the post-filter applyTo is empty but the source had a scope,
# the source was entirely internal-only globs. Synthesize universal
# scope so the rule still ships rather than landing with applyTo: "".
applyto_value = result.get("applyTo")
if had_scope and isinstance(applyto_value, str) and not applyto_value.strip():
    result["applyTo"] = _UNIVERSAL_SCOPE
```

The comment states the intent plainly: ship it rather than emit an empty scope. Those are
treated as the only two options.

## Why the fallback is the wrong default

There is a third option, and it is the correct one for this case.

If every glob in a rule's scope was filtered as internal-only, the rule is about paths
that do not exist in the plugin. `governance` is about `.agents/governance/**`.
`session-logs` is about `.agents/**`. A plugin consumer has neither directory. Shipping
[...]

PR #4559, opened 2026-08-04
docs(governance): record three pre-push traps this branch walked into

rjmurillo:
# Pull Request

### What

Records three pre-push traps this branch walked into, corrects three claims an earlier draft of one of them got wrong, and names the ratchet that an empty `endingCommit` trips.

### Why

Each entry cost a full pre-push cycle to learn, and the suite that rejects you takes about eighteen minutes. None of the three is visible from the code you are editing, which is the definition of a gotcha rather than a rule.

## Specification References

| **Issue** | Refs #4315 | Review marker and pre-push gate behavior |

Documentation change. No spec required.

## Changes

**The CLI e2e pre-push jobs need a Copilot token.** Two lefthook jobs shell out to the Copilot CLI and fail without `COPILOT_GITHUB_TOKEN`. An earlier draft of this entry got three things wrong and a reviewer on a different model family caught all three. The jobs are siblings inside a `parallel: true` group, not sequenced behind the Python suite, so a token failure surfaces at the same time as everything else rather than after it. Each job carries its own glob list rather than a shared one, so the entry now points at the config instead of reproducing a union that drifts.

**Merging main arms every glob gated pre-push job.** The third correction turned out to be the useful one. Lefthook expands its changed file set over the push range, meaning remote tip to new HEAD, not the branch's diff against main. Merging `origin/main` puts all of main's commits inside that range. Measured on this branch: the push range covered 367 files and hit both e2e glob sets, while the branch's own diff against main touched seven files and no gated path. The rule that follows is short. If you merged main, assume every glob gated job is armed.

**A rule edit can move a number a skill reference asserts.** Editing `.claude/rules/session-logs.md` grew the shipped plugin tree by 1,032 bytes, and a test asserts that byte count against the sentence in `.claude/skills/context-optimizer/references/model-context-doctrine.md`. Nothing in pre-commit catches it. The guard lives in the Python suite, so the cost of learning this is the full run. The doctrine sentence is corrected to the measured figure here, and the vendor cost figure in the same sentence is recomputed. That second figure is not guarded by any test, which the entry says out loud.

**A Copilot auth rejection is not proof of a bad credential.** The failure text says expired or revoked and tells you to rotate, while the service message in the same output says the token may still be valid. Measured transient: both e2e jobs failed on it, and the same test passed in 24.99 seconds minutes later using a token from the same `gh auth token` call. The entry carries a 25 second reproduction so nobody rotates a working credential.

**The empty `endingCommit` ratchet.** `.claude/rules/session-logs.md` now gives the full three source order the collector walks, and frames the cost as the pre-push Python job rather than as something remote and expensive. Both generated instruction mirrors are regenerated from it.

**One file crossing 500 lines fails four pre-push jobs.** A sibling branch tripped this while this one was in review. The taste linter treats a file over 500 lines as an error and only errors count toward the whole-repo ratchet, so growing one file from 402 to 691 lines raised the count by exactly one. That single increment failed four jobs with three different messages, and the fourth was a test-suite failure that reads as unrelated to anyone scanning 23,058 passes. The entry records the direct per-file check, the path the linter actually lives at, and the staging requirement the ratchet has, since it reads the git index and an unstaged new file is invisible to it. It also names why the merge-tree job's merge-or-rebase advice is wrong for this case: that job already computed the merged result, so its verdict means merging will not help.

Files changed:

- `.agents/governance/GOTCHAS.md`
- `.agents/sessions/2026-08-04-session-9034-e2e-auth-gotcha.json`
- `.agents/memory/episodes/episode-2026-08-04-session-9034-e2e-auth-gotcha.json`
- `.claude/rules/session-logs.md`
- `.github/instructions/session-logs.instructions.md`
- `src/copilot-cli/instructions/session-logs.instructions.md`
- `.claude/skills/context-optimizer/references/model-context-doctrine.md`
- `src/copilot-cli/skills/context-optimizer/references/model-context-doctrine.md`

## Type of Change

- [x] Documentation update

## Reviewer Expectations

**Review kind / scrutiny / urgency**: targeted, standard scrutiny, no rush.

**Focus areas**: whether the merge arms globs claim generalizes, and whether the corrected doctrine figures match your own measurement.

## Testing

Deliverables in this PR:

Verification performed, each claim measured rather than reasoned:

| Claim | How it was checked | Result |
|---|---|---|
| The e2e jobs are siblings in a parallel group | Read the lefthook config around the group boundary | Confirmed, the earlier draft was wrong |
| Each job carries its own glob list | Compared the two lists | Confirmed, the draft published a partial union |
| Merging main arms the globs | Counted files in the push range against the branch diff | 367 files in range, 7 in the branch diff |
| The plugin tree grew by the rule edit | Ran the corpus figure guard before and after | 82,299 to 83,331, guard red then green |
| The auth rejection is transient | Reran the same e2e test with a fresh token from the same source | Passed in 24.99 seconds |

Both generated mirrors were regenerated rather than hand edited, and verified byte identical to their canonical sources.

### Security Review

### Other Agent Reviews

Adversarial review ran on a model family other than the one that authored the change. It filed three findings at High severity against the first draft of the gotcha entry. All three were verified independently before being accepted, and all three held. The corrections are the substance of the second commit here. Two findings at Medium severity against the rule text were applied in the same pass.

## Author Pre-flight

## Notes for Reviewers

The honest part of this change is the second commit. The first draft of the token entry published three claims that were wrong, and the review that caught them is the reason the merge arms globs mechanism was found at all. The session log records this as warranting a retrospective rather than a routine fix.

One figure ships unguarded and is labeled as such. The vendor cost number in the doctrine sentence, meaning plugin tree bytes minus the narrowly scoped mirror set, has no test asserting it. It is recomputed correctly here, and it will go stale the next time either tree moves. Adding a guard for it is worth doing and is not in this change.

## Related Issues

Refs #4315


rjmurillo:
Merged `origin/main` at `3246ddcbc`, resolving three conflicts.

**`model-context-doctrine.md`** (canonical and mirror). This branch had refreshed the byte counts describing a scope inversion: a rule scoped only to internal paths lost every glob to the internal-glob filter, the empty scope defaulted to `**`, and four `.agents/`-scoped rules shipped always-on to plugin consumers with no `.agents/` directory. Main fixed that under #4317 and rewrote the section to describe the fix plus the convergence invariant it created. Took main's text wholesale: this branch's only change there was a fresher measurement of a bug that no longer exists.

**`src/copilot-cli/instructions/session-logs.instructions.md`**, deleted on main, modified here. Accepted the deletion, and resolved by running `generate_rules.py` rather than by hand. It reports `Skipped (all-internal scope): 4` and leaves `src/copilot-cli/instructions` at 23 files against 27 in `.github/instructions`. The four skipped are `governance`, `push-lock`, `secret-redaction`, and `session-logs`, each scoped exclusively to `.agents/`, `.claude/`, or `.serena/`. The asymmetry is the fix, not drift: this repository's Copilot agent edits `.claude/` and needs those rules; an installed plugin does not have the directories they point at. This branch's edit to `.claude/rules/session-logs.md` still reaches the `.github/instructions` mirror, which is the only tree that should carry it.

`tests/validation/test_always_on_corpus_claims.py`, which main added to pin the two trees together, passes at 20 tests. All 15 pre-push gates green.

Issue #4518, opened 2026-08-04
merge-tree ratchet errors out on every branch behind its base, which is the only case it exists to check


rjmurillo:
## Summary

The `Run merge-tree ratchet` step in `.github/workflows/pr-validation.yml` fetches the base branch with `git fetch --depth=1`. That leaves the fetched commit with no parents locally, so `git merge-tree` cannot find a merge base whenever the PR branch is behind the base. It aborts with `fatal: refusing to merge unrelated histories` (rc 128), and `merge_tree_ratchet_check.py` treats any rc other than 0 or 1 as a hard failure and exits 3.

A branch that is behind main is the only condition this gate was built to evaluate. The step's own comment says so:

> A branch cut before main lowered a baseline can pass its own count check yet breach the ceiling once it lands.
> merge-tree ratchet: one stale PR merging green against old main. PR #4272 is the canonical example.

So the gate crashes in its target case and passes trivially in the case where the merged result equals the branch. It never renders its intended verdict in either direction. It reports a failure, but not its own: the ratchets it exists to evaluate never run.

Refs #3755, which this gate was built to close.

## Observed

Run [30870225792](https://github.com/rjmurillo/ai-agents/actions/runs/30870225792), PR #4508, job `Validate PR`, step `Run merge-tree ratchet`:

```
git fetch --depth=1 origin "$BASE_REF"
uv run --frozen --extra dev python3 scripts/ci/merge_tree_ratchet_check.py --base-ref FETCH_HEAD
...
From https://github.com/rjmurillo/ai-agents
 * branch              main       -> FETCH_HEAD
git merge-tree failed (rc 128):
fatal: refusing to merge unrelated histories

merge-tree-ratchet: git merge-tree did not produce a tree OID.
##[error]Process completed with exit code 3.
```

The branch was **one commit** behind main at the time:

```
main tip:   83ae7a2fb16c14649bac7f4abf44dd53ecc80966
branch HEAD: 92df2286f847694e02a81bafc17c1d235479ccd3
merge base:  d32301338a073e2dc17691f2819ff27ada1eaebd   (main tip's parent)
```

`d32301338` is exactly one step past the shallow boundary, so it is unreachable.

## Reproduction with controls

Deterministic, in a two-commit synthetic repository. No network, no API budget.

```bash
git init -b main origin && cd origin
echo c1 > f.txt && git add -A && git commit -m c1     # C1
[...]

PR #4582, opened 2026-08-04
fix(ci): un-shallow the bot-skip checkout so merge-tree works on bot PRs

rjmurillo:
# Pull Request

### What

The merge-tree ratchet job aborts `rc 128` on Renovate and Dependabot PRs even after the full-depth fetch added by PR #4567 landed on main. This adds `fetch-depth: 0` to the second, conditional checkout in that job, plus a guard that walks every checkout step rather than only the fetch line.

### Why

`pr-validation.yml` has two `actions/checkout` steps in the same job. The primary one already carried `fetch-depth: 0`. The second is guarded by `if: steps.should-run.outputs.skip == 'true'` and had no `with:` block at all, so it took the action default of `fetch-depth: 1`.

That guard fires only when the bot-skip path suppressed the first checkout, which in practice means only Renovate and Dependabot PRs. Those are exactly the PRs still failing.

A shallow clone is not repaired by fetching one ref. `.git/shallow` survives `git fetch origin <base>`, the graft points stay, and `git merge-tree` still aborts on the truncated graph. The fetch fixed *missing base ref*; it did not fix *shallow*.

## Changes

- `.github/workflows/pr-validation.yml`: add `fetch-depth: 0` to the conditional bot-skip checkout, with the measured run IDs in the comment.
- `tests/ci/test_pr_validation_workflow.py`: add `test_every_checkout_in_the_merge_tree_job_is_unshallow`, which walks every checkout step in the job, and `test_the_merge_tree_job_has_a_conditional_second_checkout` as a negative control so the guard cannot pass vacuously if the bot-skip path is ever deleted.

## Evidence

Measured failures after PR #4567 merged to main:

| PR | Result |
|---|---|
| 4552 | merge-tree ratchet `rc 128` |
| 4569 | merge-tree ratchet `rc 128` |

**The guard fails against the unfixed workflow and passes with the fix.** Reverting only the workflow change and running the new guard:

```
AssertionError: checkout step 'Checkout repository for workflow validation' in the
merge-tree job has fetch-depth=None; merge-tree needs full history on every path
into this job, including the bot-skip path.
1 failed, 53 deselected
```

Restoring the fix: `54 passed`.

Why the pre-existing test could not catch this: `test_merge_tree_ratchet_fetches_base_at_full_depth` asserts on the fetch line, which was correct both before and after. It had no way to observe the second checkout.

**Required validation, all exit code 0:**

| Command | Result |
|---|---|
| `pytest tests/ci/test_pr_validation_workflow.py` | 54 passed |
| `pytest tests/ci/test_merge_tree_ratchet_check.py` | 17 passed |
| `validate_workflows.py` | All validations passed |

## Checklist

- [x] Tests added for the change
- [x] Guard proven to fail without the fix

Refs #4518

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -235,6 +235,21 @@ jobs:
       - name: Checkout repository for workflow validation
         if: steps.should-run.outputs.skip == 'true'
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          # Issue #4518 follow-up: this checkout must match the `fetch-depth: 0`
+          # on the primary checkout above. It is the only checkout a bot PR ever
+          # gets, and the `Run merge-tree ratchet` step later in this same job
+          # needs a merge base. A default checkout writes .git/shallow, and a
+          # plain `git fetch origin "$BASE_REF"` does not remove it, so
+          # merge-tree aborts with "refusing to merge unrelated histories"
+          # (rc 128) on every Renovate and Dependabot PR.
+          #
+          # Measured 2026-08-04 on #4552 (run 30944714371) and #4569
+          # (run 30939418867), both after the full-depth fetch landed: the
+          # fetch line was already `git fetch origin "$BASE_REF"` and both
+          # still failed rc 128. Human PRs take the fetch-depth: 0 checkout
+          # above and pass, which is why the failure looked bot-specific.
+          fetch-depth: 0
 
       - name: Setup uv for workflow validation
         uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
--- a/tests/ci/test_pr_validation_workflow.py
+++ b/tests/ci/test_pr_validation_workflow.py
@@ -1039,3 +1039,70 @@ def test_merge_tree_ratchet_fetches_base_at_full_depth() -> None:
                 f"step {step.get('name')!r} fetches the base shallowly ({line!r}); "
                 "merge-tree needs a merge base. See issue #4518."
             )
+
+
+def _merge_tree_job() -> dict:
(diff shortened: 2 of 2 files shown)


rjmurillo:
Closing as superseded by #4572, which is a strict superset of this change.

I filed this without finding #4572 first. My duplicate check searched `repo:rjmurillo/ai-agents+is:issue+...`, so it could not see an open pull request. That is my error, recorded below so the next person does not repeat it.

## What #4572 covers that this does not

| Route to a `.git/shallow` graft | This PR | #4572 |
|---|---|---|
| `actions/checkout` missing `fetch-depth: 0` | 1 site, one job | 1 site, scanned repo-wide |
| `git fetch --depth=<n>` inside a `run:` body | not addressed | 7 sites |
| Guard scope | the merge-tree job only | every workflow, every merge-base consumer |
| Vacuity guard | one negative control | vacuity assertion plus positive and negative controls |
| Mechanism proof | none | real git repos proving a graft kills `merge-base` and that a later undepthed fetch does not repair it |

#4572 also proves the exact thing I asserted in prose here but did not test: a later full fetch does not repair an existing graft (`test_a_later_undepthed_fetch_does_not_repair_the_graft`).

## What was independently confirmed here, and still holds

The measurement stands and supports #4572:

- The second `actions/checkout` in the merge-tree job is guarded by `if: steps.should-run.outputs.skip == 'true'`, so it fires only when the bot-skip path suppressed the first checkout. That is why the failures concentrated on Renovate and Dependabot PRs (#4552, #4569) while human PRs passed.
- The pre-existing `test_merge_tree_ratchet_fetches_base_at_full_depth` asserts on the fetch line, which was correct both before and after, so it was structurally unable to observe the second checkout.

No unique content is being lost. Everything here is either already in #4572 or is a weaker form of it.

Branch `fix/bot-skip-checkout-shallow` will be deleted.

Issue #2815, opened 2026-07-02
chore(scripts): orphan module candidates from safety audit (12 modules, UNVERIFIED, human confirm before any removal)


rjmurillo:
From the 2026-07-02 read-only safety audit (report: `.agents/audits/2026-07-02-safety-audit.md`). Severity: **Medium**, report-only. IMPORTANT: this repo dispatches skills, agents, and hooks by string name, so "no reference found" is weak evidence. Every candidate below is **UNVERIFIED - human confirm before removal**. Nothing was deleted.

## Method

Per-module `rg -l --hidden --no-ignore` fixed-string sweep over the module stem across `*.py` (imports), `*.md` (SKILL.md, agent prompts, docs), `*.json` (settings, plugin manifests, hooks config), `*.yml/*.yaml` (workflows), `*.toml` (pyproject), plus `.githooks` and `.agents`. 269 modules scanned. A first pass without `--hidden` was discarded as invalid (it missed `.claude/`, `.github/`, `.agents/`).

## Zero references anywhere (1)

- `scripts/migrations/m7t2_rewrite_hook_bootstrap.py`: no test, no doc, no workflow, no manifest. Appears to be a fully spent one-shot migration.

## Referenced only by their own tests (11)

- `scripts/compute_health_status.py`
- `scripts/consolidate_skills.py`
- `scripts/error_classification.py`
- `scripts/invoke_batch_pr_review.py`
- `scripts/invoke_session_start_gate.py` (hook-style name, but absent from `.claude/settings*.json` and `.claude/hooks`)
- `scripts/maintenance/gc_worktrees.py`
- `scripts/new_validated_pr.py`
- `scripts/openclaw_bridge.py`
- `scripts/split_bundled_skills.py`
- `scripts/normalize_line_endings.py`
- `scripts/traceability/reconstruct_trace.py`

## Recommended action

Human triage each: (a) still invoked manually or by an external caller the sweep cannot see, keep and document the entry point; (b) spent one-shot migration, archive or remove with a normal reviewed PR; (c) planned-but-unwired, wire it or file the wiring work. Do not bulk-delete on this evidence.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #2815: chore(scripts): orphan module candidates from safety audit (
- **Type**: complexity_triage
- **Priority**: high
- **Suggested Agent**: agent-devops
- New issue opened with high complexity.
[...]

PR #4595, opened 2026-08-05
fix(scripts): repoint new_validated_pr at the surviving PR script

### LINKED ISSUE #4594: new_validated_pr.py dispatches to New-PR.ps1, deleted in #1144, so every non-web run exits 2
## Problem

`scripts/new_validated_pr.py` has been dead since PR #1144. It dispatches to `.claude/skills/github/scripts/pr/New-PR.ps1`, which that PR deleted during the PowerShell to Python migration.

Every non-`--web` invocation returns exit 2:

```
ERROR: PR creation skill not found: <repo>/.claude/skills/github/scripts/pr/New-PR.ps1
```

## Why the test suite did not catch it

This is the interesting half, and it generalizes beyond this one script.

All six tests in `tests/test_new_validated_pr.py` either mock `subprocess.run` or point `get_repo_root` at an empty `tmp_path`.

`test_exits_2_when_skill_not_found` asserts exit 2 against a directory that could never contain the script. It therefore passed, correctly, for the entire period the wrapper was broken. A healthy wrapper and a permanently dead one produce identical test output.

**Mocking the filesystem cannot detect a dispatch target that stopped existing.** Any wrapper script whose tests mock the boundary it dispatches across has the same blind spot.

## Related prior signal

Issue #2815 (the 2026-07-02 safety audit) already listed `scripts/new_validated_pr.py` as an orphan module candidate, "referenced only by its own test". The audit observed the symptom but did not identify the cause, and the module was left in place.

## Impact

Low blast radius in practice, because callers reach for `.claude/skills/github/scripts/pr/new_pr.py` directly. But the script is documented in `scripts/README.md` as a supported entry point, so anyone following the docs hits a hard exit 2.

`scripts/README.md` also carries two examples that cannot work for a second, unrelated reason: `new_pr.py:638` runs `validate_conventional_commit` unconditionally, *before* the `--skip-validation` branch at line 647. `--skip-validation` does not skip the title check, so `--title "WIP: Feature"` and `--title "hotfix" --skip-validation` are both rejected.

## Expected

1. The wrapper dispatches to the surviving Python script.
2. A test resolves the dispatch target against the real tree rather than a mock, so a deleted or repointed target fails loudly.
3. The documented examples run.

## Suggested guard

Existence alone is not sufficient. Repointing at any other real script in the same directory, say `close_pr.py`, would keep an existence-only assertion green while dispatching the wrong command. Identity and existence are independent failures and both need pinning.

The same reasoning applies one level down to the flags: if the tests mock `subprocess.run`, the emitted argument list never reaches a real parser, so a flag rename in the target is invisible until runtime.

rjmurillo:
## Summary

`scripts/new_validated_pr.py` had been dead since PR #1144. It dispatched to `.claude/skills/github/scripts/pr/New-PR.ps1`, which that PR deleted during the PowerShell to Python migration. Every non-`--web` invocation returned exit 2 with "PR creation skill not found".

This repoints it at the surviving `new_pr.py`, translates the PowerShell-style flags to their Python equivalents, and rebuilds the test suite so the same failure cannot recur silently.

Files:

- `scripts/new_validated_pr.py`
- `tests/test_new_validated_pr.py`
- `scripts/README.md`

## Why no test caught a permanently broken script

All six original tests either mocked `subprocess.run` or pointed `get_repo_root` at an empty `tmp_path`.

The critical one is `test_exits_2_when_skill_not_found`. It asserted exit 2 against a directory that could never contain the script. So it passed, correctly, for the entire time the wrapper was broken. **Mocking the filesystem cannot detect a dispatch target that stopped existing.** A healthy wrapper and a permanently dead one produced identical test output.

The 2026-07-02 safety audit had already flagged this module as "referenced only by its own test" without identifying the cause.

## Tests: 6 to 26

Three new classes, each closing a failure the old suite was structurally blind to.

| Class | Closes |
|---|---|
| `TestDispatchTargetExists` | Resolves the target against the **real tree**, not a mock. Pins the literal path *and* its existence. |
| `TestWebMode` | The `--web` branch, which had zero coverage. |
| `TestFlagContractWithRealTarget` | Runs the real target in a subprocess and reads the flags **argparse itself registered**, then checks the wrapper's emitted flags against that set. |

Two of these came directly from adversarial review, described below.

Existence alone is a weak guard. Repointing at any other real script in the same directory, say `close_pr.py`, keeps an existence-only assertion green while dispatching the wrong command. Identity and existence are separate failures, so the suite asserts both.

The flag contract is the same defect one level down. Every dispatch test mocks `subprocess.run`, so the emitted argument list is only ever compared against string literals in the test file and never reaches a real parser. A flag rename in `new_pr.py` would leave the whole suite green and fail only at runtime, which is the original bug moved from the script path to the flags.

## The contract test's first two designs were both wrong

Worth recording, because both looked correct and one shipped for two rounds of review.

**Design 1, parse the target's `--help` output.** Unsound. argparse interleaves option declarations with prose, so a flag merely *named* in another flag's description counts as accepted. This is live in the shipped help text: `--audit-reason` reads "Required when `--skip-validation` is used". Deleting the `--skip-validation` argument entirely left the help-text check still reporting it accepted. A silent pass on a real deletion.

**Design 2, anchor the regex to line starts.** This was the reviewer's proposed fix. It is worse than plainly wrong, because it is terminal-width dependent: measured False at `COLUMNS` 80, 60, 52, and 48, and **True at 44 and 40**. argparse wraps prose to the terminal, so a wrapped line can begin with a flag. A test that passes locally and fails only where `COLUMNS` is narrow or unset will not reproduce.

The reviewer's own `--body-file` example did not reproduce either. argparse had wrapped it as `--body-` + newline + `file`, so the original regex missed it by luck of the wrap column rather than by design. The mechanism they identified was real; that specific instance was not.

**Design 3, shipped.** A subprocess probe intercepts `ArgumentParser.parse_args`, reads `_actions[].option_strings`, and exits before the target does any work. No prose to misread, so the whole class of defect is gone rather than narrowed.

Three choices in it are load-bearing:

- **Intercepting `parse_args` rather than calling `build_parser()`.** This pins the parser `main()` actually runs. A `build_parser()` left as dead code after `main()` started building its own would still answer a direct call, so the contract would break at runtime with the suite green. That is the original bug again.
- **A subprocess rather than an in-process import.** `new_pr.py` puts its own directory on `sys.path` and imports `validate_pr_description` under a bare name. Two modules in this repo carry that name, so an in-process import pins one in `sys.modules` for the rest of the session. Restoring `sys.path` does not undo it. Verified: `LEAKED MODULES: []`.
- **A calling-frame check.** The probe patches `parse_args` globally, so a dependency that parses at import time would otherwise be captured instead. Proved by making one do so: without the check the probe returned `["--decoy-flag", "--help", "-h"]`.

The probe also blocks `subprocess.run`, `Popen`, `check_output`, and `check_call`. `main()` parses before it does anything else today; if that stops being true the probe dies loudly instead of running `gh pr create` from a test.

Every failure mode fails loudly. Verified by mutating the target to `parse_known_args`: 3 failed, 23 passed, with the target's stderr embedded in the assertion message.

## Verified by mutation, not by assertion count

Each defect was restored and the suite confirmed red.

| Mutation | Result |
|---|---|
| pre-fix wrapper restored | 8 of 15 failed (pre-expansion) |
| repoint at a nonexistent `.py` | 1 failed |
| repoint at `New-PR.ps1` | 3 failed |
| repoint at `close_pr.py`, a real script | 1 failed, the identity pin |
| neutralize the web-mode env guard | 3 failed, the three refusal tests |
| target switched to `parse_known_args` | 3 failed, target stderr surfaced in the message |
| delete the `--skip-validation` argument | probe failed; the help-text design had passed silently |
| wrapper emits `--reviewer`, a flag the target lacks | **2 failed, both contract tests, 24 mocked tests stayed green** |

The last row is the one that matters most. It demonstrates the mocked tests are *incapable* of seeing flag drift and the contract tests are the only thing that can. The failure names the cause and enumerates the truth:

```
wrapper emits ['--reviewer'], which .claude/skills/github/scripts/pr/new_pr.py
does not accept; it accepts ['--audit-reason', '--base', '--body', '--body-file',
'--draft', '--head', '--help', '--skip-validation', '--title', '-h']
```

## Complexity

`main` measured 14 against the taste-lint ceiling of 10. That is pre-existing, measured at 14 both before and after the dispatch fix, but it is the same function this change rewrites, so it is repaired here rather than left behind.

Split along seams it already had: `_build_parser`, `_run_web_mode` (3), `_build_skill_args` (7), `main` (6).

The web-mode tests exist partly because that extraction would otherwise have had no coverage to preserve.

## README correction

`scripts/README.md` carried two unrunnable examples, `--title "WIP: Feature"` and `--title "hotfix" --skip-validation`. Both are rejected: `new_pr.py:639` runs `validate_conventional_commit` unconditionally, *before* the `--skip-validation` branch at line 647. `--skip-validation` does not skip the title check. Examples corrected and the ordering documented.

## Review

Adversarially reviewed across five rounds against `gemini-3.1-pro-preview`, a different model family than the author.

| Round | Verdict | Substance |
|---|---|---|
| 1 | BLOCKED | Importing `SKILL_RELPATH` into its own test buys existence, not identity. `_run_web_mode` had zero coverage, so "behavior preserving, evidenced by passing tests" rested on nothing for that path. |
| 2 to 3 | Iterating | Flag-contract test added, then found to leak `sys.modules` entries into the session. |
| 4 | BLOCKED | The help-text regex counts a flag named in prose as accepted. Correct diagnosis; the proposed fix was defective, see above. |
| 5 | SHIP | "Exceptionally robust." Every failure mode confirmed loud. |

It also caught two factual errors in the original commit messages (the old suite had six tests, not five). Both were verified against `git show` before correcting, and the history was squashed to remove the inaccurate messages.

Both fixes the reviewer proposed contained real defects, so each was tested rather than applied. The round 4 verdict was correct and the round 4 remedy was not; that gap is why design 3 exists.

## Validation

- 26 passed
- ruff clean, mypy clean
- every function under the complexity ceiling
- live dispatch confirmed end to end: the wrapper reaches `new_pr.py`, and `new_pr.py`'s own conventional-commit check fires on a bad title

Fixes #4594
Refs #2815

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -104,15 +104,18 @@ Creates a PR with all guardrails enforced.
 
 **Usage**:
 
+Titles must follow conventional commit format. That check runs before
+`--skip-validation` is considered, so the flag does not bypass it.
+
 ```bash
 # Normal PR (runs validations)
 uv run --frozen python -m scripts.new_validated_pr --title "feat: Add feature" --body "Description"
 
 # Draft PR
-uv run --frozen python -m scripts.new_validated_pr --title "WIP: Feature" --draft
+uv run --frozen python -m scripts.new_validated_pr --title "feat: Add feature" --draft
 
 # Force mode (creates audit trail)
-uv run --frozen python -m scripts.new_validated_pr --title "hotfix" --skip-validation --audit-reason "hotfix"
+uv run --frozen python -m scripts.new_validated_pr --title "fix: Restore login" --skip-validation --audit-reason "hotfix"
 
 # Interactive mode
 uv run --frozen python -m scripts.new_validated_pr --web
--- a/scripts/new_validated_pr.py
+++ b/scripts/new_validated_pr.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
 """Create a validated PR with all guardrails enforced.
 
-Wrapper around the New-PR skill that provides a convenient interface for
-creating PRs with validation. Delegates to the skill for better cohesion.
+Wrapper around the `new_pr` skill script that provides a convenient interface
+for creating PRs with validation. Delegates to the skill for better cohesion.
 
 EXIT CODES:
   0  - Success
@@ -19,11 +19,19 @@ import os
 import shutil
 import subprocess
 import sys
+from pathlib import Path
 
 from scripts.github_core.repo import get_repo_root
 
+SKILL_RELPATH = Path(".claude/skills/github/scripts/pr/new_pr.py")
+"""Dispatch target, relative to the repo root.
 
-def main(argv: list[str] | None = None) -> int:
+Exported so tests can assert the wrapper points at a script that exists rather
+than restating the path and drifting from it.
+"""
+
+
+def _build_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(description="Create a validated PR with guardrails")
     parser.add_argument("--title", default="", help="PR title in conventional commit format")
     parser.add_argument("--body", default="", help="PR description body")
@@ -36,7 +44,49 @@ def main(argv: list[str] | None = None) -> int:
     )
     parser.add_argument("--skip-validation", action="store_true", help="Skip validation checks")
     parser.add_argument("--audit-reason", default="", help="Reason for skipping validation")
-    args = parser.parse_args(argv)
+    return parser
+
+
+def _run_web_mode(base: str) -> int:
+    """Hand off to `gh pr create --web`, which needs a browser and so refuses in CI."""
+    is_ci = os.environ.get("CI") or os.environ.get("GITHUB_ACTIONS")
+    if is_ci or not os.environ.get("DISPLAY"):
+        print("ERROR: Web mode not available in CI or headless environments", file=sys.stderr)
+        return 2
+
+    gh_args = ["gh", "pr", "create", "--web"]
+    if base:
+        gh_args.extend(["--base", base])
+
+    sys.stdout.flush()
+    return subprocess.run(gh_args).returncode
+
+
+def _build_skill_args(skill_script: Path, args: argparse.Namespace) -> list[str]:
+    """Translate this wrapper's flags into the target script's command line."""
+    skill_args = [
+        sys.executable, str(skill_script),
+        "--title", args.title, "--base", args.base,
+    ]
+
+    if args.head:
+        skill_args.extend(["--head", args.head])
+    if args.body:
+        skill_args.extend(["--body", args.body])
+    if args.body_file:
+        skill_args.extend(["--body-file", args.body_file])
+    if args.draft:
+        skill_args.append("--draft")
+    if args.skip_validation:
+        skill_args.append("--skip-validation")
+        if args.audit_reason:
+            skill_args.extend(["--audit-reason", args.audit_reason])
+
+    return skill_args
+
+
+def main(argv: list[str] | None = None) -> int:
+    args = _build_parser().parse_args(argv)
 
     repo_root = get_repo_root()
     if not repo_root:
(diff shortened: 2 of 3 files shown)


rjmurillo:
## This PR was created by the script it fixes

Strongest available evidence that the dispatch works, so recording it here.

`scripts/new_validated_pr.py` on this branch was used to open this PR:

```bash
uv run --frozen python scripts/new_validated_pr.py \
  --title "fix(scripts): repoint new_validated_pr at the surviving PR script" \
  --body-file <body> \
  --head fix/new-validated-pr-dead-dispatch
```

It reached `new_pr.py`, ran all six pre-creation validations, and created PR #4595. On `main` the same command exits 2 without reaching the first validation.

An earlier attempt on the same branch got through every validation and then failed at `gh pr create` with `GraphQL: API rate limit already exceeded`. That run is useful too: it proves the wrapper hands off correctly and that the only remaining failure was external.

## A note on what the tests can and cannot prove

`TestFlagContractWithRealTarget` parses the emitted arguments with `new_pr.py`'s real parser, so a flag rename is caught. It does not prove the target *behaves* correctly on those flags, because the parser is exercised, not the command.

That gap is intentional. Executing `new_pr.py` for real in a unit test would create GitHub PRs. The dogfood run above is what covers it, and it covers it better than a test could.

rjmurillo:
Independent live reproduction, on a branch that does not carry this fix.

While opening #4599 from `docs/push-status-misread`, I invoked the wrapper the normal way:

```
uv run --frozen python -m scripts.new_validated_pr \
  --title "docs: record that a push in flight owns its working tree" \
  --base main --head docs/push-status-misread --body-file <path>
```

Result:

```
ERROR: PR creation skill not found: .../.claude/skills/github/scripts/pr/New-PR.ps1
EXIT=2
```

Unplanned, and it confirms three things the mutation testing could only argue:

1. The dispatch is still dead on current `main`, not only in a reconstructed pre-fix state.
2. The failure is total. The wrapper has no usable non-`--web` path at all.
3. The target it names is the PowerShell script deleted in #1144, exactly as described above.

Re-running the same arguments against `.claude/skills/github/scripts/pr/new_pr.py` directly succeeded and created #4599, so the target this PR repoints at is the correct one.

PR #4617, opened 2026-08-05
fix(session): preserve episode chronology on preserve runs

rjmurillo:
## Summary

Fixes #4588

- Preserves explicit event timestamps during `--preserve --pending-stage`.
- Keeps deterministic fallback only for missing timestamps.
- Rejects causal edges that point backward in measured time.

Note: this branch was built on `fix/3972-4441-memory-metrics` as requested. That remote branch was deleted before PR creation, so this PR now targets `main` and carries those prerequisite commits too.

## Validation

- 380 memory extractor tests passed after merging `origin/main`.
- `ruff check` passed on changed memory files before the final merge.
- `scripts/validation/pre_pr.py` passed after merging `origin/main`.
- `scripts/ci/check_plugin_lib_mirrors.py` passed before the final merge.
- Push verified: `6ba0f1572ca638bf7f4f798d6bafa4308582b891`.

## Mutation proof

| Mutant | Result | Named failing test |
| --- | --- | --- |
| explicit event timestamps normalized to midnight | DEAD | `tests/skills/memory/test_extract_session_episode.py::TestMergePreserving::test_event_timestamps_keep_explicit_source_time` |
| backward edge validator disabled | DEAD | `tests/skills/memory/test_extract_session_episode.py::TestSequentialEventLinks::test_backward_milestone_to_commit_edge_is_invalid` |
| preserved timestamp docstring reworded | SURVIVED | cosmetic control |

Issue #4605, opened 2026-08-05
Fix append-target merge conflicts for governance files


rjmurillo:
## Problem

A landing sweep found 16 open PRs in CONFLICTED state. The conflicts cluster on append-target files where independent additions should both survive.

Evidence:

- `.serena/memories/memory-index.md`: PRs 4592, 4591, 4531, 4551, 4479, 4521
- `.agents/governance/GOTCHAS.md`: PRs 4574, 4565, 4563, 4543, 4521

## Current behavior

Both files receive append-only edits from contributors and agents. Two branches that each append a row or section collide during merge. Each conflict costs a rebase, and manual resolution can drop one side's entry.

## Constraint

`.gitattributes` correctly documents that custom `merge=<name>` drivers do not work without per-clone `git config merge.<name>.driver`. It overgeneralizes from that point. Git has built-in low-level merge drivers: `text`, `binary`, and `union`. `union` needs no per-clone registration.

## Scope

- `GOTCHAS.md`: safe candidate for built-in `merge=union` because sections are append-only prose under headings.
- `memory-index.md`: riskier because duplicate rows for the same memory path would corrupt the index if both sides modify the same row. A fix must make duplicate rows loud or auto-repair them before main.

## Acceptance criteria

- A real two-branch merge test proves independent appends survive under the chosen attributes.
- The memory index duplicate-row case is either rejected or healed, not silently accepted.
- The `.gitattributes` comment preserves the warning about custom drivers but explains built-in `union` correctly.
- The fix includes positive, negative, and edge tests.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #4605: Fix append-target merge conflicts for governance files
- **Type**: `complexity_triage`
- **Priority**: medium
- **Suggested Agent**: agent-qa
- Issue labeled with low complexity.
Suggested agents: agent-qa

### Route issue #4605 to agent-qa
- **Type**: `agent_routing`
- **Priority**: low
- **Suggested Agent**: agent-qa
- Issue content suggests routing to: agent-qa
Complexity: low

rjmurillo:
Implemented by PR #4618. `memory-index.md` now uses the union merge driver, duplicate rows are healed by the token updater, and real two-branch merge tests pass. `GOTCHAS.md` intentionally keeps loud conflicts by design.

Issue #4588, opened 2026-08-04
bug(session): preserve rewrites episode chronology


rjmurillo:
## Problem

The pre-commit episode extractor rewrites explicit milestone timestamps and causal edges when it runs with `--preserve --pending-stage`. The resulting episode passes `--validate` despite placing post-commit milestones before the commits they describe.

## Reproduction

1. Set timestamped `workLog` entries in a session log after their referenced commits.
2. Run `extract_session_episode.py <session> --output-path .agents/memory/episodes --force`. The generated episode uses the explicit timestamps and orders the causal edges correctly.
3. Stage the session and episode, then commit normally.
4. The `extract-session-episodes` pre-commit hook runs `git_hook_policy.py extract-episodes`, which invokes the extractor with `--preserve --pending-stage`.
5. Inspect the staged episode. The timestamped milestones have changed to midnight, their `caused_by` edges are empty, and they now lead to an earlier commit.

Observed on `.agents/sessions/2026-08-04-session-4435-reconcile-4435-4101-merged.json`:

- `Finalized closeout chronology` changed from `2026-08-04T10:26:49+00:00` after commit `02d2f0da2` to midnight before it.
- `Recorded final merged outcome` changed from `2026-08-04T19:35:33+00:00` after commit `0fd1ea30f` to midnight before it.
- `extract_session_episode.py <episode> --validate` still reported zero violations.

## Expected

`--preserve` retains explicit source timestamps and valid causal edges, or rebuilds them without moving later milestones before earlier commits. Validation rejects any causal edge whose timestamp runs backwards, including milestone-to-commit edges.

## Impact

Normal commits can silently corrupt session chronology after a clean manual generation. Reviewers then see false event order even though every local validator passes.

Refs #4435.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #4588: bug(session): preserve rewrites episode chronology
- **Type**: `complexity_triage`
- **Priority**: medium
- **Suggested Agent**: agent-qa
- Issue labeled with low complexity.
Suggested agents: agent-qa, agent-milestone-planner

### Route issue #4588 to agent-qa, agent-milestone-planner
- **Type**: `agent_routing`
- **Priority**: low
- **Suggested Agent**: agent-qa
- Issue content suggests routing to: agent-qa, agent-milestone-planner
Complexity: low

Issue #4561, opened 2026-08-04
Session numbers are allocated against the local worktree, so parallel branches collide and clash at merge


rjmurillo:
## Summary

Session number allocation detects collisions only against the local working tree, so two agents on two branches reliably pick the same number. The clash is invisible until the branches merge, where it lands as an `add/add` conflict on two files at once.

Hit today on PR #4477 merging current `main`:

```text
CONFLICT (add/add): Merge conflict in .agents/memory/episodes/episode-2026-08-04-session-9998-pr-landing.json
CONFLICT (add/add): Merge conflict in .agents/sessions/2026-08-04-session-9998-pr-landing.json
```

Both sides are real, distinct sessions. Neither is a stale copy:

```text
ours   branch: docs/pr-gate-knowledge      endingCommit: 0ae5751c6
theirs branch: docs/check-reader-severity  endingCommit: cad77bc9b
```

## Root cause

`.claude/skills/session-init/scripts/new_session_log.py:234-256` retries on collision:

```python
try:
    with open(filepath, "x", encoding="utf-8") as f:
        ...
except FileExistsError:
    session_number += 1
```

`open(..., "x")` is atomic, and that is the right primitive for two processes sharing one filesystem. It is structurally blind to the case this repository actually runs in. Each agent works in its own worktree on its own branch. Neither tree contains the other's session file, so `FileExistsError` never fires, both writes succeed, and both records claim the same number.

The detector cannot observe the namespace it is supposed to be protecting. The session number namespace is shared across every branch that will merge into `main`; the check is scoped to one branch's checkout.

## Why this is not covered by #2379

#2379 is the same failure class, closed `completed` on 2026-06-05. Its own body scopes the fix to the single instance and says the mechanism was left in place:

> The specific collision is fixed in #2359, but the allocation risk remains for parallel agents.

The resolution recorded there is manual renumbering, which is what I did again today. Two months and one collision later, the allocation risk is still live, and fleet mode makes parallel branches the normal case rather than the exception.

## Impact

Every collision costs a hand resolution across two files, and the conflict shape invites the wrong fix. `add/add` on a JSON pair tempts a "take theirs" resolution, which silently discards an entire session's audit record. The episode file carries a `_source_session` back-reference, so a careless resolution can also leave the surviving episode pointing at a session log that no longer exists.

## Options

[...]

PR #4647, opened 2026-08-05
chore(landing): roll up 16 pull requests into one CI cycle

### LINKED ISSUE #4497: mutation_harness_ciperms.py: a failed restore exits 1 with a traceback instead of the documented exit 2, and can leave the target file empty
## Problem

`scripts/ci/mutation_harness_ciperms.py` documents exit 2 for "a mutated file could not be restored". The code cannot reach exit 2 on the realistic failure. When the restore write raises an `OSError`, the harness exits **1** with an uncaught traceback, prints neither the documented diagnostic nor the recovery hint, and leaves the mutant on disk. One shape leaves the target file empty.

Verified against `scripts/ci/mutation_harness_ciperms.py` at blob `fda848d80c2ce0420f335e259c6f1ba807881b44`, which is the content on `main` today. The file was added by merged PR #4222 (merge commit `0d1dbd190`, 334 lines, `ADDED`). One commit has touched it since, `adb2e0a42` (#4447), which added `errors="replace"` to a `subprocess.run` call and left the restore flow byte-identical.

### The documented contract (verbatim, lines 11-15)

```text
Exit codes:
  0 - every mutant DEAD
  1 - any SURVIVED, DID-NOT-APPLY, or NOT-RUN
  2 - a mutated file could not be restored (the tree is left dirty; recover
      with ``git checkout -- <file>`` before rerunning)
```

### The code (verbatim, lines 114-129, in `apply_mutation`)

```python
    target.write_bytes(mutated)
    note = ""
    try:
        proc = _run_tests(mutation.test_filter)
        outcome, note = _classify(proc)
    finally:
        # Always restore.
        target.write_bytes(backup)

    # Verify restore.
    restored = target.read_bytes()
    if restored != backup:
        print(f"ERROR: restore of {target.name} failed!", file=sys.stderr)
        sys.exit(2)
```

The restore write (line 121) and the read-back (line 124) are unguarded. An `OSError` there propagates out of `apply_mutation`, out of `main()`, and out of `sys.exit(main())` at line 335. Nothing catches it, so CPython prints a traceback and the process exits 1.

Lines 126-127, the diagnostic and `sys.exit(2)`, are reachable only when the restore write **returns successfully and leaves different bytes on disk**. `Path.write_bytes` does not produce that outcome on its own; it needs external interference. So the documented exit-2 path is effectively dead code, and the failure it was written for takes the exit-1 path instead.

## Reproduction

Isolated copy of the harness, one OS process per scenario, no tracked file touched. `_run_tests` is stubbed to a DEAD verdict so only the restore path varies.

```bash
WORK="$(mktemp -d)"
git show main:scripts/ci/mutation_harness_ciperms.py > "$WORK/harness.py"
cat > "$WORK/driver.py" <<'PY'
import os, pathlib, subprocess, sys
sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
import harness

scenario = sys.argv[1]
work = pathlib.Path(sys.argv[2]); work.mkdir(parents=True, exist_ok=True)
target = work / "sample.py"; target.write_bytes(b"ORIGINAL\n")
harness._run_tests = lambda f: subprocess.CompletedProcess(["pytest"], 1, "1 failed", "")

real = pathlib.Path.write_bytes
if scenario == "readonly":            # real filesystem permission, no write_bytes mocking of the failure
    def w(self, data):
        n = real(self, data)
        if data == b"MUTANT\n":
            os.chmod(self, 0o444)     # the restore write will raise PermissionError
        return n
    pathlib.Path.write_bytes = w
elif scenario == "truncate_then_fail":
    state = {"n": 0}
    def w(self, data):
        state["n"] += 1
        if state["n"] == 2:                # the restore write
            open(self, "wb").close()       # truncate, then the device fills
            raise OSError(28, "No space left on device")
        return real(self, data)
    pathlib.Path.write_bytes = w
elif scenario == "silent_wrong_bytes":
    def w(self, data):
        return real(self, b"SOMETHING ELSE\n") if data == b"ORIGINAL\n" else real(self, data)
    pathlib.Path.write_bytes = w

print("RESULT", harness.apply_mutation(
    harness.Mutation(scenario, target, b"ORIGINAL", b"MUTANT", "t")).outcome)
PY

for s in control readonly truncate_then_fail silent_wrong_bytes; do
  python3 "$WORK/driver.py" "$s" "$WORK/w_$s" >"$WORK/out_$s" 2>"$WORK/err_$s"
  echo "$s exit=$? file=[$(cat "$WORK/w_$s/sample.py" | tr -d '\n')]"
done
```

### Observed (Python 3.14.6, Linux, non-root)

| scenario | exit | `ERROR: restore of ...` printed | file after |
| --- | --- | --- | --- |
| control, restore succeeds | 0 | n/a | `ORIGINAL` |
| restore blocked by real FS permission (`PermissionError`) | **1** | no | `MUTANT` |
| restore raises `OSError(ENOSPC)` after truncate | **1** | no | **empty, 0 bytes** |
| restore write succeeds with wrong bytes | 2 | yes | `SOMETHING ELSE` |

stderr tail for the two exit-1 cases is a bare traceback ending in `PermissionError: [Errno 13] Permission denied: .../sample.py` and `OSError: [Errno 28] No space left on device`. The string `ERROR: restore of` appears zero times in both.

Multi-mutation run, 3 mutations with the restore of M2 blocked: exit 1 after only M1's verdict is printed, M2's file left holding the mutant, M3 never runs, no summary block, no recovery hint.

## Expected vs actual

| | documented | actual on `OSError` |
| --- | --- | --- |
| exit code | 2 | 1 |
| stderr | `ERROR: restore of <file> failed!` plus the `git checkout -- <file>` hint | Python traceback |
| dirty tree signalled | yes | no |
| original content | preserved, mutant left for `git checkout` | may be destroyed, 0 bytes |

## Why it matters

- Exit 1 is already this harness's "a mutant SURVIVED / DID-NOT-APPLY / NOT-RUN" code. A caller cannot distinguish "your test is weak" from "your working tree now holds a live mutant".
- The recovery hint the docstring promises never prints on the one path where a human needs it.
- **Data-loss risk**: the truncate-then-fail shape leaves the target at 0 bytes. For a tracked file `git checkout -- <file>` recovers it, but the harness emits no signal that recovery is needed, so the empty file can be committed or can silently poison the next run.

## Existing coverage

`tests/ci/test_mutation_harness_ciperms.py` contains 23 test functions. The closest, `TestApplyMutation::test_file_is_restored_when_the_run_raises`, makes `_run_tests` raise and asserts the restore **succeeded**. No test makes the restore itself fail, so lines 126-127 are never executed by the suite. The documented exit-2 contract is untested and, under `OSError`, unreachable.

## Minimal fix scope

One file, `scripts/ci/mutation_harness_ciperms.py`, function `apply_mutation` only:

1. Wrap the `finally:` restore write in `try/except OSError`. On failure, print the documented diagnostic including the underlying OS error and the `git checkout -- <file>` recovery hint, then `sys.exit(2)`.
2. Guard the read-back on line 124 the same way. An `OSError` while verifying is also a failed restore, not an exit 1.
3. Do not swallow an in-flight exception from the `try` body. A restore failure during exception unwinding must still surface as exit 2 with the original error visible, not discarded.

No change to the success path, the docstring's exit table, or any other file. The code moves to the contract, not the contract to the code.

## Test scope

New cases in `TestApplyMutation` in `tests/ci/test_mutation_harness_ciperms.py`:

- restore write raises `PermissionError` -> `SystemExit(2)`, stderr contains the diagnostic and the recovery hint
- restore write raises `OSError(ENOSPC)` after truncating -> `SystemExit(2)`
- read-back raises `OSError` -> `SystemExit(2)`
- restore fails while `_run_tests` is already raising -> `SystemExit(2)` and the original error is not silently lost
- negative control: successful restore still returns a `Result` and does not exit
- multi-mutation run: a restore failure exits 2, not 1

## Acceptance criteria

- [ ] every restore failure reachable from `apply_mutation` exits 2, not 1
- [ ] no traceback escapes `apply_mutation` for a restore `OSError`
- [ ] stderr on that path contains `ERROR: restore of <file> failed!`, the underlying OS error, and the `git checkout -- <file>` hint
- [ ] lines 126-127 are executed by the test suite
- [ ] all six test cases above exist and pass; the existing 23 tests still pass unchanged
- [ ] the docstring exit table at lines 11-15 is unchanged

## Related, not duplicate

- **PR #4222** (merged) added this file. The defect ships with it and survives on `main`.
- **#4346** (open) covers the symptom class "mutation harness leaves live mutants in the working tree", evidenced on `scripts/validation/git_hook_policy.py` via the debate-log mutation test, and prescribes scratch-copy mutation, signal handlers, a marker file, and a fail-closed pre-push check. It does not cover this harness, the exit-code contract mismatch, the missing diagnostic, or the content-destruction shape.
- **PR #4110** (open) enforces exit contracts elsewhere; none of its changed files touch `scripts/ci/mutation_harness_ciperms.py`.
- **#4493 / #4494** (open) concern `tests/mutation/test_mutate_baseline_ratchet_integrity.py`, a different harness.

## Duplicate search

GraphQL `search(type:ISSUE)`, which covers open and closed issues and PRs:

| query | hits | relevant |
| --- | --- | --- |
| `mutation_harness_ciperms` | 0 | none |
| `mutation harness restore exit code in:title` | 0 | none |
| `ciperms restore in:body` | 1 | PR #4222 itself |
| `"restore of" "exit 2" in:body` | 4 | all unrelated merged PRs |
| `is:issue is:open mutation harness exit contract` | 4 | none on this harness |
| `"content-destruction" OR "truncated" harness in:title` | 48 | #4346, #4493, #4494 only, all different scope |

No existing issue covers this harness's restore-failure exit contract.

### LINKED ISSUE #4524: fix(validation): unify portability Git process isolation
## Problem

`scripts/validation/portability_common.py::_git_lines` has weaker Git process
isolation than the other portability helpers.

- `_git_lines` strips only a named subset of `GIT_*` variables.
- It does not pass `--no-replace-objects`.
- `scripts/validation/portability_git.py::run_git` and
  `check_doc_interpreter_portability.py::_run_git` strip every `GIT_*`
  variable and disable replacement objects.

`_git_lines` feeds `tracked_coverage_by_root`, which gates baseline updates for
sibling portability ratchets. Local Git overrides can therefore change the
coverage evidence on those paths.

## Expected

All portability Git reads use one hardened subprocess primitive. It must strip
every `GIT_*` variable, pass `--no-replace-objects`, and preserve each caller's
bytes or text output contract.

## Evidence

Found during the architect axis for the documented-interpreter portability
follow-up at reviewed tip `6a6c51874`.

### LINKED ISSUE #4526: Suppressed Copilot review findings never become threads, so no completion gate can see them (28 merged unaddressed on PR #4277)
## Summary

Copilot posts review submissions whose body carries a `Suppressed comments (N)` section. Those findings never become review threads and never become inline review comments. Every completion criterion in this repository that measures reviewer feedback counts review threads or comment-map statuses, so a review carrying N suppressed defects is indistinguishable from a clean review.

On PR #4277, 28 suppressed findings were merged unaddressed. Two of the defects they named still reproduce on `main` and required successor work in PR #4484.

## Measured reproduction

The six Copilot reviews on PR #4277, read from `reviews.nodes.bodyText` via GraphQL:

| Review submitted (UTC) | Inline comments | Suppressed findings |
|---|---|---|
| 2026-08-02T11:43:35Z | 4 | 6 |
| 2026-08-03T04:53:56Z | 0 | 3 |
| 2026-08-03T07:59:00Z | 1 | 1 |
| 2026-08-03T13:39:24Z | 0 | 8 |
| 2026-08-03T15:54:30Z | 0 | 7 |
| 2026-08-03T16:44:52Z | 0 | 3 |
| **Total** | **5** | **28** |

Four of the six reviews report `generated no new comments` in the same body that carries 3, 8, 7, and 3 suppressed findings respectively.

PR #4277 merged with 5 review threads, all resolved, and 131 check contexts with zero failures. No gate ever saw the 28.

The suppressed findings are not noise. They repeat three defects verbatim across carriers:

1. The vocabulary grep matches `Status:` while the comment map emits `**Status**:`, so the pending count matches nothing.
2. The claim that any unrecognized status counts as pending (fail closed) is false for a grep that enumerates a fixed status list.
3. The PR description claims a Phase 8.1 severity change the branch does not contain.

Defects 1 and 2 reproduce on `main` today. PR #4484 exists to restore them. Issue #4465 tracks the same class in `.claude/skills/github/references/patterns.md`.

## Expected behavior

A gate that claims all reviewer feedback is addressed accounts for findings the reviewer suppressed rather than threaded, or states explicitly that suppressed findings are out of scope and reports their count.

## Actual behavior

Nothing in the repository reads them:

- No script under `.claude/skills/github/scripts/pr/` selects `bodyText`, so review submission bodies are never retrieved.
- `get_pr_reviewers.py` keeps author identity only and discards review body and state.
- No source file parses the string `Suppressed comments`.
- `run_completion_gate.py` dispatches criteria whose verdicts come from thread-counting commands, so suppressed findings cannot reach any `pass_when` expression.

Issue #4378 specified a `get_pr_reviews.py` returning review submission bodies. That issue is closed, and no such script exists on `main`, so the capability to read this data is still absent.

## Impact

Suppressed findings bypass every feedback gate at once. The reviewer with the strongest observed signal on this repository files a defect, the gate reports zero unaddressed feedback, the PR merges, and the defect reaches `main`. PR #4277 is one measured instance where that path cost a successor PR. The failure is silent by construction: the more findings a reviewer suppresses, the cleaner the PR looks.

## Acceptance criteria

1. A skill script returns review submissions with author, state, submitted time, and body, with pagination completeness.
2. A parser extracts suppressed findings from a review body, yielding file, line, and finding text, and returns a count of zero when the section is absent.
3. The completion criteria surface a non-zero suppressed-finding count as an explicit verdict rather than silence. Either the count blocks, or it is reported and requires a recorded disposition.
4. Tests cover positive (body with a suppressed section), negative (body with no such section), and edge cases (zero-count section, malformed body, multiple reviews on one PR, suppressed section alongside inline comments).
5. Re-running the check against PR #4277 reports 28 suppressed findings rather than zero.

## Deduplication

Searched open and closed issues and pull requests for: `suppressed Copilot review`, `suppressed findings`, `Copilot low confidence comments`, `review findings never become threads`, `unaddressed review findings`, `completion gate counts threads`, `hidden review comments`, `review comments bypass gate`, `Suppressed comments section review body`, `parse review body findings`.

Closest prior art, none of which covers this defect:

- #4378 (closed) adds review-body API surface but does not gate on suppressed findings, and its script is absent from `main`.
- #1493 (closed) adds verification that bot claims are correct, which presumes the claim was visible.
- #369 (closed) adds CI check verification, a different signal.

Refs #4277, #4484, #4445, #4465, #4054

### LINKED ISSUE #4544: detect_scope_explosion counts merge-brought files, so every git merge origin/main resolution is blocked as scope explosion
## Summary

`scripts/detect_scope_explosion.py` counts the files a `git merge origin/main` brings in as if they were the branch's authored scope. Any branch whose base has moved by more than 50 files cannot commit a merge resolution without an override, even when the branch itself touches a handful of files.

The script already prefers `origin/<base>` over the stale local ref (issue #2207), so this is not the stale-base-ref defect. The count is wrong while `MERGE_HEAD` exists, because the staged tree during a merge legitimately contains every file the merge carried.

## Reproduction

Measured on branch `docs/closing-ref-fixes-vs-refs` while committing a `git merge origin/main` resolution:

```text
git diff --name-only origin/main...HEAD   ->    5 files   (real authored scope)
git merge-base HEAD origin/main           ->    a81239d0cb
git diff --name-only <merge-base>...HEAD  ->    5 files
MERGE_HEAD                                ->    49338f940d
detect_scope_explosion.py                 ->  322 files   BLOCKED (over the 50-file hard limit)
```

Both correct measurements agree on 5. The gate reported 322 and blocked the commit.

## Why this matters now

`main` moves fast enough that a branch of any age must merge before it can pass the ratchets. Every such merge carries well over 50 files, so the gate fires on the one commit shape that is guaranteed not to be scope explosion. The documented remediations do not apply either: you cannot "split into smaller PRs" or "stash" your way out of a merge commit.

The effect is that the sanctioned escape hatch `SKIP_SCOPE_CHECK=1` becomes routine for merge commits, which erodes its value as a signal for genuinely large PRs.

## Expected

When `MERGE_HEAD` is present, count only the files the branch authored, for example `git diff --name-only <merge-base>...HEAD`, or exclude paths whose staged content matches the merged parent. A merge that carries upstream files unchanged is not scope the author introduced.

## Duplicate search

Checked open issues for `scope`, `explosion`, `50-file`, `merge commit`, `lefthook`, `pre-commit`, `commit gate`. Nearest is #4492, which is a different hook and mechanism: it covers pre-push `{push_files}` glob selection when `@{push}` is unresolvable. This one is the pre-commit `scope-policy` job and `detect_scope_explosion.py`. Both share the theme of counting files outside the branch delta, so they may deserve a common fix, but neither report covers the other's code path.

Refs #4492.
Refs #2207.

### LINKED ISSUE #4580: implementer blocks on all three .agents/ gates when the consumer owns the directory (#1908 carve-out keys on absence)
## Summary

`implementer` hard-stops whenever `.agents/` exists in the working repo without `HANDOFF.md` and `AGENT-INSTRUCTIONS.md`. The vendor-install carve-out from #1908 keys on `.agents/` being absent, so a consumer whose repo contains `.agents/` for any other reason never reaches it. I hit this with a consumer-owned `.agents/`; the run is below.

This repo's own rule points at a second route to that state with no consumer error at all. `.claude/rules/plugin-self-containment.md:27` says `.agents/` stays behind and is not part of what a marketplace consumer receives. Line 38 of the same file classifies a *path reference* inside a shipped file, one telling an agent to write its output into the consumer's `.agents/`, as "the plugin doing its job. Not a defect." That row rules on the reference, not on the runtime write and not on what the gate then does with the resulting directory. But a write of that shape would create the directory the gate reads as a torn scaffold. I did not execute that route; the case I executed is the consumer-owned one.

This is a disagreement with a deliberate design decision, not a violation of one. `agents/implementer.agent.md:114` names this case and rules it in scope on purpose. My argument is that the state the gate keys on does not have a unique cause outside this repository, and that a shipped file telling an agent to write into the consumer's `.agents/` is a reference this repo's own rule file classifies as not a defect.

Terminology: by "vendor install" I mean any consumer repo that obtained the agent definitions without this repo's `.agents/` scaffold. Mine is a plugin install, version 0.6.170, which is one such case.

One question sizes this whole issue: does the toolkit reserve `.agents/` as its own namespace inside a consumer's workspace? If yes, most of what follows collapses to a one-line message fix. "Why it matters" at the end states exactly what survives each answer, so you can stop reading there if the answer is obvious to you.

## The argument

Line 105 fires on a state, not a cause: `.agents/` present, `HANDOFF.md` missing. Line 114 supplies the reading:

> The hard stops still fire when `.agents/` is present but incomplete. That case is the real misconfiguration the gate guards against.

Inside this repo, where `.agents/` is the toolkit's own scaffold, that reading has one plausible cause and is sound. In a repo that is not this one, the same state has two others:

1. residue of a prior toolkit run, since the toolkit writes into `.agents/` (see the write path below)
2. a consumer using the name for something unrelated, which is my case

Neither is a torn scaffold, because there was no scaffold to tear. The gate cannot tell either one from the case it was built for, and the cost of reading them as that case is the whole agent.

Case 1 is the one I would look at first, because `.claude/rules/plugin-self-containment.md` uses a destination under the same `.agents/` parent as its worked example of a reference that is not a defect. Its example is `.agents/planning/`; the agents below target `.agents/retrospective/`. The gate keys on the parent, so the distinction does not matter to it.

Line 27 of that rule establishes what a marketplace consumer receives:

> Each marketplace entry names exactly one `source` directory. A consumer who installs the plugin receives that directory and nothing above it. `docs/`, `.agents/`, `build/`, `scripts/`, `templates/`, and the repository root stay behind.

So a consumer who installs this way is shipped no `.agents/`, and therefore neither `HANDOFF.md` nor `AGENT-INSTRUCTIONS.md`. Mine is that kind of install, and its plugin root holds `.claude-plugin/`, `agents/`, `docs/`, `hooks/`, `instructions/`, `lib/`, and `skills/`, and nothing else. I have not checked what other distribution methods you support.

Line 38 is a row in that file's classification table, deciding whether a path reference inside a shipped file is a portability defect:

> **Consumer-workspace path**: a location in the installing repository that the agent reads or writes | an agent told to write its output to `.agents/planning/` or `docs/adr/` in the consumer's repo | Fine. This is the plugin doing its job. Not a defect.

Read narrowly, that row is about references, not runtime, and it grants no permission by itself. What it does establish is the intent behind such a reference: the example it chose is an agent told to write into the consumer's `.agents/`, and the verdict on that example is that the plugin is doing its job.

Put beside the gate, that is the problem. A shipped agent may correctly be told to write into the consumer's `.agents/`. Doing so creates the directory. Nothing in that write creates `HANDOFF.md`. `implementer` then stops, in a repo that had no toolkit scaffold until the toolkit put a directory there. At the gate, that directory is indistinguishable from a torn scaffold.

`agents/retrospective.agent.md:717` and `:1408` are such a reference, to `.agents/retrospective/`, and I found no vendor carve-out in that agent. `agents/pr-comment-responder.agent.md:486` targets the same tree. I did not execute any of them. I am reading destinations off the agent text and matching them against the rule's own example, not reporting an observed write. I also did not verify that those writes are reached without other preconditions, nor that running `retrospective` before `implementer` is a supported order.

One question this does not answer: whether the toolkit reserves `.agents/` as its own namespace inside the consumer's workspace. Line 38 places the directory in the consumer's workspace but says nothing about who owns the name, so both readings survive it. That answer decides case 2, the one I executed. It would not decide case 1, since there the directory would be the toolkit's own output, though I have not executed case 1 and cannot report it as a second observed cause.

## What blocks

`agents/implementer.agent.md:104` states the principle the carve-out protects, and the degradation it prescribes when `.agents/` is absent: skip the scaffold gates, still read the consumer's root `AGENTS.md` and `CLAUDE.md` if present, note `[INFO] Vendor install: no .agents/ scaffold`, and proceed. Its closing sentence gives the reason:

> A consumer that installed only the agent prompt should not be refused service for lacking files it was never shipped.

`agents/implementer.agent.md:105` through `:107` are three separate stops, all keyed on the same directory:

> If `.agents/` exists but `.agents/HANDOFF.md` is missing → stop and report `[BLOCKED] No prior session context available`. Do not proceed.

> If `.agents/` exists but `.agents/AGENT-INSTRUCTIONS.md` is missing → stop and report `[BLOCKED] Project configuration incomplete`.

> If `.agents/` exists but the root `AGENTS.md` is missing → stop and report `[BLOCKED] Missing root agent instructions`.

The third did not fire for me, because my consumer repo has a root `AGENTS.md` of its own. It would fire for a consumer that has a `.agents/` directory and no root `AGENTS.md`. A plugin install is not literally "installed only the agent prompt", but it is equally a case where the scaffold files were never shipped. The detection tests for the directory, so the principle at line 104 does not reach it.

## Repro

Preconditions: project-toolkit installed as a plugin (0.6.170 here), in a repo that has never contained this repo's `.agents/` scaffold.

```bash
mkdir -p .agents/tools          # any consumer-owned use of the name
ls .agents/HANDOFF.md           # expect: No such file or directory
```

Invoke the `implementer` agent with any task.

Observed, in my consumer repo, on 0.6.170. I gave it a read-only task, report the current git branch, and instructed it to follow its startup procedure and to report any stop verbatim rather than work around it. It refused, and reported two conditions:

- `[BLOCKED] No prior session context available`, because `.agents/` exists and `.agents/HANDOFF.md` is missing
- `[BLOCKED] Project configuration incomplete`, which it said would independently trigger because `.agents/AGENT-INSTRUCTIONS.md` is also missing

It described both as mandatory preconditions whenever `.agents/` is present.

Lines 105 through 107 are three separate gates keyed on the same directory, so patching only the `HANDOFF.md` check leaves the other two. The run refused on the first and named the second; I did not isolate the second by restoring `HANDOFF.md` alone, and the third did not arise because my repo has a root `AGENTS.md`. The task needed no prior session context of any kind. The block is on `implementer`; I did not test the other agents that share this pattern.

Expected: one of two, and which one depends on the namespace question above.

If the toolkit does not reserve `.agents/`, the agent recognizes there is no toolkit scaffold and degrades the way line 104 already prescribes for the absent case: skip the scaffold gates, read the consumer's root docs, note the vendor install, proceed.

If the toolkit does reserve `.agents/`, then the block is the right outcome and I withdraw the severity claim below for my own case. What is left is that none of the three messages says so. Each names a missing scaffold file, which tells a consumer who was never shipped any of them nothing about the actual rule. Saying "`.agents/` is reserved by this toolkit; rename your directory" would cost one line in each, and would be self-serving to you as well, since the current messages invite exactly the bug report you are reading. I am not assuming an opt-out exists; if none does, saying so plainly is the whole ask on this branch.

What I am asking not to happen is the current outcome, which is neither.

Recovery available in #1908 does not apply. That carve-out triggers only when the directory is absent, so reaching it means removing or renaming a directory the consumer is using.

Context you cannot check from your repo, offered only as provenance: I hit this in `Wcd.Infra.ConfigurationGeneration` on Azure DevOps, Windows, Copilot CLI. Its `.agents/` holds only `tools/`, a set of Azure DevOps PowerShell scripts unrelated to this toolkit, and none of `HANDOFF.md`, `AGENT-INSTRUCTIONS.md`, `retrospective/`, or `architecture/`. Nothing else about that repo is load-bearing: the preconditions listed above are the whole of what is required. I did not re-run it in a fresh repo to demonstrate that.

## The write path, quoted

I did not execute these. I am reporting documented destinations, not observed writes:

- `agents/retrospective.agent.md:717` -> ``Save to: `.agents/retrospective/YYYY-MM-DD-[scope].md` ``
- `agents/retrospective.agent.md:1408` -> ``Save retrospective document to `.agents/retrospective/` ``
- `agents/pr-comment-responder.agent.md:486` -> `Save analysis to: .agents/retrospective/PR-[number]-needs-split-analysis.md`

All three are repo-relative, so in a consumer they resolve against the consumer's tree. That is the shape `plugin-self-containment.md:38` classifies as the plugin doing its job. None of them writes a `HANDOFF.md`. I skipped a `retrospective` run this session rather than find out what happens.

## Possible directions

Framing, not a prescription. I do not know your constraints here.

1. Detect toolkit ownership with a positive signal, such as a marker file the toolkit itself writes, rather than the presence of `.agents/`. This does not fully separate a vendor install from a torn toolkit install, since a torn install could also lack the marker. That tradeoff is yours: today the ambiguity is resolved against the consumer, and it costs them the agent entirely.
2. Reconcile the gate with the rule. They address different concerns and can both stand as written, which is why this is easy to miss: `plugin-self-containment.md:38` classifies a reference, `implementer.agent.md:105` gates on runtime state. The composition is what bites, because a reference the rule calls correct produces the state the gate rejects. Deciding which one yields, and saying so in the block message, is cheaper than any detection change.
3. Sweep `agents/*.agent.md` for `.agents/` references to size this class in one pass. `_VENDOR_FORBIDDEN` in the skill tests is a ready-made pattern for it, though it cannot be applied to the agent files as a blocking assertion, because it would flag the HANDOFF gate text itself.
4. On the write path, do not create `.agents/` subdirectories in a repo with no toolkit marker. Write under the plugin's own directory or a caller-supplied output path, and report which one was used.

## Why it matters

The failure is total rather than degraded: a task needing no prior context at all was refused. The recovery documented in #1908 is unavailable, because it triggers only when `.agents/` is absent, and the consumer needs that directory present.

How much of that survives your answer to the namespace question:

- If `.agents/` is **not** reserved, the whole of the above stands, and it applies to both cases.
- If `.agents/` **is** reserved, the refusal is correct for my case 2 and the severity claim shrinks to a message defect: the consumer is told a scaffold file is missing when the real rule is a name collision, and the only recovery is one they cannot infer from the message. Case 1 is untouched by that answer, since there the directory would be the toolkit's own output, but I have not executed case 1, so I am not asking you to weigh it as observed.

The structural proposals above are entailed only by the first branch. On the second branch, the ask is item 2 alone, and only its message half.

## Environment

- project-toolkit 0.6.170, installed at `~/.copilot/installed-plugins/ai-agents/project-toolkit`
- Copilot CLI on Windows
- Consumer repo has no toolkit scaffold files

### LINKED ISSUE #4602: detect_scope_explosion.py silently passes on a detached HEAD
`scripts/detect_scope_explosion.py` returns 0 on a detached `HEAD` for staged content that returns 1 on a branch. The gate does not report that it skipped; it reports success.

## Reproduction

Same commit, same staged files, same tree. The only variable is whether `HEAD` is attached.

```
git worktree add --detach ~/probe <branch-tip>
cd ~/probe
git merge --no-commit --no-ff <origin/main tip>
uv run --frozen python scripts/detect_scope_explosion.py; echo "rc=$?"
# rc=0, 97 files staged

git checkout -B probe/scope-merge-check <branch-tip>
git merge --no-commit --no-ff <origin/main tip>
uv run --frozen python scripts/detect_scope_explosion.py; echo "rc=$?"
# rc=1, BLOCKED: 105/50 files
```

Both directions reproduce on demand.

## Why it matters

Two costs, and the second is the expensive one.

A detached `HEAD` is not exotic. `git bisect`, `git worktree add --detach`, a CI checkout of a SHA, and a rebase in progress all detach. Committing from any of those states skips the scope gate with no signal.

The larger cost is that it corrupts measurement of the gate itself. A probe run detached reports PASS on input the gate blocks, so anyone testing this check the obvious way gets a false negative. That is not hypothetical: it is how #4544 came to be closed as already-fixed while still reproducing. Any past finding about this detector that was measured on a detached `HEAD` should be re-measured attached.

## Suggested direction

Whatever the base-ref resolution does when it cannot find a branch, the one thing it must not do is exit 0. A gate that cannot determine its own scope should say so and fail, or state plainly that it is skipping and why. Exiting 0 makes "I did not check" indistinguishable from "I checked and it is fine," which is the same defect class as #4504 and #4441.

rjmurillo:
## Summary

Integration branch that lands 16 open pull requests in one CI cycle.

Branch protection on `main` sets `strict_required_status_checks_policy=true` with 17
required checks. Measured consequence: every merge to `main` marks every other open PR
`behind`, so each PR needs a branch update plus a full CI cycle before it can merge.
A single PR was measured end to end at **11m38s** wall clock (`update-branch` at
11:34:04Z, checks green at 11:45:39Z, merge at 11:45:42Z). Landing 16 PRs one at a
time costs roughly 3.1 hours of strictly serial time. This branch does it in one cycle.

## Constituent pull requests

Merged heads: #4604, #4606, #4609, #4610, #4611, #4612, #4614, #4616, #4617, #4618,
#4619, #4620, #4621, #4623, #4626, #4627.

PR #4622 is deliberately excluded. It is built on a stale base and would revert the
already-merged #3972 fix, and it duplicates #4617. The exclusion was verified by
comparing resulting file content rather than the three-dot diff, which cannot show a
revert-by-stale-base because it diffs against the merge base.

## Issues closed by this rollup diff

Fixes #4497
Fixes #4524
Fixes #4526
Refs #4527
Fixes #4544
Fixes #4580
Fixes #4602
Refs #4605

The following fixes are already on `main` after #4611, #4612, #4613, or #4620. They are not claimed by this rollup diff because the spec gate validates only files changed relative to current `main`: issues 4377, 4519, 4520, 4600.

## Integration-level regression found and fixed

Each constituent PR measured its own branch against `main` and passed the taste count
ratchet at or below the baseline of 583. The merged result measured 584. No per-PR gate
can see this, because each PR is green in isolation and the regression only exists in
the combination. `merge_tree_ratchet_check.py` exists for exactly this case and caught it.

Isolating the delta required diffing error-severity violation sets between this tree and
a pristine `origin/main` worktree. Nine of the ten apparent deltas were the same
file-size violations with changed line counts, because the violation message embeds the
line count. The single net-new violation was
`tests/ci/test_pr_validation_workflow.py::_history_offenders_for_merge_base_jobs` at
cyclomatic complexity 11 against a maximum of 10.

Fix: extracted `_shallow_checkout_offenders(job_id, steps)`, preserving offender
ordering. 55 tests pass. The ratchet now reports `OK (count == baseline 583)`.

## Conflict resolution

Three constituent PRs collided add/add on `.agents/sessions/2026-08-05-session-9999.json`
and its episode file. The merge base is empty on both sides, confirming a true add/add
rather than divergent edits. These are real session logs from three different agents,
each of which picked the placeholder number 9999 to satisfy the gate that requires a
session log when staging under `.agents/**`. Nothing enforces uniqueness, so parallel
agents collide. That is open issue #4561.

Resolved by preserving all sides, renumbering the incoming logs into free slots and
rewriting the internal `session.number` to match. No session log was dropped.

## Validation

- Full pre-push suite green: 46 validators passed, 0 failed.
- `python-tests` 943.30s, 0 failures.
- Taste count ratchet `OK (count == baseline 583)`.
- Merge-tree ratchet unblocked.

## Notes for review

`SKIP_SCOPE_CHECK=1` was set for the push. `detect_scope_explosion.py` measures
branch-total files against the merge base rather than per-commit files, which makes any
integration branch structurally unable to commit. The variable is that tool's own
documented override for justified large changes and is not one of the forbidden hook
bypasses.

## Correction: #4588 is NOT closed by this branch

This branch originally carried a closing reference to issue 4588. That was wrong and the
reference is removed. The issue number is written without a hash below so that no closing
keyword can bind to it.

Issue 4588 needs the duplicate-preserved-event timestamp upgrade. Measured directly by
comparing the resulting file rather than by trusting either PR description:

- This branch, carrying #4617: `_preserved_timestamp` exists, but there is no
  duplicate-timestamp upgrade path.
- PR #4622: carries a dedicated function documented as
  "Upgrade a duplicate event from synthesized midnight to source time",
  taking the incoming event and the synthesized midnight.

An independent reproduction confirmed the behavioral difference. On this branch and on
`origin/main`, event `e009` stays at `2026-08-04T00:00:00+00:00` and the edge
`e009 -> e010` survives. On the corrected #4622, `e009` becomes
`2026-08-04T10:26:49+00:00` and the order becomes `e011 -> e009`, matching `--force`.

#4617 is closed and superseded. Issue 4588 lands through #4622, which has since merged
`origin/main` and therefore no longer carries the stale base that originally excluded it
from this rollup.

## Closing claim audit correction

- `Refs #4527`: The diff lacks Data.action and per-thread reporting for missing replies or commit SHAs.
- `Refs #4605`: the rollup retains the same GOTCHAS.md conflict path as PR #4618.


rjmurillo:
Closing this rollup. It is not viable in its current shape, and the reason is structural rather than fixable by another rebase.

## Why

The rollup mentions 31 pull requests. Only 12 of those are still open: #4604, #4606, #4609, #4610, #4614, #4616, #4618, #4621, #4623, #4626, #4627, and this PR itself. The rest have merged or closed since it was cut, so a large part of what the branch carries is already on `main` or was abandoned.

Its own merge state is `DIRTY`, and the conflict is real rather than a stale GitHub cache. Verified locally in a scratch worktree rather than trusting the API verdict:

```
git merge-base --is-ancestor origin/main HEAD   -> exit 1
git merge --no-commit --no-ff origin/main       -> exit 1
```

Ten files conflict, across memory, workflow, validation, and test paths. Four of the PRs it rolls up (#4609, #4610, #4618, and this one) each carry real conflicts of their own, so resolving the rollup means resolving those conflicts a second time, in a branch where a bad resolution is harder to attribute back to the PR that introduced it.

## What replaces it

The premise behind the rollup was sound: one CI cycle instead of sixteen. But the thing that made sixteen cycles expensive was not the number of PRs. It was that a shared gate was red for a reason unrelated to any of them. Two systemic causes are now identified with measurements:

1. The merge-tree ratchet failed on every branch behind its base, because four earlier steps in the same job ran `git fetch --depth=1` and grafted `.git/shallow` onto a `fetch-depth: 0` checkout. That was the single check failing on 28 open PRs. Covered by #4518 and PR #4642 for `pr-validation.yml`, and by #4680 and PR #4681 for the remaining instance in `pytest.yml`.
2. `Aggregate Results` blocked 17 PRs because the upstream review agents hit `HTTP 403: API rate limit exceeded for user ID 6811113` and wrote `DID_NOT_RUN`. The jobs were spending a human user's API budget rather than an isolated runner token.

Fixing those two unblocks the same PRs individually, without a merge that has to be resolved twice.

## If a rollup is still wanted

Cut a fresh one from current `main` against the 12 still-open PRs, after the two fixes above have landed. Rebuilding is cheaper than untangling this branch, and the new one will not be carrying 19 references to work that has already left the queue.

Nothing here is lost. Every PR it references remains open and independently mergeable.

Issue #4698, opened 2026-08-06
A session set core.bare=true in shared .git/config, breaking the main worktree and 3 of 5 linked worktrees; no gate detects it


rjmurillo:
## Problem

A working session left `core.bare = true` in the repository's shared `.git/config`. A bare repository cannot have work trees, so this broke the main worktree and three of five linked worktrees at once. Every affected worktree failed every git command that needs a work tree:

```
fatal: this operation must be run in a work tree
```

`git status`, `git diff`, and `git rev-parse --show-toplevel` all failed. Repairing it was a single command, `git config core.bare false`, after which all six worktrees returned to normal immediately.

## Why it is worth a gate rather than a shrug

The failure does not present as a git problem. It presents as several unrelated test and gate failures at once, each of which invites its own wrong diagnosis. In this session it surfaced as four separate symptoms:

- Two pre-PR gate checks failing, "Documented Interpreter Portability" and "Skill Markdown Portability", which look like portability defects.
- A lefthook integration test failing, which looks like a test-isolation defect. I filed an issue for that and had to close it as invalid.
- `git diff` failing inside a worktree, which looks like a corrupt worktree.

Three separate diagnoses were attempted before the shared cause was found. The generalizable lesson is that when several unrelated checks fail simultaneously, a shared substrate is a better first hypothesis than several independent causes, but the tooling gives no signal pointing there.

## Likely mechanism

The repository has `extensions.worktreeConfig = true`:

```
[extensions]
	worktreeConfig = true
```

When that extension is enabled, `core.bare` is worktree-specific and belongs in the main worktree's `config.worktree` file, not in the shared config. A `core.bare` written to the shared config is inherited by every linked worktree, which is exactly the observed blast radius. Any code path that runs `git config core.bare true` without an explicit repository target, while the current directory resolves to this repository, produces this state.

I could not identify the specific writer. No source file in this repository sets `core.bare` directly, so the likely candidates are a test that creates a bare fixture repository and runs a `git config` or `git init --bare` with the wrong target directory, or a tool invoked during a gate.

## Second problem in the same file

`.git/config` is 44648 bytes across 1268 lines, with 422 sections, of which 416 are `[branch ...]` entries. Four are leftover test branches named `wt/t_<hex>` and `wt/push-batch`. Ordinary repository config is a few hundred bytes. Branch sections accumulate and nothing prunes them, so the file grows without bound as worktrees and branches are created by automation.

[...]

Issue #4425, opened 2026-08-03
session-init and session-end deadlock: a new session log cannot be committed in either order


rjmurillo:
## Summary

The documented session-end flow cannot complete for a session log created on a feature branch. The pre-commit `session-policy` hook refuses to commit a session log whose `sessionEnd` MUST items are incomplete, and `complete_session_log.py` refuses to mark `changesCommitted` while any file is dirty. The session log is always that dirty file, so neither side can go first.

I hit this on `docs/grep-ere-pipe-escape` and had to hand-populate `sessionEnd` to get out.

## Repro

```bash
uv run --frozen python .claude/skills/session-init/scripts/new_session_log_json.py \
  --session-number NNNN --objective "..."
git add .agents/sessions/<file>.json
git commit -m "docs(session): add session log"
```

Result, from `session-policy`:

```text
Session validation FAILED:
  Incomplete MUST: sessionEnd.validationPassed
  Incomplete MUST: sessionEnd.serenaMemoryUpdated
  Incomplete MUST: sessionEnd.markdownLintRun
  Incomplete MUST: sessionEnd.changesCommitted
  Incomplete MUST: sessionEnd.handoffPreserved
  Incomplete MUST: sessionEnd.checklistComplete
exit status 1
```

Now try the other order. Leave the log uncommitted and run the completion script:

```bash
uv run --frozen python .claude/skills/session-end/scripts/complete_session_log.py \
  --session-path .agents/sessions/<file>.json
```

Result:

```text
[TODO] Uncommitted changes exist - commit before completing
[FAIL] Session validation failed. Fix issues above and re-run.
```

## Why neither order works

`complete_session_log.py:287` treats any porcelain output as uncommitted:

```python
def _test_uncommitted_changes() -> bool:
    result = subprocess.run(["git", "status", "--porcelain"], ...)
    if result.returncode != 0:
        return True
    return bool(result.stdout.strip())
```

The session log is itself a tracked-or-untracked change, so it always appears in that output. `changesCommitted` (line 543) can only be set when the tree is clean, and the tree is not clean while the session log is waiting to be committed.

On the other side, `lefthook.yml:62` runs the policy on every staged path under `.agents/**`:

```yaml
- name: session-policy
  run: uv run --frozen python scripts/validation/git_hook_policy.py session {staged_files}
  glob: ".agents/**"
  skip:
    - merge
```

[...]

PR #4703, opened 2026-08-06
docs(gotchas): record four measured push-time and repository-state traps

rjmurillo:
## Summary

Records four repository traps measured during a fleet session that triaged 95 open issues. Each cost real diagnostic time, none is inferable from reading the code, and each entry states the symptom as it actually prints, the measurement that separates it from a real defect, and the repair.

Documentation only. No behavior changes.

Refs #4698

## Why these four and not more

`.claude/rules/testing.md` already carries fifteen rules covering the test-quality lessons from this session, including per-consumer wiring, picking an input the unfixed implementation gets wrong, and assertions no input can falsify. Adding restatements would spend always-on rule budget for nothing, so nothing was added there.

The gap was push-time and repository-state behavior, which is what `.agents/governance/GOTCHAS.md` exists for and which `AGENTS.md` already points at.

## The entries

**A green pytest run is not a green push.** The pre-push stage runs roughly twenty jobs and pytest is one. One branch passed 23693 tests, zero failures, and was still rejected by five gates pytest never executes. The `python-type-check` failure was four trivial mypy errors in that branch's own new test files, which no test could catch because a passing test does not type-check itself. The cost is asymmetric: that gate failed in 1.30 seconds when run directly, and took eleven minutes to discover through a push that also truncated the output. The entry names the gate set, points at `lefthook.yml` as the source of truth for invocations, and records that the mypy gate deliberately refuses a bare call.

**A staged file in the shared checkout fakes a red main.** Whole-tree ratchets measure the tracked tree, which includes staged but uncommitted files. A `git add` that was never committed is invisible to `git log`, to `git diff HEAD`, and to a HEAD versus origin comparison, yet it moves the measurement. Measured: a ratchet reported 587 against a baseline of 583 on a checkout whose HEAD matched `origin/main` exactly, caused by 68 staged additions from another agent. Unstaging returned the test to green. The entry gives the discriminating command and specifies `git restore --staged`, which preserves the files, rather than a checkout that would discard someone else's work.

**core.bare=true appears during push and breaks every worktree.** It broke the main checkout and three of five linked worktrees at once and surfaced as four unrelated-looking failures, producing three separate wrong diagnoses before the shared cause was found. A serial suite run does not reproduce it, so the trigger involves the parallel pre-push stage. The entry includes the worktree-level immunization, which works because this repository sets `extensions.worktreeConfig`, and it was verified by flipping the shared value to true afterwards and confirming all six worktrees kept working.

**When several unrelated checks fail at once, suspect the substrate.** The generalization of the above, plus a second instance: a test that passed in three full-suite runs and failed alone looked order-dependent, but those runs differed in time rather than ordering. That misreading produced an issue that had to be closed as invalid.

## Notes

Two constraints worth recording for whoever reviews this.

`markdownlint-cli2` on the changed file reports `Summary: 0 issues in 0 files`, because `.agents/**` is excluded by the lint ignore list. That is a zero-scope result, not a pass, so it is recorded that way in the session log rather than cited as evidence.

The session log hit the deadlock in issue #4425 from a third angle. A commit touching `.agents/` must stage the session log, a staged log must have a complete `sessionEnd`, and a complete `changesCommitted` wants an `endingCommit` that cannot exist until the commit is made. Leaving it empty produced an episode with `commits: 0` alongside `files_changed: 3`, which the episode metrics ratchet correctly flags as failed commit collection and which grew the violation count from 21 to 22. Resolved by recording the SHA in a follow-up commit rather than amending, since GOTCHAS warns specifically against amending after recording it.

## Changed Files

- `.agents/governance/GOTCHAS.md`: the four entries.
- `.agents/sessions/2026-08-06-session-10003.json`: session log, required in the same commit by the `.agents` session policy.
- `.agents/memory/episodes/episode-2026-08-06-session-10003.json`: generated episode, regenerated after the ending commit was recorded.


rjmurillo:
owner: local:pr-autofix
session: session-10004
acquired_at: 2026-08-07T01:17:18Z
expires_at: 2026-08-07T01:32:18Z
base_sha: c97ca405efd5cb4251bea11e276536feeee35774

rjmurillo:
owner: local:pr-autofix
session: session-10004
acquired_at: 2026-08-07T01:24:15Z
expires_at: 2026-08-07T01:39:15Z
base_sha: c97ca405efd5cb4251bea11e276536feeee35774

rjmurillo:
owner: none
session: session-10004
acquired_at: 2026-08-07T01:41:32Z
expires_at: 2026-08-07T01:41:32Z

rjmurillo:
owner: local:pr-autofix
session: session-48c94212-resolve-dirty
acquired_at: 2026-08-07T02:46:12Z
expires_at: 2026-08-07T03:01:12Z
base_sha: fca70f0c3c7e0a4e23462912ce97d0a218a96ab1

rjmurillo:
owner: local:pr-autofix
session: session-4703-disarm-automerge-b
acquired_at: 2026-08-08T01:03:27Z
expires_at: 2026-08-08T01:18:27Z
base_sha: 1f2ffeb04962168a213dddda5ee0bcd649ccf273

rjmurillo:
owner: local:pr-autofix
session: session-4703-disarm-automerge-b
acquired_at: 2026-08-08T01:04:59Z
expires_at: 2026-08-08T01:19:59Z
base_sha: 1f2ffeb04962168a213dddda5ee0bcd649ccf273

rjmurillo:
owner: none
session: session-4703-disarm-automerge-b
acquired_at: 2026-08-08T00:50:01Z
expires_at: 2026-08-08T00:50:01Z

Issue #2761, opened 2026-06-27
tech-debt(worktrees): 113 accumulated worktrees (10.6G) starve the markdown LSP; add .worktrees gitignore + worktree GC


rjmurillo:
## Problem

Git worktrees created for agents and PRs accumulate without cleanup. Current state on one machine: **113 registered worktrees totalling 10.6G** (`.claude/worktrees` 5.4G, `.worktrees` 5.2G). This produced **333,979 markdown files** under the workspace root (vs 3,712 tracked in the repo).

## Impact (breaks tooling, 5+ min, recurring)

1. **Markdown language server (Marksman) times out.** Serena/Marksman crawls the workspace for markdown; with 150,893 `.md` under the non-gitignored `.worktrees/` it never finishes and hits serena's `timeout=235`, so the markdown LS dies. With the LS down, the LSP read/grep guards block native `Read`, forcing the `LSP_DOWN=true` escape hatch (see #2759). The real cause is the worktree markdown volume, not the escape hatch.
2. Slow `git status`, `find`, ripgrep, and any workspace-wide scan.
3. 10.6G disk.

## Two fixes

1. **Gitignore the bare `.worktrees/` dir** (separate PR). `.gitignore` already ignores `.claude/worktrees/` and `worktree-*/` but missed the top-level `.worktrees/`. Adding it drops the workspace `.md` Marksman sees from 333,979 to 3,712 (Marksman honors `.gitignore`). This is the immediate LSP-timeout fix.
2. **Worktree garbage collection.** Provide a cleanup script / periodic task that removes worktrees whose branch is merged or whose PR is closed, and prunes orphaned admin entries. Guard against data loss: never remove a worktree with uncommitted changes or unpushed commits; skip `locked` worktrees. 113 accumulated because nothing reaps them after the agent/PR finishes.

## Evidence
- `git worktree list | wc -l` => 113; `git worktree list | grep -c locked` => 10.
- `find . -name '*.md' | wc -l` => 333,979; `git ls-files '*.md' | wc -l` => 3,712.
- `du -sh .claude/worktrees .worktrees` => 5.4G, 5.2G.
- serena log: `markdown: Request timed out (timeout=235)` even after rooting on the repo.

Refs #2759.

rjmurillo:
[...]

PR #4708, opened 2026-08-06
fix(maintenance): make worktree GC answer one safety question in one place

rjmurillo:
# Pull Request

### What

Makes `gc_worktrees` answer one question with one function, and fixes the three
ways that function was reachable with a wrong answer. Each defect was built as a
real git repository and reproduced before the fix, then re-broken afterwards to
prove the fix is what catches it.

### Why

The tool's whole reason to exist is that a worktree's admin directory holds the
only reflog anchoring commits made in that worktree. Remove the directory and
those commits become unreachable with no warning and no undo.

Stacked on #4694. Review that one first. A second branch stacks on this one and
carries the loss channels found in later review rounds.

## Specification References

| **Issue** | Refs #2761 | Worktree accumulation starves the markdown LSP |

## Changes

Three defects, in the order the safety argument breaks down.

**1. The plan did not predict what apply does.** Dry-run listed candidates that
apply then refused, and refused candidates apply would have taken. The two ran
different logic. Now one function decides, and both call it.

**2. Stale entries were gated on git's `prunable` flag alone.** Git omits
`prunable` for a locked worktree and for an entry whose path is occupied by
something else, so both classes were invisible. A blanket `git worktree prune`
was also being recommended; it cannot be scoped and takes every sibling with it.
Removal is per path now.

**3. Clean, merged, and fully pushed did not mean safe.** All three describe
the worktree's current HEAD. They say nothing about a commit it reached and then
left. A detached worktree that commits and then checks out something else leaves
that commit anchored by its own `logs/HEAD` and by nothing under `refs/`. The
tree is clean, HEAD is an ancestor of the base, every branch is pushed, and the
commit dies with the admin directory. The reflog is now read before a removal is
approved.

One review finding was tested and rejected rather than implemented. A reviewer
proposed `-z` on the porcelain listing. A newline in a worktree path does split
the record, confirmed on git 2.43.0, but both reachable constructions were built
and run: with a worktree occupying the truncated path, removal targets that
worktree and a commit stranded in the newline one survived in the object
database; with nothing occupying it, the tool refuses because it cannot locate
the admin entry. Neither loses a commit, so the option was not adopted and the
parser docstring now records why.

The 517-line module was split into probe, reason, apply, and parse, because a
safety argument nobody can hold in their head is not a safety argument. The
entry point is 472 lines now, and the safety question lives in one place.

### Per-file changes

- `scripts/maintenance/gc_worktrees.py`: the one decision function, consulted by
  both plan and apply.
- `scripts/maintenance/_gc_stale.py`: the stale probes, including the reflog
  read that catches an abandoned commit.
- `scripts/maintenance/_gc_apply.py`: per-path removal, replacing the blanket
  prune.
- `scripts/maintenance/_gc_reasons.py`: the human-readable reasons, including
  every channel a stale entry can lose work through.
- `scripts/maintenance/_gc_parse.py`: porcelain parsing, plus why it does not
  pass `-z`.
- `scripts/maintenance/worktree_report.py`: reports through the shared decision.
- `tests/test_gc_worktrees_real_git*.py`: the real-git suites, where every
  defect above was reproduced before it was fixed.
- `tests/test_gc_worktrees_stale*.py`, `tests/test_gc_stale_probes.py`: the
  stale-entry suites. Their `_stale` helper now takes an explicit typed
  signature instead of splatting an untyped dict, so a misspelled field name
  fails at the call rather than building a different worktree.
- `.serena/memories/`, `.agents/sessions/`: three memories recording what a
  stale entry actually allows, why a session log cannot say "todo list" in
  evidence, and that a revert experiment which injects code runs it for real.

## Type of Change

- [ ] Documentation update
- [ ] Refactoring (no functional changes)

## Reviewer Expectations

**Review kind / scrutiny / urgency**: full, high scrutiny, no rush.

**Focus areas**: whether one decision function is genuinely consulted by both
callers, or whether a second path can still construct an approval. That
divergence is defect 1, and it is the kind that comes back.

Second focus: the reflog probe. It is the only thing standing between a
detached worktree's abandoned commit and deletion, and every cheaper signal
(clean tree, merged HEAD, pushed branches) reports safe in that case.

## Testing

- [x] Tests added/updated

Every fix here is backed by a test against a real git repository, not a mock.
The mock suite passed while real git caught a real bug several times during this
work, so the real-git suites are where the evidence lives.

```
uv run pytest tests/ -q -k "gc_worktrees or worktree_report or gc_real or gc_stale"
205 passed, 23621 deselected in 12.44s
```

Ratchets, both green at this commit:

```
uv run python scripts/ci/taste_count_ratchet.py
taste count ratchet: OK (count == baseline 583).

uv run python scripts/ci/memory_index_count_ratchet.py
memory index count ratchet: OK. 424 violations <= baseline 425 (-1 slack).
```

### Security Review

### Other Agent Reviews

Adversarial review across three model families. Every finding was reproduced
against real git before any code changed, and three were rejected on evidence.

Each fix was revert-proofed: the guard was deleted or weakened, the suite was
run, and the failures were counted. A guard whose removal breaks nothing is not
a guard.

## Author Pre-flight

- [x] Builds locally (or N/A)

### DIFF
--- a/.agents/memory/episodes/episode-2026-08-05-session-10005.json
+++ b/.agents/memory/episodes/episode-2026-08-05-session-10005.json
@@ -1,0 +1,318 @@
+{
+  "id": "episode-2026-08-05-session-10005",
+  "session": "2026-08-05-session-10005",
+  "timestamp": "2026-08-05T00:00:00+00:00",
+  "causal_order_version": 2,
+  "outcome": "success",
+  "task": "Make the worktree GC dry-run plan predict what --apply actually does, and stop it removing any worktree whose HEAD no ref contains",
+  "decisions": [],
+  "events": [
+    {
+      "id": "e001",
+      "timestamp": "2026-08-05T00:00:00+00:00",
+      "type": "milestone",
+      "content": "defect",
+      "caused_by": [],
+      "leads_to": [
+        "e025"
+      ],
+      "_source_session": "2026-08-05-session-10005"
+    },
+    {
+      "id": "e002",
+      "timestamp": "2026-08-05T00:00:00+00:00",
+      "type": "milestone",
+      "content": "correction",
+      "caused_by": [],
+      "leads_to": [
+        "e025"
+      ],
+      "_source_session": "2026-08-05-session-10005"
+    },
+    {
+      "id": "e003",
+      "timestamp": "2026-08-05T00:00:00+00:00",
+      "type": "milestone",
+      "content": "search-before-building",
+      "caused_by": [],
+      "leads_to": [
+        "e025"
+      ],
+      "_source_session": "2026-08-05-session-10005"
+    },
+    {
+      "id": "e004",
+      "timestamp": "2026-08-05T00:00:00+00:00",
+      "type": "milestone",
+      "content": "rejected",
+      "caused_by": [],
+      "leads_to": [
+        "e025"
+      ],
+      "_source_session": "2026-08-05-session-10005"
+    },
+    {
+      "id": "e005",
+      "timestamp": "2026-08-05T00:00:00+00:00",
+      "type": "milestone",
+      "content": "change",
+      "caused_by": [],
+      "leads_to": [
+        "e025"
+      ],
+      "_source_session": "2026-08-05-session-10005"
+    },
+    {
+      "id": "e006",
+      "timestamp": "2026-08-05T00:00:00+00:00",
+      "type": "milestone",
+      "content": "test",
+      "caused_by": [],
+      "leads_to": [
+        "e025"
+      ],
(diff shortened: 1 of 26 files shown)


rjmurillo:
owner: local:pr-autofix
session: session-10004
acquired_at: 2026-08-07T00:15:40Z
expires_at: 2026-08-07T00:30:40Z
base_sha: efc27ee9ba68d36a46a749211517be554368f3af

rjmurillo:
owner: none
session: session-10004
acquired_at: 2026-08-07T00:00:47Z
expires_at: 2026-08-07T00:00:47Z

PR #4728, opened 2026-08-07
fix(maintenance): stop worktree GC destroying commits its admin directory alone anchors

rjmurillo:
# Pull Request

### What

Closes three more ways `gc_worktrees` could remove a worktree and destroy the
only reference to a commit. All three are about the same thing: what a
worktree's git admin directory anchors, and how confidently the tool can say
"nothing". Each was built as a real git repository and reproduced before the
fix, then re-broken afterwards to prove the fix is what catches it.

### Why

`git worktree remove` deletes `<repo>/.git/worktrees/<name>/`. Anything that
directory alone was naming goes with it. Two channels live there that no check
run from the main worktree can see: the per-worktree ref namespaces
(`refs/worktree/`, `refs/bisect/`, `refs/rewritten/`), which `rev-list --not
--all` and `for-each-ref` both miss, and the worktree's own reflogs, which name
commits no ref names any more.

Stacked on #4718 (`fix/gc-worktrees-stale-entries`), which is stacked on
#4694 (`fix/worktree-walk-timeouts`). The middle link, the one-decision
branch, already merged as #4708, and that squash severed the ancestry under
#4718, so #4718 carries a merge commit that reattaches it. Review #4694, then
#4718, then this.

## Specification References

| **Issue** | Refs #2761 | Worktree accumulation starves the markdown LSP |

## Changes

Three defects, in the order they were found.

**7. `git rev-list --not --all` cannot see another worktree's local refs.**
`refs/worktree/`, `refs/bisect/`, and `refs/rewritten/` are per-worktree
namespaces that git stores under `.git/worktrees/<name>/refs/`, not in the shared
ref store. Every ref query run from the main checkout is blind to them, and
`update-ref` on one writes no HEAD reflog entry, so the reflog probe added for
defect 5 misses them too. A commit anchored only by `refs/worktree/mywork` read
as safe to every probe in this branch, and `git worktree remove` deleted the
anchor with the directory. Reproduced end to end on git 2.43.0, with a negative
control that loses the commit at the next `gc --prune=now`.

The probe now walks `<admin>/refs/**` and feeds those oids into the same
reachability query the reflog oids already use, at the cost of one file walk and
no extra subprocess. It skips `ref:` symrefs and the null oid, and answers
"unknown" rather than "no risk" on a file it cannot read or parse. The probe, the
warning, and the keep reason are renamed from "reflog" to "admin", because the
reflog is no longer the only thing they cover, and a user told to check the
reflog would find nothing there.

Defect 4 came back in a narrower form and is fixed here too:
`git checkout-index --prefix=DIR/` creates `DIR` only when it writes at least
one file, and it writes none for an index of unmerged stages. So the `cp` that
carried the stages died with `No such file or directory` in exactly the two
cases it was added for. The chain now leads with `mkdir -p` and carries
`git -C`, so it also runs from outside the repository.

**8. `logs/HEAD` is not the only reflog the removal deletes.** The reader added
for defect 5 opened one file. `update-ref --create-reflog` on a per-worktree ref
writes `<admin>/logs/refs/worktree/<name>`, and that reflog goes on naming a
commit after the ref itself has moved off it. So a commit could clear the ref
probe from defect 7 (no ref names it) and the reflog probe from defect 5 (not in
`logs/HEAD`) while a file in the admin directory was its last remaining name.

The reader now walks every reflog under `<admin>/logs/`. Two things that walk
had to get right, each of which cost a test:

- `os.scandir`, not `Path.rglob`. `rglob` swallows the `PermissionError` from an
  unreadable subdirectory and yields nothing, which this module would have read
  as "no anchors here", the exact silent all-clear it exists to prevent.
- `stat` plus `nothing_at`, not `Path.is_dir`, which reports `ENOTDIR` the same
  way it reports a missing directory. A corrupt admin record has to answer
  "unknown", not "empty".

**9. One good line in a reflog vouched for the whole file, and a reflog that
vanished mid-walk read as empty.** Two ways the reader from defect 8 handed back
a confident answer it had not earned.

The parser marked a file understood if any line anywhere in it opened with an
object id, then skipped every line it could not parse. A reflog with one good
line and one garbage line returned the ids from the good line alone while
signalling success, and downstream a partial list reads as "these are the only
ids at risk". That is the same silent all-clear defect 8 was about, one layer
in. Every line git writes to a reflog opens with the old and the new id, so the
judgement is now per non-blank line: a line that does not is a line this reader
does not understand, and the whole file is unknown.

Separately, the path handed to the reader always came from the walk, which had
just seen the file. Absence at read time therefore means the file went away
underneath the probe, not that there was never one. Reading that as an empty
reflog would clear a worktree from a snapshot the probe already knows is stale.
Absence is now unknown; a missing `logs` directory stays the only route to an
empty list.

The over-strictness risk here runs the other way: a parser that rejects a
reflog real git writes turns every healthy worktree into "unknown" and the tool
stops cleaning anything. A real-git test drives an amend, a checkout and a
branch move through the stricter parser to hold that line. `_OID` already
covered both 40-hex and 64-hex ids, so a `sha256` repository parses too.

### Three findings from adversarial review that are not defects

An adversarial review on GPT-5.6 Sol raised four Critical findings. Each repro
ended with "removed the worktree, ran `gc --prune=now`, object gone", which
cannot distinguish a real loss channel from an object git was never protecting.
Adding the missing half, run `gc --prune=now` with the worktree still in place,
settles it. Measured on git 2.43.0:

| Claim | Survives gc with worktree | After removal | Verdict |
|---|---|---|---|
| `ORIG_HEAD` is a lost anchor | survives | survives | Not a loss. The *shared* branch reflog holds it and `worktree remove` does not touch it. |
| `FETCH_HEAD` is a lost anchor | pruned | pruned | git never protected it. |
| A blob named by `refs/worktree/` | pruned | pruned | `rev-list` is commit-oriented, and git does not protect the blob either. |
| A symlinked `refs/` directory | pruned | pruned | git resolves the ref; gc prunes regardless. |

Only the reflog finding survived its control, and that is defect 8 above. The
symlink observation is real in one narrower sense: `Path.rglob` genuinely will
not descend a directory symlink where git resolves through it. That is not a
loss channel, but a walk that skipped such a subtree would clear anchors it
never opened, so `walk_files` answers "unknown" on meeting one.

### Per-file changes

- `scripts/maintenance/_gc_anchors.py`: new. The two admin-directory readers and
  the walk beneath them. Split out of `_gc_stale.py`, which was answering three
  questions in one file and had crossed the size cap.
- `scripts/maintenance/_gc_files.py`: new. `regular_file` and `nothing_at`, the
  pair that separates "not there" from "could not ask". Shared by the state
  probes and the anchor readers, which is why it is neither's file.
- `scripts/maintenance/_gc_stale.py`: the state probes only now, plus the
  worktree-local ref channel wired into `unreachable_admin_commits`.
- `scripts/maintenance/_gc_reasons.py`: the warning and the keep reason name the
  admin directory rather than the reflog, because the reflog is one of two
  things in there that can be the last name for a commit.
- `scripts/maintenance/gc_worktrees.py` and `worktree_report.py`: the renamed
  probe reaches its callers.
- `tests/test_gc_anchor_readers.py`: new. Unit cases for the readers: the walk's
  refusals, the per-line parse, and the vanished file.
- `tests/test_gc_worktrees_real_git_anchors.py`: new. Both channels proved
  against real git, each with a control showing the commit dies once the channel
  is gone.
- `tests/test_gc_stale_probes.py`: trimmed by the move above, back under the cap,
  and carrying the worktree-local ref probe cases.
- `tests/test_gc_worktrees_real_git_stale.py`, `tests/test_gc_worktrees_stale_apply.py`,
  `tests/test_gc_worktrees_stale_warnings.py`: follow the rename from reflog to
  admin directory.
- `.agents/sessions/2026-08-05-session-10005.json`: the session log.
- `.serena/memories/`: index rows only. The two memories written during this
  work, the worktree-local ref blindness and the review subagent that cleaned
  the worktree it was given, are neither about the collector nor dependent on
  it, so they ship on `docs/memory-squash-severs-stack` off main instead. That
  also takes this branch from 51 files to 49, under the scope limit that the
  stacked base pushes it against.

## Type of Change

- [ ] Documentation update
- [ ] Refactoring (no functional changes)

## Reviewer Expectations

**Review kind / scrutiny / urgency**: full, high scrutiny, no rush.

**Focus areas**: the marker list in `_gc_stale.py`. Eight markers is a claim
about what git writes, and a git state that holds unreachable work and writes
none of them is a defect. If you can build one, that is the finding.

Second focus: the three guards in the apply module and whether any is redundant.
They look overlapping and are not. The HEAD check catches a commit that moved
HEAD. The reflog re-probe catches one that did not. The operation probe catches
work that produced neither. A reviewer who deletes any of them should be able to
name which real-git test goes red, because each was deleted during development
and each broke exactly one class of test.

Third focus: the recovery strings in the reasons module are pasted into a shell
by a human. Read them as shell, not as prose.

## Testing

- [x] Tests added/updated

Every fix here is backed by a test against a real git repository, not a mock.
Eight times during this work the mock suite passed while real git caught a real
bug, so the real-git suites are where the evidence lives.

Each guard was revert-proofed: the guard was deleted or weakened, the suite was
run, and the failures were counted. A guard whose removal breaks nothing is not
a guard. Every commit message here records its own count.

```
uv run pytest tests/ -q
23857 passed, 33 skipped, 2 warnings in 654.06s
```

Ratchets, both green:

```
uv run python scripts/ci/taste_count_ratchet.py
taste count ratchet: OK (count == baseline 583).

uv run python scripts/ci/memory_index_count_ratchet.py
memory index count ratchet: OK. 424 violations <= baseline 425 (-1 slack).
```

### Security Review

### Other Agent Reviews

Adversarial review rounds across four model families. Every finding was
reproduced against real git before any code changed, and three were rejected on
evidence. A late round found defect 5, which every round before it missed, and
that is the reason the reviews kept going. The last round found defect 9 in the
fix for defect 8, so the reviews were still paying at the end.

One finding was a hole in an earlier fix, and the test covering it passed
because it substituted pytest's `tmp_path`, a directory the fixture creates, for
the placeholder the command was supposed to create. That is now a rule in
`.claude/rules/testing.md`, and the shared extractor asserts the substituted
path does not exist.

One further defect came out of reviewing the fixes rather than from a reviewer:
the apply window between the recheck and a candidate's own removal was open to
the same suspended merge.

## Author Pre-flight

- [x] Builds locally (or N/A)


rjmurillo:
owner: local:pr-autofix
session: session-4728-20260807T072508Z
acquired_at: 2026-08-07T14:27:59Z
expires_at: 2026-08-07T14:42:59Z
base_sha: 6224ee932153aedac678f1096d6b83489bfefd52

rjmurillo:
owner: none
session: session-4728-20260807T072508Z
acquired_at: 2026-08-07T14:18:22Z
expires_at: 2026-08-07T14:18:22Z

Issue #4607, opened 2026-08-05
ci: BOT_PAT authenticates as the human account, so CI shares one rate-limit budget with every agent session


rjmurillo:
## Summary

The `BOT_PAT` secret authenticates as the human account `rjmurillo` (user ID
`6811113`), not the `rjmurillo-bot` service account (user ID `250269933`) that
ADR-026 Decision 5 specifies.

CI and every interactive agent session therefore share one REST budget and one
GraphQL budget. An agent polling `gh` can starve CI and turn unrelated pull
requests red.

This is a settings change. No code changes.

## Evidence

Account IDs, read live:

```
$ gh api user --jq '"login=\(.login) id=\(.id)"'
login=rjmurillo id=6811113

$ gh api users/rjmurillo-bot --jq '"login=\(.login) id=\(.id)"'
login=rjmurillo-bot id=250269933
```

The AI review context step authenticates with `bot-pat` only.
`.github/actions/ai-review/action.yml` lines 183 to 190:

```yaml
    - name: Build context
      id: context
      shell: bash
      env:
        GH_TOKEN: ${{ inputs.bot-pat }}
        ...
      run: python3 "$GITHUB_WORKSPACE/scripts/ci/build_ai_review_context.py"
```

All 22 caller sites pass `secrets.BOT_PAT` into that input.

That step failed on PR #4567, run `30929337949`:

```
##[error]Failed to fetch PR diff for #4567 ... HTTP 403: API rate limit
exceeded for user ID 6811113
```

The step's only credential is `BOT_PAT`. The refusal names `6811113`. The bot
is `250269933`. So `BOT_PAT` does not hold the bot's token.

## Impact

1. **CI has no budget of its own.** Interactive agent polling drains the same
   buckets CI uses. Measured on PR #4567 and PR #4471: all ten review agents
   failed at the context step, and the required `Aggregate Results` check went
   red while every underlying job reported success.
2. **The audit trail ADR-026 asked for does not exist.** Decision 5 states the
   goal as "clearer audit trail (bot actions vs. GitHub Actions runner)".
   Automated actions are attributed to the human account instead.

## Fix

Re-issue `BOT_PAT` from the `rjmurillo-bot` account and update the repository
secret. Then confirm:

```bash
gh api user --jq .id     # run with the new token, expect 250269933
```

After that, CI gets its own 5,000 per hour REST budget and its own GraphQL
budget. No amount of interactive agent polling can starve it.

## Relationship to #4547

[...]

PR #4731, opened 2026-08-07
fix: ratchet discriminator and triage auth

### LINKED ISSUE #4087: Skill-shape discriminator is change-triggered, so pre-existing candidates stay invisible
## Problem

`check_agent_skill_discriminator.py` only scores agents whose files appear in
the PR's changed-file list. An agent that is already skill-shaped on `main`
is therefore invisible until an unrelated edit happens to touch it.

## Evidence

Measured against an unmodified `origin/main` worktree, with no PR changes applied:

```
$ git worktree add --detach /tmp/discm origin/main
$ cd /tmp/discm
$ uv run python scripts/validation/check_agent_skill_discriminator.py \
    --repo-root . \
    --changed-files .claude/agents/architect.md .claude/agents/qa.md

  [CANDIDATE] architect (score 3/3: c1=Y c2=Y c3=Y, pipelines=1, isolation_required=no)
  [CANDIDATE] qa (score 2/3: c1=Y c2=n c3=Y, pipelines=1, isolation_required=no)
```

Both are at or above the failing threshold (score >= 2) on `main` right now.
The gate has never fired on them because nothing triggered a scoring run.

## Impact

Two failure modes, both degrading the gate's signal:

1. **Latent debt stays invisible.** Skill-shaped agents accumulate on `main`
   with no ticket, no owner, and no visible score.
2. **The override becomes the default path.** When the gate fires on a PR
   that did not cause the condition, the correct action for that PR is the
   `[skill-discriminator: ...]` token. Every such use trains contributors to
   reach for the escape hatch first, and the gate decays into a formality.

Observed on PR #4067 (replaces prohibited em dashes with commas) and PR #4063
(fixes Pester examples). Neither change can move c1, c2, or c3, yet the gate
fired on both.

Correction to an earlier version of this issue: it said both PRs were
"blocked". They were not. This check is not in the repository's required
status check set, so a red result is advisory and does not gate merge. That
makes the decay argument sharper rather than weaker. An advisory check that
fires on changes which cannot cause the condition it detects is pure noise on
the PR page, and noise is exactly what trains contributors to reach for the
override without reading it.

## Suggested direction

Not prescribing an implementation. Options worth weighing:

- A scheduled full-corpus scan that scores every agent and reports drift,
  separate from the per-PR gate.
- A checked-in baseline of known candidates, so the per-PR gate fails only on
  a score *increase* relative to the baseline.
- Grandfathering via `isolation_required: true` on the agents that are
  deliberately staying agent-shaped, which converts latent debt into an
  explicit, reviewable declaration.

The baseline option preserves the gate's purpose (stop new debt) while
removing the false positives that currently push contributors to the override.

### LINKED ISSUE #4210: backlog-triage workflow is broken: bare python3 invocation of a script with an unguarded scripts.* import
## Summary

`.github/workflows/backlog-triage.yml` is broken. It runs `python3 scripts/issue_triage.py` by path after `actions/setup-python` with no project install, and that script imports `scripts.github_core.validation` with no `sys.path` bootstrap. The import cannot resolve, so the scheduled triage job fails before it does any work.

Thirteen more scripts carry the same unguarded import. This is the third time this class has landed (#3657, #3711), and both prior fixes were case by case, so nothing stops the fourth.

## Reproduction

The exact command the workflow runs, on a clean checkout with a system interpreter:

```
$ /usr/bin/python3.12 scripts/issue_triage.py
    from scripts.github_core.validation import is_github_name_valid
ModuleNotFoundError: No module named 'scripts'
$ echo $?
1
```

The workflow has no `pip install`, no `uv sync`, and no `uv pip` step:

```
$ grep -n 'pip install\|uv sync\|uv pip' .github/workflows/backlog-triage.yml
(no output)
```

Running the same file under `uv run` succeeds, which is why local runs and lefthook never caught it. `uv` puts the project root on `sys.path` through `.venv/lib/python3.14/site-packages/__editable__.ai_agents-0.1.0.pth`. A runner that injects the path cannot validate an import that CI executes bare.

## Scope

Directly executable scripts that import `scripts.*` and lack the bootstrap. Each was confirmed to raise `ModuleNotFoundError: No module named 'scripts'` under `/usr/bin/python3.12`:

```
scripts/homework_scanner.py
scripts/invoke_batch_pr_review.py
scripts/invoke_pr_maintenance.py
scripts/invoke_session_start_gate.py
scripts/issue_triage.py            <- live, invoked bare by backlog-triage.yml
scripts/memory/detect_stale.py
scripts/memory/memory_health.py
scripts/memory_sync/cli.py
scripts/memory_sync/__main__.py
scripts/new_validated_pr.py
scripts/security/invoke_precommit_security.py
scripts/security/invoke_security_retrospective.py
scripts/security/run_semgrep.py
scripts/sync_mcp_config.py
```

Only `scripts/issue_triage.py` is confirmed live today. The other thirteen are latent, reachable the moment any caller switches from `uv run` or `python -m` to a bare path, which is exactly what happened here.

Checked and found healthy: `.github/scripts/invoke_pr_maintenance.py` already carries the bootstrap.

## Prior art, and why a gate is warranted now

- #3657 fixed `scripts/validation/traceability.py` after it reported a false FAIL to every contributor.
- #3711 catalogued five sibling validators with the identical unguarded import and stated the defect was "latent, not live" because "every one of them is reached only through `uv run` (lefthook) or `python3 -m` (CI)."

That premise no longer holds. `backlog-triage.yml` reaches one by path. And the class recurred again in PR #4206, where I added a `scripts.validation` import to `check_skill_md_exec_portability.py`, passed every local gate under `uv run`, and turned `Validate Vendor Portability` red. Three occurrences, two prior fixes, no detector.

## Proposed fix

1. Add the bootstrap to `scripts/issue_triage.py` to unbreak the workflow.
2. Add a validator that fails when a module with an `if __name__ == "__main__"` block imports `scripts.*` at module level without `sys.path` containing the repo root. The check is a static scan and needs no execution.
3. Alternatively, or additionally, run the affected entry points under a project free interpreter in CI so the gate matches how they are invoked.

Item 2 is the one that stops recurrence. Items 1 and 3 fix the current instance.

## Acceptance

- `python3 scripts/issue_triage.py --help` exits 0 under a system interpreter with no project install.
- A new test adds an unguarded `scripts.*` import to a fixture with a `__main__` block and the validator fails; removing the import passes.
- The scheduled `backlog-triage` workflow completes its discover step.

rjmurillo:
## Summary

Fixes #4087
Fixes #4210
Refs #4607

## Issue #4087

Reproduced: changed-file mode returned 0 with no changed agents, while full-corpus mode returned 1 on 11 existing candidates. Pre-existing candidates were either hidden in PR mode or failed as unresolved debt in full-corpus mode.

Changed: added `scripts/validation/agent_skill_discriminator_baseline.json`, a full-corpus baseline ratchet, helper modules, and workflow path coverage. Current baseline records 11 candidates. `check_agent_skill_discriminator.py --all` now prints the existing set and exits 0 unless a new candidate appears.

Both-directions control: with the production script temporarily restored from `origin/main`, `test_full_corpus_baseline_allows_existing_candidates` failed with `unrecognized arguments: --baseline`. Restored the fix and the discriminator tests passed.

## Issue #4210

Reproduced: `python3 -I -S scripts/issue_triage.py --help` failed with `ModuleNotFoundError: No module named 'scripts'` at the `scripts.github_core.validation` import.

Changed: backlog triage now invokes `python -m scripts.issue_triage`, and `scripts/issue_triage.py` bootstraps the repository root for direct script execution before retrying the shared import.

Both-directions control: with `scripts/issue_triage.py` and `.github/workflows/backlog-triage.yml` temporarily restored from `origin/main`, both new tests failed. The direct-script test failed with `ModuleNotFoundError`, and the workflow test saw `uv run --frozen python scripts/issue_triage.py`. Restored the fix and the targeted tests passed.

## Issue #4607

Reproduced: `gh api user --jq '.login + " " + (.id|tostring)'` showed `rjmurillo 6811113`, not `rjmurillo-bot 250269933`. Repository code cannot rotate `BOT_PAT` or prove the secret identity changed.

Changed: documented the operator check in `docs/copilot-cli-setup.md`, including the required bot identity and the fact that `gh api rate_limit` does not expose secondary rate limits. This is repository-side guidance only. Issue #4607 must remain open until an operator mints `BOT_PAT` from `rjmurillo-bot` and verifies account ID `250269933`.

Both-directions control: not applicable to the secret rotation. No repository checkout can perform the operator action.

## Validation

- `uv run --frozen python -m pytest tests/ -q -p no:randomly --timeout=300`: passed, 23592 passed, 33 skipped, 50 deselected, 2 warnings.
- `uv run --frozen python scripts/validation/git_hook_policy.py mypy $FILES`: passed.
- `uv run --frozen python scripts/ci/merge_tree_ratchet_check.py`: passed.
- `uv run --frozen python scripts/validation/pre_pr.py`: passed.
- Pre-push hook: passed, including python-tests, pre_pr, merge-tree-ratchet, retrospective-policy, and workflow-local-run.

GitHub Actions have not been observed yet after PR creation.


rjmurillo:
Closing this combined PR before taking issue #4210. It mixes #4087, #4210, and #4607, conflicts with current main, and has five unresolved review threads in the #4087 implementation. Issue #4607 still requires operator action. I will replace only the #4210 portion with a focused PR. Issues #4087 and #4607 remain open.

PR #4734, opened 2026-08-07
docs(memory): record that a __file__-rooted script validates its own worktree

rjmurillo:
## Summary

A repo script invoked by absolute path from a different worktree validates the
**script's** worktree, not the one you are standing in. It then reports a clean
PASS for a tree you never checked.

This adds one Serena memory recording the trap, what was measured, and how to
check a candidate invocation before you run it.

## Why this matters

The failure is silent and it fails in the safe-looking direction. You get exit 0
and "counts are current" while the tree you actually care about is stale. Nothing
in the output names the tree that was read.

## Measured proof, two directions

Planted a deliberately stale token count (1132 changed to 9999) in a worktree's
`memory-index.md`, then ran the same validator two ways against the same file and
the same working directory.

| Invocation | Reads | Verdict | Exit |
| --- | --- | --- | --- |
| Main clone's copy, by absolute path | the main clone | "token counts are current" | 0 |
| The worktree's own copy | the worktree | "stale: recorded 9999, actual 1132" | 1 |

`uv` does not save you. It printed `Built ai-agents @ file:///.../wt-scriptroot`,
correctly resolving the project from the working directory, while the script still
read the other tree.

## The actual invariant

Not "use a relative path." An absolute path inside the intended worktree is safe,
and a relative path resolving through a symlink into another worktree is not. The
invariant is that the **resolved script path must be contained by the intended
worktree**.

The memory ships a detector that takes the candidate path as an argument and tests
containment in Python. It was verified in three directions: in-tree returns
"contained by" exit 0, the other tree returns "WRONG TREE" exit 1, and a set
`GIT_DIR` returns a guard message exit 1.

## Scope, measured rather than asserted

| Measure | Count |
| --- | --- |
| Files under `scripts/` matching `Path(__file__).resolve().parent` | 119 |
| Python files under `scripts/` recursively | 431 |

That is about 28 percent, so the pattern is common rather than universal.
`restructure_memories.py:394` and `check_unreachable_code.py:119` correctly use
the working directory instead.

## The pre-commit hook is not fooled

`lefthook.yml` invokes the updater by relative path from inside the committed tree,
so the hook itself cannot hit this. It also repairs before it checks:
`memory-token-update` (line 260) and `stage-memory-index` (line 268) both run
before `memory-token-counts --check` (line 291). Rejection needs a merge commit or
`SKIP_AUTOFIX=1`, which matches `.claude/rules/knowledge-persistence.md` MUST-6.

## Adversarial review

Reviewed by `gpt-5.6-sol` (different model family from the author). Verdict came
back **INCORRECT** with three High findings. Every finding was independently
re-verified against source before being accepted, per the Sol integrity rule, and
all of them held:

1. The original draft claimed the hook rejects stale counts. It repairs them.
2. The original draft claimed an unconditional mutation. The write at
   `update_memory_index_tokens.py:140` is conditional on a diff.
3. The original detector rebuilt its input as `$TREE/$SCRIPT`, so it was
   tautological and could never fire.

The memory was rewritten around all three and the detector replaced.

## Testing

- Two-direction control on the trap, above.
- Three-direction control on the corrected detector.
- `memory_index.py --ci`: 326 files indexed, 0 missing, 0 keyword issues.
- Token counts recomputed with the updater, never hand-edited.
- Style gate: zero em/en dashes, zero banned words, every bash fence passes
  `bash -n`, zero angle-bracket placeholders, both `Related` targets resolve.

Refs #4415

### DIFF
--- a/.serena/memories/git/git-a-script-run-by-absolute-path-validates-its-own-worktree.md
+++ b/.serena/memories/git/git-a-script-run-by-absolute-path-validates-its-own-worktree.md
@@ -1,0 +1,175 @@
+# A script that derives its root from `__file__` reads its own worktree, whatever directory you are standing in
+
+## The trap
+
+Many scripts in this repository find the repository root from their own
+location on disk rather than from the current directory.
+`scripts/update_memory_index_tokens.py:171` is the shape:
+
+```python
+repo_root = Path(__file__).resolve().parent.parent
+```
+
+With `git worktree`, that is a silent correctness hole. Standing in worktree B
+and invoking worktree A's copy of a validator makes the validator read, check,
+and report on **A's** files. It exits 0 because A is fine. You conclude B is
+fine. B was never examined.
+
+The failure is silent and it inverts the signal: the more worktrees you run, the
+more confident and the more wrong the PASS becomes.
+
+## Measured proof
+
+Observed 2026-08-05. Same working directory, only the script copy differs.
+
+Setup: in worktree `wt-scriptroot`, edit one recorded token count in
+`.serena/memories/memory-index.md` from `1132` to `9999`. That worktree's index
+is now provably stale. The other clone's index is untouched and current.
+
+Direction A, the other tree's script by absolute path:
+
+```
+$ cd ~/src/scratch/wt-scriptroot
+$ uv run --frozen python ~/src/GitHub/rjmurillo/ai-agents/scripts/update_memory_index_tokens.py --check
+Creating virtual environment at: .venv
+   Building ai-agents @ file:///home/richard/src/scratch/wt-scriptroot
+memory-index.md token counts are current
+exit=0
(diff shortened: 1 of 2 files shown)


rjmurillo:
owner: local:pr-autofix
session: autofix-pr-4734-80d53b29-f88a-499e-93df-f187be65ce89
acquired_at: 2026-08-07T14:28:29Z
expires_at: 2026-08-07T14:43:29Z
base_sha: fff0cbe8480330af767d5cfa2970d0067458b4e8

rjmurillo:
owner: none
session: autofix-pr-4734-80d53b29-f88a-499e-93df-f187be65ce89
acquired_at: 2026-08-07T14:15:41Z
expires_at: 2026-08-07T14:15:41Z

Issue #4725, opened 2026-08-07
semgrep python36 compatibility rules fire on a repository with a 3.14 floor, keeping the scan permanently red


rjmurillo:
## Summary

`semgrep-cloud-platform/scan` reports two blocking findings on every pull request that touches `scripts/validation/run_workflow_local_test.py`. Both are false positives, and both are unfixable as stated, because the rules assert a floor this repository is nine minor versions above.

```
python.lang.compatibility.python36.python36-compatibility-Popen1
  the `errors` argument to Popen is only available on Python 3.6+
python.lang.compatibility.python36.python36-compatibility-Popen2
  the `encoding` argument to Popen is only available on Python 3.6+
```

Both point at the same call:

```python
proc = subprocess.Popen(
    cmd,
    cwd=str(cwd) if cwd else None,
    env=env,
    stdout=subprocess.PIPE,
    stderr=subprocess.PIPE,
    encoding="utf-8",
    errors="replace",
    start_new_session=_supports_pgroup,
)
```

The rules fire on `encoding=` and `errors=` and warn they need Python 3.6 or newer.

```
$ grep -E '^requires-python' pyproject.toml
requires-python = ">=3.14"
```

The floor is 3.14. Arguments available since 3.6 are unconditionally available. There is nothing to fix in the code; the finding is the rule being wrong about this repository.

## Why this is worth an issue rather than a shrug

The lines are not new. They arrived with #4374 (kill process group on timeout to prevent gh-act port leak), so any pull request that touches this file inherits a red `semgrep-cloud-platform/scan`.

`semgrep-cloud-platform/scan` is not in the branch protection ruleset, so it does not block a merge:

```
$ gh api repos/rjmurillo/ai-agents/rules/branches/main --jq '[.[] | select(.type=="required_status_checks") | .parameters.required_status_checks[].context] | map(select(test("semgrep";"i"))) | length'
0
```

That is exactly what makes it worth filing. A check that is always red and never blocking trains readers to skip it, and the next time semgrep reports something real on this file it will be read as more of the same. The cost of a permanently red signal is paid later, by whoever ignores a true positive because the check is always red.

## Options

1. Disable the `python.lang.compatibility.python36.*` rule family in the semgrep configuration for this repository. It is the correct scope: the whole family is inapplicable at a 3.14 floor, not just these two rules on this one call. This is preferred.
[...]

Issue #4672, opened 2026-08-05
Fresh project-toolkit install blocks Copilot CLI sessions on Windows Dev Box


snehalchaudhari98:
## Environment

- OS: Windows Dev Box
- GitHub Copilot CLI: `1.0.78-2`
- Plugin: `project-toolkit@ai-agents`
- Installation: Fresh marketplace install on August 5, 2026

## Installation

```text
/plugin marketplace add rjmurillo/ai-agents
/plugin install project-toolkit@ai-agents
```

## Problem
Immediately after installing the plugin, Copilot CLI chat sessions began failing with:
```
Denied by preToolUse hook from "project-toolkit@ai-agents"
```

The failure affected all newly started chat sessions, preventing normal tool use.
## Workaround
Removing or disabling `project-toolkit@ai-agents` in Copilot's `settings.json` restored normal operation. But by disabling it I am not able to use this extension.

## Expected behavior
Installing the plugin should not block Copilot CLI tool calls or make new chat sessions unusable.

rjmurillo:
## Velocity Accelerator: 1 Opportunities Detected

### Triage issue #4672: Fresh project-toolkit install blocks Copilot CLI sessions on
- **Type**: `complexity_triage`
- **Priority**: medium
- **Suggested Agent**: orchestrator
- Issue labeled with medium complexity.
Suggested agents: none detected

PR #4744, opened 2026-08-07
docs(rules): verify a red remote check with equivalent evidence

rjmurillo:
## Summary

One always-on rule, plus the repository-specific facts behind it moved somewhere retrievable.

On PR #4701 a `semgrep-cloud-platform/scan` check was red. I ran semgrep locally, got zero findings, and published a comment saying the red was unrelated pre-existing noise that could be disregarded. That was wrong. The local run was `semgrep --config=auto`; CI runs the Semgrep Cloud App with a server-side ruleset, and this repository ships no semgrep config for a CLI to read, so my clean result was evidence about a different check.

Both findings were real. One sat beside a genuine defect: a URL parse deriving `github.com/owner` from a remote naming no repository, which would have queried the wrong repository.

## The rule

```
4. Verify a red remote check with equivalent evidence, or report it
   unreproduced. MUST NOT claim a red remote check is cleared by a local run
   unless the checker, ruleset, flags, and version match. State the command and
   why it is the same check; if you cannot, say the check is red and you could
   not reproduce it.
```

The semgrep specifics went to `GOTCHAS.md` rather than into the rule: which ruleset CI actually runs, that findings arrive as review comments while the check run itself carries no annotations or output text, that `p/security-audit` gets closer than `auto` but is still a guess, and that the suppression hook forbids the obvious escape. Those are facts you look up while staring at a red scan, not judgment that must bind every turn.

## What review changed, which is most of the value here

I first wrote three rules. Adversarial review recommended cutting two. It was right on both counts, and one of the cuts matters more than the rule that survived.

The rule I cut said: when one pull request is red and its siblings are green, the cause is in your diff. Two problems. It is a debugging heuristic any competent reader already applies, so it fails the admission test for always-on bytes. And its cited evidence was false. I wrote that a sibling modified "the same file at the same lines" and passed. The hunk headers say otherwise:

```
PR 4723: @@ -542  @@ -904  @@ -928
PR 4701: @@ -721
```

Same file, different hunks. I published an unverified claim inside the rule written to stop me publishing unverified claims. The reviewer caught it by checking rather than reading, which is the lesson the surviving rule states.

The third rule, about paginating list endpoints and reducing duplicate check names by best outcome, is a fact you look up while doing the thing. It lives in the Serena memory `git/github-api-lists-truncate-at-per-page-and-dedup-must-take-best` instead.

## Budget

The instruction budget caught the first draft at 99,128 bytes against a 99,000 ceiling. Raising the ceiling was the wrong move and the gate was right: a budget that rises to meet its contents measures nothing.

| | Always-on bytes | Growth |
|---|---|---|
| Before | 71,033 | |
| Three rules | 72,300 | +1,267 |
| After review, one rule | 71,484 | +451 |

1,765 bytes of headroom remain. Three documents pin these figures and a test asserts each against a live measurement, so all three moved twice, once per rule change.

Review also caught that the new figures still credited commit `a4129aa44`, which measured the previous corpus. Attributing a new measurement to a commit that produced a different number is the same defect as citing a check you did not run, so the attribution now names the change actually measured.

## Testing

23,802 pass, 33 skipped. All three ratchets at baseline. `pre_pr.py` passes. markdownlint clean on both changed documents. No em or en dashes, no banned vocabulary.

One push-time failure worth recording: the episode store ratchet blocked on `commits==0 but files_changed>0`. The cause was an empty `endingCommit`, which the session validator had already reported as a warning rather than an error. The extractor derives its commit range from that field, so the warning at write time becomes a blocked push later.

Refs #4672
Refs #4725


rjmurillo:
All six threads are addressed at `ec9ac7378`. Verified each against that commit rather than assuming the push carried them.

**ADR and consensus.** Correct, and the process failure was mine: `governance.md` MUST 2 requires an ADR for a new rule and MUST 3 requires consensus for one governing every role, and I had done neither. `ADR-093` now exists, with the consensus recorded as a debate log at `.agents/critique/ADR-093-debate-log.md`.

The review it records is worth more than the paperwork. The architect returned accept with a wording change; the critic returned revise, arguing the rule would not fire for an agent who believed the tools already matched. That is a fair reading of the original wording, and two competent reviewers reading a rule oppositely means the rule was underspecified. It now requires the equivalence to be **demonstrably identical** rather than merely matching: belief satisfies "match", it cannot satisfy "demonstrably".

The critic also asked whether a CI-side fix had been evaluated. It had not, which was a real gap, and checking changed the advice. `semgrep ci --help` states that when logged in it runs the rules configured on Semgrep App, so `SEMGREP_APP_TOKEN=<token> semgrep ci` is an exact reproduction. The check is locally reproducible after all, and the GOTCHAS entry claimed the opposite.

**Percentage.** Correct. 14,152 / 71,640 is 19.76%, and the text said 20.1%. Fixed to 19.8%.

**"Necessary condition" contradiction.** Correct, and the sharper phrasing is yours: a pass from a different ruleset is neither necessary nor sufficient. The text said "treat a local pass as a necessary condition" two paragraphs after saying a clean local result says nothing. Rewritten, and it now names the one command that does reproduce the check.

**The #4725 citation.** Correct. A single trailing `Refs #4725` made the python36 false positives look like evidence for "both findings were real". Those are two incidents with opposite conclusions, and the text now says so explicitly in both directions: do not read the taint conclusion as covering the compatibility rules, or #4725 as licence to dismiss a taint finding.

**`handoffRead` evidence.** Correct, and I did not back-fill it. I read `.agents/HANDOFF.md` during this remediation, not at session start, so the log now records that order rather than restating the claim to match the checkbox. Writing "read at session start" would have been the fabrication the surviving rule exists to prevent.

**Stale attribution.** Correct, and this one is the sharpest catch: the diff repeated the exact error the pull request describes, attributing a new measurement to a commit that produced a different number. `b2729ee54` no longer appears; the line names the change actually measured.

The figures moved twice more after these fixes, once when the third rule was cut and once for the demonstrably-identical wording, so the current numbers are 71,505 always-on, 97,256 on a Python edit, and 71,640 at source. Always-on growth from this branch is 472 bytes.

PR #5025, opened 2026-08-15
fix(hooks): exempt matcher-shimmed hooks from atomic-commit count

### LINKED ISSUE #4857: atomic-commit counter treats generated Copilot hook shims as authored files
## Problem

`check_atomic_commit` counts a generated Copilot hook shim as an authored
file, so every hook change pays a phantom slot against the five-file limit.

`_GENERATED_MIRRORS` in `scripts/validation/git_hook_policy.py:490-500`
declares `src/copilot-cli/hooks/` a mirror of `.claude/hooks/`, and
`_is_generated` resolves a mirror path by rewriting the prefix and checking
that the canonical source is staged. That works for a verbatim companion,
whose name is unchanged:

```
src/copilot-cli/hooks/PreToolUse/_push_pr_guard_lex.py
  -> .claude/hooks/PreToolUse/_push_pr_guard_lex.py   staged, exempt
```

It does not work for a matcher shim, because `generate_hooks.py` appends a
matcher suffix to the name:

```
src/copilot-cli/hooks/PreToolUse/invoke_push_pr_script_identity_guard__Bash_f620ca.py
  -> .claude/hooks/PreToolUse/invoke_push_pr_script_identity_guard__Bash_f620ca.py
     does not exist, so the file is counted as authored
```

## Evidence

Measured on branch `fix/4764-postmerge-regressions`. Staging one canonical
guard change plus its regenerated tree:

```
INFO: 1 generated file(s) excluded from atomic-commit count:
  src/copilot-cli/hooks/PreToolUse/_push_pr_guard_tables.py
ERROR: commit touches 6 authored files (limit is 5).
Authored files staged:
  .claude/hooks/PreToolUse/_push_pr_guard_tables.py
  .claude/hooks/PreToolUse/invoke_push_pr_script_identity_guard.py
  build/scripts/generate_hooks_events.py
  src/copilot-cli/hooks/PreToolUse/_push_pr_guard_git_tables.py
  src/copilot-cli/hooks/PreToolUse/invoke_push_pr_script_identity_guard__Bash_f620ca.py
  tests/hooks/test_push_pr_guard_bundle.py
```

The companion is exempt on the same run, so the rule works; only the
suffixed name misses. Issue #4671 added `_GENERATED_MIRRORS` for exactly this
class of cost, and the shim was left behind because its name is rewritten.

## Effect

Any commit that touches a matcher-shimmed hook has four usable slots, not
five, and the counter's message names a file no human wrote. Splitting the
push-pr guard for the 500-line ceiling cost several extra commits purely to
route around it.

## Suggested fix

Teach `_mirror_source` the matcher-suffix grammar. `generate_hooks.py`
exposes `_matcher_suffix`, and the shim carries an `AUTO-GENERATED MATCHER
SHIM` header plus a `# Matcher:` line, so either the name grammar or the
header can resolve the canonical owner. Strip the suffix before the
staged-source check, and keep the existing behavior when no suffix is
present.

Refs #4764

rjmurillo:
## Summary

The atomic-commit counter treats generated Copilot hook matcher shims as
authored files because `_mirror_source` does not strip the matcher suffix
before looking up the canonical source.

`generate_hooks_emit` appends `__{sanitized}_{6hex}` to shim filenames
when a hook is registered with a matcher pattern. The mirror lookup maps
`src/copilot-cli/hooks/PreToolUse/invoke_guard__Bash_f620ca.py` to
`.claude/hooks/PreToolUse/invoke_guard__Bash_f620ca.py`, which does not
exist. The canonical source is `invoke_guard.py` (without suffix).

## Changes

**`scripts/validation/git_hook_policy.py`**: Add `_HOOK_MATCHER_SUFFIX_RE`
regex matching `__(?:[A-Za-z0-9_]+_)?[0-9a-f]{6}` before `.py`. Strip it
in `_mirror_source` only for paths resolving to `.claude/hooks/`. Verbatim
companions and non-hook mirrors are unaffected.

**`tests/validation/test_atomic_commit_mirror_exemption.py`**: Add
`TestMatcherShimSuffixStripping` (7 tests): positive (sanitized+hex,
bare hex), negative (companion unchanged, lib not stripped, skills not
stripped), integration (exempt when staged, not exempt when missing).

## Acceptance

Before:
```
src/copilot-cli/hooks/PreToolUse/invoke_push_pr_script_identity_guard__Bash_f620ca.py
  -> .claude/hooks/PreToolUse/invoke_push_pr_script_identity_guard__Bash_f620ca.py
     NOT FOUND -> counted as authored (phantom slot)
```

After:
```
src/copilot-cli/hooks/PreToolUse/invoke_push_pr_script_identity_guard__Bash_f620ca.py
  -> .claude/hooks/PreToolUse/invoke_push_pr_script_identity_guard.py
     FOUND (staged) -> exempt from count
```

Fixes #4857


rjmurillo:
## Response to review thread (scripts/validation/git_hook_policy.py)

Good catch. The original regex matched greedily from the first `__`, so a source named `foo__bar.py` generating shim `foo__bar__Bash_f620ca.py` would incorrectly strip to `foo.py`.

Fixed in commit f2afec8: the regex now uses `(?!__)` negative lookahead to ensure it only matches at the last `__` boundary. This works because `_matcher_suffix` collapses non-alnum runs to single `_`, so the suffix itself never contains `__`.

Added regression test `test_source_with_double_underscore_preserves_stem`.

rjmurillo:
owner: local:pr-autofix
session: session-14708-pr5025
acquired_at: 2026-08-15T09:43:48Z
expires_at: 2026-08-15T09:58:48Z
base_sha: 08452097c153bf8e36e59e5996b66c13c70b723e
claim_id: 15a57885671547709b70c3d484e12d62

rjmurillo:
## Response to thread (regex precision)

The pattern is intentionally slightly broader than the exact `_matcher_suffix` grammar. The broader match is safe because of the second-stage check: `_is_generated` still requires `_is_staged_regular_file(root, source)` to return True. If the resolved canonical source is not staged, the file is NOT exempt.

So even if the regex matches a filename that is not actually a matcher shim, the exemption only fires if the resolved source happens to be staged - which means that source IS the canonical file this output was generated from.

The cost of a false-positive match on the regex alone (with no staged source) is zero: the file stays counted as authored. The cost of a false-positive match WITH a staged source is also zero: it means the file genuinely is a mirror of that source.

Tightening further (e.g. requiring `_` before the digest) is possible but adds complexity with no safety gain.

Issue #3343, opened 2026-07-25
branch-context-policy permanently wedges any branch that merges main


rjmurillo:
Merging `main` into a feature branch imports the previously-merged branch's session log. Once that merge is committed, `branch-context-policy` blocks every subsequent push on the feature branch, permanently, even though the branch has its own valid session log.

## Reproduction

Observed live on branch `fix/3328-foreign-prose-sha`, which carries its own session log for that branch:

```
ERROR: branch context mismatch: current=fix/3328-foreign-prose-sha,
       session=fix/2993-2967-wire-ratchets
       (log: 2026-07-25-session-2993-wire-orphaned-ratchets.json)
```

The named log belongs to a PR that already merged to `main`. It arrived in the branch through a routine `git merge origin/main` to clear a conflict. The branch has never been near that work.

Steps:

1. Land a PR that includes a session log.
2. On any open feature branch with its own session log, merge `main` to clear a conflict.
3. Commit the merge.
4. Push. Blocked, and stays blocked, because the imported log is newer by mtime.

## Root cause

`check_branch_context` in `scripts/validation/git_hook_policy.py` selects the session log by recency alone and never asks whether a log for the current branch exists. `_today_session_log` globs today and yesterday and returns the newest by mtime. A freshly merged file always wins that comparison.

The function already anticipates this hazard. Its own docstring says a merge is exempt because it "legitimately imports another branch's newer session log into the tree, which would otherwise read as a mismatch." The exemption is keyed on `_merge_in_progress`, which tests for `MERGE_HEAD`. That file disappears the moment the merge commit is created, so the exemption covers the few seconds during conflict resolution and expires exactly when the imported log becomes permanent.

## Why P1

There is no clean way out. The branch is correctly configured; the guard is reading the wrong file. The workarounds are all bad:

- `--no-verify` is prohibited.
- Touching the branch's own log to win the mtime race re-breaks on the next merge and is a lie about when the session happened.
- Deleting the imported log reverts a file that legitimately belongs to `main`.

Every contributor who resolves a conflict by merging `main` hits this.

## Fix

Prefer a recent session log whose `branch` field matches the current branch. Fall back to newest-wins only when no recent log names the current branch.

That keeps the original guard intact. Issue #682 wrote this to catch PR co-mingling: you are on branch X while the only session context is for branch Y. In that case no log names X and the fallback still blocks. It only stops firing when a log for the current branch is sitting right there, which is the case where the branch context is demonstrably correct.

[...]

PR #5226, opened 2026-08-21
fix(hooks): resolve merged-history upstream via a fallback ladder

### LINKED ISSUE #5220: check_branch_context hard-blocks every commit after a merge of main when origin/HEAD is unset
## Summary

On a clone with no `origin/HEAD` ref, merging `main` into a feature branch blocks every subsequent commit on that branch. The exemption written for exactly this case (issue #3343) cannot fire, and the error message names none of the three real causes.

Reproduced on `claude/adr-evaluation-tooling-6od8rd` on 2026-08-21.

## What happens

`git merge origin/main` imports the previously merged branch's session log into `.agents/sessions/`. That file is newer by mtime than anything the current branch owns, so `_today_session_log` picks it, `_session_branch` reads a branch this checkout has never been on, and `check_branch_context` blocks:

```
ERROR: branch context mismatch: current='claude/adr-evaluation-tooling-6od8rd',
session='claude/pr-automerge-goal-eu2soz'
(log: 2026-08-21-session-99926-a1b2c3d4e-pr-automerge-goal.json)
  Fix: switch to the expected branch, update the session log branch field,
       or create a new session log for the current branch.
```

Every one of those three suggested fixes is wrong here. The branch already owns a correct session log. The imported log is settled history, not a claim about current work.

## Why the exemption misses

`scripts/validation/git_hook_policy.py` has the #3343 exemption:

```python
if _session_log_for_branch(sessions_dir, current_branch) is not None and _is_merged_history(
    repo_root, session_log
):
    return 0
```

`_is_merged_history` resolves the upstream default branch first:

```python
head = _run_git(repo_root, ["rev-parse", "--abbrev-ref", "origin/HEAD"])
upstream = head.stdout.strip() if head.returncode == 0 else ""
if not upstream:
    return False
```

With no `origin/HEAD` ref, that command exits 128 and the function returns `False`. This is the documented fail-closed contract and the function is behaving as written. The problem is that the *precondition* it depends on is not guaranteed.

Measured on this clone before the fix:

```
$ git rev-parse --abbrev-ref origin/HEAD
fatal: ambiguous argument 'origin/HEAD': unknown revision or path not in the working tree
rc=128
```

After `git remote set-head origin --auto`:

```
current branch          : claude/adr-evaluation-tooling-6od8rd
winning log             : 2026-08-21-session-99926-a1b2c3d4e-pr-automerge-goal.json
branch owns a log       : .agents/sessions/2026-08-21-session-5189-...-adr-corpus-evaluation-and-tooling.json
winner is merged history: True
```

Exemption fires, commits proceed.

## Why this matters

`origin/HEAD` is not set by every clone path. `git clone` sets it; a fetch into an existing repo, a shallow or filtered clone, and several CI checkout actions do not. A fresh remote-execution container is one of the environments where it is absent, which is where this was hit.

The consequence is disproportionate to the cause. A developer or agent that merges main, which the repo actively wants (`MAIN_MERGE_BLOCK_THRESHOLD = 40` exists to reward it), finds every commit blocked, with a message pointing at three non-fixes. `.claude/rules/universal.md` MUST-NOT-2 forbids all six hook bypasses, and issue #3408 records that a previous version of this same gate drove people to `--no-verify` on every worktree commit. A gate that leaves no correct move is how that happens again.

## Proposed fix

Two parts, both small:

1. **Do not depend on a ref that may not exist.** `_is_merged_history` should fall back to `origin/main` (and then `main`) when `origin/HEAD` does not resolve, the same ladder `resolve_push_update` already uses at `git_hook_policy.py` around line 5604. That keeps the fail-closed contract for a genuinely unresolvable trunk while removing the false negative for the common one.

2. **Say which precondition failed.** When the mismatch blocks *and* the winning log is present on a candidate trunk but `origin/HEAD` did not resolve, the message should say so and name `git remote set-head origin --auto`. The current three suggestions are correct for the co-mingling case from issue #682 and wrong for this one, with nothing to tell them apart.

## Tests

The existing suite pins the fail-open-on-missing-git contract (`test_branch_context_fails_open_when_git_is_unavailable`). This needs the mirror: a case with `origin/HEAD` deleted, a merged-history log present, and the branch owning its own log, asserting the exemption still fires. That case fails today.

## Refs

- `scripts/validation/git_hook_policy.py`, `_is_merged_history` and `check_branch_context`
- Issue #3343 (the exemption this defeats)
- Issue #3408 (the same gate previously driving `--no-verify` use)
- Issue #682 (the co-mingling case the exemption must not weaken)
- `.claude/rules/universal.md` MUST-NOT-2

Found while merging `origin/main` into `claude/adr-evaluation-tooling-6od8rd` during the ADR corpus campaign (PR #5209). Worked around locally with `git remote set-head origin --auto`; no hook was bypassed.

rjmurillo:
### What

`_is_merged_history` in `scripts/validation/git_hook_policy.py` now resolves the upstream default branch through a two-rung fallback (`origin/HEAD`, then `origin/main`) instead of depending on `origin/HEAD` alone, via a shared `_resolve_upstream_default` resolver also used by the pre-existing `_read_upstream_default_blob` (previously a duplicate, looser implementation). `check_branch_context`'s block message now names `git remote set-head origin --auto` specifically when the branch owns a session log but no upstream ref resolves at all, instead of always printing the generic co-mingling remedies.

### Why

`origin/HEAD` is only set by `git clone`. A fetch into an already-initialised repo, a shallow or filtered clone, and several CI checkout actions leave it absent, which is exactly the shape a fresh remote-execution container has. When `origin/HEAD` doesn't resolve, `_is_merged_history` unconditionally returned `False`, so the issue #3343 merged-history exemption in `check_branch_context` could never fire, and a developer who merged `main` and owned their own session log still got permanently blocked on every subsequent commit.

Reproduced on this exact checkout before the fix (`origin/HEAD` unresolved, `origin/main` fine), with the same log names issue #5220 reports.

## Specification References

| **Issue** | Fixes #5220 | check_branch_context hard-blocks every commit after a merge of main when origin/HEAD is unset |

## Changes

- `scripts/validation/git_hook_policy.py`:
  - Extracted `_resolve_upstream_default` (renamed from an earlier `_merged_history_upstream`), resolving `origin/HEAD` then falling back to `origin/main`. Used by both `_is_merged_history` (the co-mingling exemption) and `_read_upstream_default_blob` (a separate, pre-existing session-content-on-upstream check), which previously duplicated this logic with looser existence checking.
  - `check_branch_context`'s block message now distinguishes "branch owns a log but no upstream ref resolves" (unresolvable trunk, issue #5220's second proposed fix) from ordinary co-mingling, via a new `_print_branch_context_mismatch` helper (kept the function's cyclomatic complexity at the repo's ceiling of 10).
- `tests/test_lefthook_integration.py`: added
  - `test_branch_context_merged_history_survives_missing_origin_head` (positive: exemption fires via the `origin/main` fallback)
  - `test_branch_context_merged_history_does_not_trust_local_main` (negative: a session log present only on local `main`, with no origin remote, still blocks; proven to discriminate by temporarily reintroducing that rung and watching the test fail)
  - `capsys` assertions on `test_branch_context_merged_history_exemption_needs_an_upstream` and a new `test_branch_context_message_stays_generic_without_an_owned_log`, pinning the two message shapes
- `.agents/sessions/2026-08-21-session-5220-session-log-false-positive.json`, `.agents/qa/issue-5220-branch-context-missing-origin-head-test-report.md`: session log and QA report, including the security review, the two automated spec-coverage findings addressed, and the base-branch merges this branch needed (below).

An initial version of the fix also fell back to local `main` (mirroring `resolve_push_update`'s push-base ladder). Security review (see below) found that rung over-permissive for this call site: it would grant the exemption to a session log that only exists on the developer's own local `main`, exactly the issue #682 co-mingling shape the exemption must never excuse. It was dropped before this PR was opened.

## Acceptance criteria

- [x] `check_branch_context` no longer blocks when `origin/HEAD` is unresolved but `origin/main` is, and the branch owns its own session log
- [x] A session log present only on local `main` still does not grant the exemption (issue #682 case preserved)
- [x] Existing branch-context tests (co-mingling, merge exemption, worktree exemption, fail-open contracts) all still pass
- [x] issue #5220's second proposed fix (name which precondition failed) implemented, with tests pinning both message shapes

## Type of Change

## Reviewer Expectations

**Review kind / scrutiny / urgency**: targeted, high (git hook / co-mingling guard), no rush

**Focus areas**: whether the two-rung ladder (`origin/HEAD` then `origin/main`) is the right stopping point, whether the new negative tests actually pin the local-`main` exclusion and message differentiation correctly, and whether consolidating `_read_upstream_default_blob` onto the shared resolver is a safe behavioral tightening for that unrelated call site.

## Testing

- [x] Tests added/updated

`uv run --frozen python -m pytest tests/test_lefthook_integration.py tests/test_validate_session_json.py -q` gives 1239 passed, 1 skipped
`uv run --frozen ruff check scripts/validation/git_hook_policy.py tests/test_lefthook_integration.py` gives All checks passed
`uv run --frozen mypy scripts/validation/git_hook_policy.py` gives Success
`uv run --frozen python scripts/validation/pre_pr.py` gives 57/57 passed
`AI_AGENTS_PYTEST_WORKER_CAP=4 uv run --frozen python scripts/validation/git_hook_policy.py pytest` (full suite, matches the pre-push job) gives 27659+24+46+30 passed, 0 errors on a clean tree

### Security Review

- `scripts/validation/git_hook_policy.py`

`Task(subagent_type="security")` reviewed the three-rung version and found:

- **SEC-001** (Low): local `main` grants the exemption to a log that only exists on local `main`, the exact issue #682 shape. **Fixed**: rung dropped.
- **SEC-002** (Medium, rule violation): the docstring overclaimed parity with `resolve_push_update` without noting its fallback is conditional. **Fixed**: rung removal made the claim moot; current docstring instead explains why the two-rung version is intentionally stricter than `resolve_push_update`.
- **SEC-003** (Medium): the local-`main` rung had no test coverage. **Resolved by removal**, plus a new negative test pinning that local `main` is never trusted.

No injection risk, fail-open contract regression, or ref-peeling issues found. Full findings in the QA report.

### Other Agent Reviews

- [x] QA verified test coverage (`.agents/qa/issue-5220-branch-context-missing-origin-head-test-report.md`)

This PR's own automated review also caught two real, addressed findings: `Validate Spec Coverage` returned PARTIAL, correctly flagging the unimplemented second proposed fix (now implemented) and the duplicate `_read_upstream_default_blob` resolver (now consolidated). Both are in the "Changes" section above. `Validate PR` also caught an em dash in an earlier revision of this description (universal.md MUST NOT entry 5); rewritten to remove it.

## Author Pre-flight

- [x] Builds locally (or N/A)
- [x] Pre-PR validation passed (`uv run python scripts/validation/pre_pr.py`)
- [x] Comments added only where the why is non-obvious
- [x] Documentation updated (session log + QA report)

Note: `taste-lints` (advisory, non-blocking) flags `tests/test_lefthook_integration.py` for exceeding 500 lines and two pre-existing functions exceeding complexity 10, both unrelated to this change and not addressed here (see the QA report's "Pre-existing findings" section). A complexity finding this PR itself introduced (`check_branch_context` briefly at 11) was fixed via extraction; `taste-lints` is clean on `scripts/validation/git_hook_policy.py`.

## Notes for Reviewers

This branch merged `origin/main` twice while open, both times to pick up an already-landed, unrelated fix that was blocking this branch's own pre-push suite rather than anything wrong with this diff:

1. `36bfb510`: picked up PR #5219's re-pin of a stale `setup-uv` SHA in a test unrelated to this change.
2. `f170c11c`: picked up PR #5225's pip pin bump for a newly disclosed `pip-audit` finding (`PYSEC-2026-3721`), also unrelated to this change.

Both are recorded with the confirming evidence in the QA report. Neither touched anything this PR's diff depends on.

## Related Issues

Fixes #5220
Refs #3343, #682

### DIFF
--- a/.agents/memory/episodes/episode-2026-08-21-session-5220-session-log-false-positive.json
+++ b/.agents/memory/episodes/episode-2026-08-21-session-5220-session-log-false-positive.json
@@ -208,6 +208,46 @@
       "caused_by": [
         "e017"
       ],
+      "leads_to": [
+        "e025"
+      ],
+      "_source_session": "2026-08-21-session-5220-session-log-false-positive"
+    },
+    {
+      "id": "e022",
+      "timestamp": "2026-08-21T00:00:00+00:00",
+      "type": "milestone",
+      "content": "Pushed, opened draft PR #5226 (Fixes #5220), subscribed to its activity",
+      "caused_by": [],
+      "leads_to": [],
+      "_source_session": "2026-08-21-session-5220-session-log-false-positive"
+    },
+    {
+      "id": "e023",
+      "timestamp": "2026-08-21T00:00:00+00:00",
+      "type": "milestone",
+      "content": "Diagnose a second post-push CI failure: Python Security Checks / pip-audit PYSEC-2026-3721 in pinned pip 26.1.2",
+      "caused_by": [],
+      "leads_to": [],
+      "_source_session": "2026-08-21-session-5220-session-log-false-positive"
+    },
+    {
+      "id": "e024",
+      "timestamp": "2026-08-21T00:00:00+00:00",
+      "type": "milestone",
+      "content": "Merge origin/main a second time to pick up the pip-audit pin fix",
+      "caused_by": [],
+      "leads_to": [],
+      "_source_session": "2026-08-21-session-5220-session-log-false-positive"
+    },
+    {
+      "id": "e025",
+      "timestamp": "2026-08-21T17:20:04+00:00",
+      "type": "commit",
+      "content": "Commit: f170c11c6013cf565a5e3d553b313e9c8f1240c5",
+      "caused_by": [
+        "e021"
+      ],
       "leads_to": [],
       "_source_session": "2026-08-21-session-5220-session-log-false-positive"
     }
@@ -217,7 +257,7 @@
     "tool_calls": null,
     "errors": 0,
     "recoveries": 0,
-    "commits": 3,
+    "commits": 4,
     "files_changed": 5
   },
   "lessons": []
--- a/.agents/memory/episodes/episode-2026-08-21-session-5220-session-log-false-positive.json
+++ b/.agents/memory/episodes/episode-2026-08-21-session-5220-session-log-false-positive.json
@@ -96,7 +96,77 @@
       "type": "commit",
       "content": "Commit: fb0f0fa15b6dd0d20c1f2d75650dcc991f78ebaf",
       "caused_by": [
-        "e007"
+        "e007",
+        "e015"
+      ],
+      "leads_to": [
+        "e017"
+      ],
+      "_source_session": "2026-08-21-session-5220-session-log-false-positive"
+    },
+    {
+      "id": "e011",
+      "timestamp": "2026-08-21T00:00:00+00:00",
+      "type": "milestone",
+      "content": "Run the security agent on the git_hook_policy.py change per .claude/rules/security.md MUST-1 before pushing",
+      "caused_by": [],
+      "leads_to": [],
+      "_source_session": "2026-08-21-session-5220-session-log-false-positive"
+    },
+    {
+      "id": "e012",
+      "timestamp": "2026-08-21T00:00:00+00:00",
+      "type": "milestone",
+      "content": "Drop the local-main rung, keeping only origin/HEAD -> origin/main",
+      "caused_by": [],
+      "leads_to": [],
+      "_source_session": "2026-08-21-session-5220-session-log-false-positive"
+    },
+    {
+      "id": "e013",
+      "timestamp": "2026-08-21T00:00:00+00:00",
+      "type": "milestone",
+      "content": "Add a negative test proving local main is never trusted, and prove it discriminates",
+      "caused_by": [],
+      "leads_to": [],
+      "_source_session": "2026-08-21-session-5220-session-log-false-positive"
+    },
+    {
+      "id": "e014",
+      "timestamp": "2026-08-21T00:00:00+00:00",
+      "type": "milestone",
+      "content": "Re-run the full suite, ruff, mypy, and pre_pr.py against the final tree",
+      "caused_by": [],
+      "leads_to": [],
+      "_source_session": "2026-08-21-session-5220-session-log-false-positive"
+    },
+    {
+      "id": "e015",
+      "timestamp": "2026-08-21T00:00:00+00:00",
+      "type": "test",
+      "content": "865 passed, 1 skipped; ruff and mypy clean; pre_pr.py 57/57",
+      "caused_by": [],
+      "leads_to": [
+        "e010"
+      ],
+      "_source_session": "2026-08-21-session-5220-session-log-false-positive"
+    },
+    {
+      "id": "e016",
+      "timestamp": "2026-08-21T00:00:00+00:00",
+      "type": "milestone",
+      "content": "Unshallow the clone after the first push attempt failed the push-ref-policy hook",
+      "caused_by": [],
+      "leads_to": [],
+      "_source_session": "2026-08-21-session-5220-session-log-false-positive"
+    },
+    {
+      "id": "e017",
+      "timestamp": "2026-08-21T16:27:34+00:00",
+      "type": "commit",
+      "content": "Commit: c19f1e73c0786f5af43c67d8c6c14e7362f21d20",
+      "caused_by": [
+        "e010"
       ],
       "leads_to": [],
       "_source_session": "2026-08-21-session-5220-session-log-false-positive"
@@ -107,8 +177,8 @@
     "tool_calls": null,
     "errors": 0,
     "recoveries": 0,
-    "commits": 1,
-    "files_changed": 2
+    "commits": 2,
+    "files_changed": 5
   },
   "lessons": []
 }
--- a/.agents/memory/episodes/episode-2026-08-21-session-5220-session-log-false-positive.json
+++ b/.agents/memory/episodes/episode-2026-08-21-session-5220-session-log-false-positive.json
@@ -248,6 +248,64 @@
       "caused_by": [
         "e021"
       ],
+      "leads_to": [
+        "e031"
+      ],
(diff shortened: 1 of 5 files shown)


rjmurillo:
Closing unmerged: superseded by `main`'s own fix, already landed independent of this branch.

`origin/main` commit `746e2c36a` (PR #5229, tracking issue #5228, "Discontinue session log file creation") rewrote the same two functions this PR touches, `_is_merged_history` and `check_branch_context` in `scripts/validation/git_hook_policy.py`, and its version is a strict superset of this one:

- It routes `_is_merged_history` through the pre-existing `_read_upstream_default_blob`, which already carried an `origin/HEAD` to `origin/main` fallback before either branch started (`git_hook_policy.py:927-929` on `main`). That closes the exact gap issue #5220 reports, without needing the new `_resolve_upstream_default` resolver this PR added.
- It goes further than this PR by dropping the "current branch owns a same-day log" precondition entirely, which this PR kept. That precondition is now permanently unsatisfiable (session log creation is discontinued per `.claude/rules/session-logs.md` MUST 1 as of the same commit), so keeping it would have reintroduced a false-positive block the moment any other branch's same-day log lands upstream.
- It adds its own regression coverage for the merged-history exemption, including a case this PR did not cover (a locally tampered, uncommitted edit to an already-upstream log path).

I verified this empirically rather than trusting the diff: checked out `origin/main` at `746e2c36a` in an external worktree, reproduced issue #5220's exact scenario (a fresh clone-shape repo with `origin/HEAD` unresolved, a same-day log from another branch imported by merge, and the current branch owning its own log), and called `check_branch_context` directly. It returned `0` (pass), where the pre-fix code returned `1` (block).

This branch's own merge attempt against `origin/main` failed pre-push's `merge-tree-ratchet` because both sides changed the same functions; reconciling would mean discarding this PR's narrower fix in favor of the already-merged, more complete one, which leaves nothing left for this PR to contribute.

Refs #5220 (fixed by `746e2c36a` on `main`, not by this PR). Refs #5228, #3343, #682.

---

Issue #5318, opened 2026-08-25
perf(hooks): the pre-push tail is unmeasured, and the local selector still disagrees with CI's filter


rjmurillo:
Follow-up to #5315. ADR-104 cut the median pre-push cost and explicitly did not address these. Raised by the adr-review debate on that record.

## 1. The tail is unmeasured

ADR-104 lists five pre-push jobs that match its own rule 3 (CI takes expensive, isolated, matrixed, or credentialed work) and stay:

| Job | Declared cap | Glob-gated |
|---|---|---|
| `hook-anchoring-e2e` | 20m | yes, hook paths |
| `plugin-load-e2e` | 20m | yes, skills/commands/plugin manifests |
| `workflow-local-run` | 30m | yes, `.github/workflows/**` |
| `python-type-check` | 15m | yes, `**/*.py` |
| `security-scan` | 15m | no |

Four cost nothing on the Markdown change class that was measured. **None has been timed when its glob does fire.** Nobody in this repository can say what a `.claude/skills/**` push costs, and those are the pushes that will still outlive a container. Summing declared timeouts through lefthook's group semantics gives a worst case far above ADR-104's 300s target.

Ask: measure each of the five when its glob fires, on a real push, per `ci-scripts.md` MUST-16. Then decide per job.

## 2. Pre-commit's 60s target has no measurement behind it

ADR-104 states a 60s pre-commit target and audited pre-push only. Pre-commit has 47 jobs. Two real commits in the originating session measured 8.79s and 8.45s, but those staged 3 to 4 small files and are not a representative sample. Agents commit far more often than they push.

Ask: measure a representative pre-commit run and either keep the number or replace it.

## 3. The local selector and CI's paths filter disagree by construction

`scripts/test_selection/select_tests.py` bails to the full suite on **any** non-Python path. `.github/workflows/pytest.yml` uses a curated allowlist. `scripts/test_selection/runtime_read_patterns.txt` is nearly a copy of that allowlist, so the two are already trying to express the same thing in two places.

ADR-104's first revision rejected "narrow the selector" on the ground that a hand-derived allowlist fails open. That reasoning was wrong and is corrected in the record: CI already depends on exactly such an allowlist, so the objection would condemn the existing required gate. The real objection is narrower: an allowlist does not bound cost for a large Python change.

Ask: make one file the single source read by both, with a test asserting they agree. #5316's filter widening moved them closer by hand; that is the drift this would prevent.

[...]

Issue #5317, opened 2026-08-25
ci(hooks): pre-pr-validation duplicates four fast-stage gates, and the obvious fix is unsound


rjmurillo:
Split out of #5315 so the duplication is tracked rather than lost, and so the reasoning that made the first attempt wrong is written down.

## The duplication is real

`pre-pr-validation` runs four checks the pre-push fast stage also runs as their own jobs:

| pre_pr gate | fast-stage job |
|---|---|
| Count Ratchets | the eight ratchet jobs (`checks_ratchet.RATCHETS` job names match `lefthook.yml` exactly) |
| Unreachable Code Detection | `python-unreachable-statements` |
| Path Normalization | `path-normalization` |
| Planning Artifacts | `planning-artifacts` |

Measured on a 4-CPU container: the ratchet re-run alone was 38.37s of that job's 103.25s (`.serena/memories/ci/ci-pre-push-wall-clock-is-python-tests.md`). A reviewer measured the job at 65.02s with the four gates skipped against 103.25s recorded without, so the duplication is worth roughly 40s of a ~150s hook.

## Why the first attempt was reverted

The skip was gated on `AI_AGENTS_PRE_PR_FAST_STAGE_RAN=1`, set only by the `pre-pr-validation` job, justified by the hook being `piped: true`.

`piped: true` proves no earlier job **failed**. It does not prove one **ran**. Every deferral target carries a `glob:`; `pre-pr-validation` carries none, so it always runs and always set the flag.

Measured on lefthook 2.1.10 against a fixture repo with one glob-gated job and one un-gated job, pushing a docs-only commit:

```
|  py-only-gate (skip) no matching push files
|  always-gate > ALWAYS GATE RAN
summary: (done in 0.02 seconds)   OK always-gate
EXIT=0
```

A glob-skipped job is indistinguishable from a passed one. Resolving the real globs against three push shapes:

| push contains | pre_pr gates that would run nowhere |
|---|---|
| only `.py` | Count Ratchets (partial), Path Normalization, Planning Artifacts |
| only `.agents/architecture/*.md` | Count Ratchets (partial), Unreachable Code, Planning Artifacts |
| only `.github/workflows/*.yml` | all four |

[...]

Issue #5066, opened 2026-08-15
ci(hooks): add a fast-fail stage to pre-push so cheap gates run before pytest


rjmurillo:
## Problem

The pre-push suite is a 12 to 17 minute serialization point where cheap gates effectively surface after the expensive ones. A ratchet or policy failure that takes seconds to detect is discovered after the full pytest run.

## Evidence

- Measured: `pre-pr-validation` 75.86s, `python-tests` 741.85s (`.serena/memories/ci/run-count-ratchets-before-the-expensive-pre-push.md`).
- `.agents/retrospective/2026-08-05-pr-queue-and-doctrine.md`: 13 pushes burned, about 3 hours, in one session. Quote: "I treated the pre-push suite as a checker rather than as a 15-minute serialization point." The retro proposes a "pre-push gate trio" (about 20 seconds) that was never built.
- Retros for PRs #4636, #4684, #4924, #4944, #4945, #5002 each re-derive the same ordering lesson as a discipline note. The fix is ordering in tooling, not discipline.

## Proposal

Restructure `lefthook.yml` pre-push into two stages:

1. Stage A (fast fail, target under 60s): count ratchets, retrospective-policy, session-log policy, QA-report discovery (filename glob match), branch-freshness check vs origin/main, dash guard.
2. Stage B: python-tests, semgrep, mypy, build_all --check, workflow-local-run, cli-hook-e2e.

Stage B does not start until Stage A passes. Lefthook `piped: true` groups can express this.

## Acceptance criteria

- [ ] A failure detectable in under 60 seconds never costs a full pytest run
- [ ] Total pre-push wall time for a clean push is not worse than today by more than 10 percent
- [ ] All existing gates still run; only ordering changes

Refs #4251, #4491

rjmurillo:
Closing: fixed by PR #5083 (merged 2026-08-16). `lefthook.yml:390-720` is now staged exactly as proposed: singleton guards, then a piped cheap-policy group, then a parallel ratchet/policy group, then security-scan, then the expensive group (python-tests, mypy, e2e). Five of six Stage A gates named in the issue are now fast-staged; the sixth (dash guard) is the sole documented remainder, already split into its own issue #5086 (still open) per this PR's own description. Nothing left to fix under #5066 itself.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #4710, opened 2026-08-06
perf: reduce local hook and validation latency


rjmurillo:
## Problem

Local pre-commit, pre-push, and pre-PR validation repeats avoidable work. Pushes can block for several minutes, then restart after a branch race or CI correction.

Measured on 2026-08-06:

- `uv run --frozen python -c pass` took 0.136 to 0.283 seconds. Direct venv Python took 0.024 to 0.026 seconds.
- The hook graph starts about 44 `git_hook_policy.py` processes.
- The function-scoped HEAD guard ran two Git subprocesses per test across more than 23,000 tests.
- Full workflow actionlint took 0.43 to 0.54 seconds. One changed workflow took 0.021 seconds.
- Main-suite line coverage took 438.91 seconds. Branch coverage took 465.93 seconds and failed on mixed child-process data.

## Planned stack

1. Scope pre-PR Markdown, workflow, and YAML checks to committed, staged, unstaged, and untracked paths. Cache only the GitHub PR-base probe within one validation run.
2. Replace common-case per-test `git rev-parse HEAD` calls with direct loose-ref reads. Keep Git fallback for packed refs and every unproven state.
3. Execute coverage-pin tests once. Keep the main suite on statement coverage, preserve isolated 100% branch gates, and merge projected pin lines into the final report.
4. Record the measured Lefthook stdin and autouse-fixture rules in canonical sources and generated mirrors.

## Acceptance criteria

- No validation coverage is removed.
- Full pytest suite passes.
- `scripts/validation/pre_pr.py` passes.
- Actionlint, mypy, Ruff, workflow dry-run, generation, rule budget, and taste ratchets pass.
- GPT 5.6 Sol correctness and performance reviews return clean after all fixes.
- Changes land as atomic stacked PRs.

## Deferred work

- Consolidating dispatcher process launches needs a separate architecture change.
- Parallelizing pre-push stdin consumers remains unsafe. Lefthook 2.1.10 corrupts large stdin payloads under `parallel: true`; `piped: true` delivers complete copies.

Related: #3175, #3862, #4692, #4695, #4699, #4700.

rjmurillo:
## Velocity Accelerator: 2 Opportunities Detected

### Triage issue #4710: perf: reduce local hook and validation latency
- **Type**: `complexity_triage`
- **Priority**: high
- **Suggested Agent**: agent-architect
- Issue labeled with high complexity.
Suggested agents: agent-architect, agent-devops, agent-qa

[...]

PR #5319, opened 2026-08-25
perf(hooks): collect the pytest suite in pre-push instead of running it

### LINKED ISSUE #5315: perf(hooks): pre-push runs the whole pytest suite locally and outlives container pushes
## Symptom

A `git push` from a remote-container session takes long enough that the container is reclaimed mid-push. The push dies, the session restarts, re-derives its state, and pushes again. Each cycle costs wall clock and tokens and produces no signal that CI would not have produced.

## Measurement

Measured on a 4-CPU container, full clone, `AI_AGENTS_PYTEST_WORKER_CAP=4`:

| pre-push job | wall clock |
|---|---|
| `python-tests` (`git_hook_policy.py pytest`) | **382s** |
| fast stage (14 parallel gates) | ~20s isolated, 56s in-hook |
| every other blocking gate | under 20s each |

> **Correction (2026-08-25).** This issue originally said the 382s run "exited 1" on `tests/mutation/test_mutate_debate_log_path.py` for container reasons. That was wrong. The mutation harness refuses to run when one of its target files is dirty, and the target was being edited while the run was in flight. The failure was self-inflicted; the `gh` GraphQL line quoted alongside it came from a partition that passed. 382s is a duration measurement and nothing more.

`.serena/memories/ci/ci-pre-push-wall-clock-is-python-tests.md` records the same shape independently: 679s total to push one Markdown file on a 4-CPU container, 498.52s of it in `python-tests`.

## Root cause

`scripts/test_selection/select_tests.py` falls back to the full suite on `non-Python change: `. Probed on this checkout:

```
README.md                        -> FULL (non-Python change)
lefthook.yml                     -> FULL (runtime-read pattern)
.serena/memories/memory-index.md -> FULL (non-Python change)
scripts/validation/pre_pr.py     -> 25 tests via import graph
```

This repository's changes are mostly Markdown, so nearly every push takes the full-suite path. CI does not: `.github/workflows/pytest.yml` carries a `dorny/paths-filter` allowlist and skips the suite for unrelated Markdown, with the comment "unrelated Markdown, such as session logs, still skips pytest". The local gate is therefore strictly more expensive than the remote gate it duplicates, and the duplicate is what stops the push from reaching the remote gate.

Second, smaller cause: `pre-pr-validation` re-runs four checks the pre-push fast stage already ran as their own jobs (the eight count ratchets, unreachable code, path normalization, planning artifacts). Measured previously at 38.37s of that job's 103.25s for the ratchets alone.

## Ask

1. Stop executing the whole suite inside pre-push. Keep the import-graph subset when the graph can narrow the diff; collect rather than execute when it cannot, so import and syntax breakage still blocks before it wastes a CI matrix run.

   **Scope, corrected twice. Read the second correction; the first was wrong.**

   *First narrowing (2026-08-25).* "Collect rather than execute" is not a cheaper version of the suite. Collection imports each test module and instantiates each item, so it blocks on some defect classes and not others, and only the probed ones may be claimed.

   *Second correction (2026-08-26), after review.* The first narrowing listed **three** catches, including a same-basename module collision. That one is false here. A basename collision is a collection error under pytest's default `prepend` import mode and collects clean under `--import-mode=importlib`, which `pyproject.toml` sets. Every probe behind the original claim ran in a throwaway tree with no config and silently measured the other mode, so it measured a pytest this repository never runs.

   Re-probed under production configuration:

   | class | exits |
   |---|---|
   | broken import | **2** (blocks) |
   | syntax error | **2** (blocks) |
   | same-basename module collision | **0** (collects clean) |
   | missing fixture | 0 (collects clean) |
   | two same-named test functions in one module | 0 (collects clean) |

   So this ask is satisfied by **two catches**, not three. Missing fixtures, duplicate test-function names, and basename collisions all stay with CI by design and are not part of what satisfies it.

   The same two-catch, three-miss list is stated in `_pytest_collection_command`, in ADR-104 rule 5, and in the text the hook prints on a fallback push. Those three are pinned against each other by `tests/validation/test_collection_contract_surfaces.py` with a mutation control. This issue body is a fourth surface that no repository test can read, and it went stale within one commit of being documented as hand-maintained, which is the whole argument for not counting it as enforcement.

2. ~~Skip the four `pre-pr-validation` gates the fast stage already ran, under a flag only the hook sets.~~ **Withdrawn 2026-08-25: the premise is false.** See below.
3. Write the ADR that says which checks belong in pre-commit, which in pre-push, and which in CI, so the boundary is a decision with a stated budget rather than an accretion.

## Ask 2 is withdrawn, and the reason is worth keeping

The proposed skip was implemented, reviewed, and reverted. It rested on the pre-push hook being `piped: true`, which proves no earlier job **failed**, not that one **ran**. Every gate the skip deferred to carries a `glob:`, and `pre-pr-validation` carries none, so it always ran and always set the flag.

Measured on lefthook 2.1.10 against a fixture repo with one glob-gated job and one un-gated job, pushing a docs-only commit:

```
|  py-only-gate (skip) no matching push files
|  always-gate > ALWAYS GATE RAN
summary: (done in 0.02 seconds)   OK always-gate
EXIT=0
```

A glob-skipped job is indistinguishable from a passed one. Resolving the real globs, a Python-only push would have skipped Path Normalization and Planning Artifacts with nothing running them; a workflow-only push would have skipped all four. All four are whole-tree scans, so they were the drift backstop the globbed jobs are not.

The duplication is real and worth roughly 40s of a ~148s hook. **Issue #5317** carries three costed options for removing it soundly. The correct behavior for this issue is the opposite of the original ask: those four gates must run on every push, and a test now fails if the flag, a deferral, or the outright deletion of a gate returns.

## Out of scope, tracked separately

**Issue #5318**: measuring the glob-gated pre-push tail on a workstation, the unmeasured pre-commit target, the selector/CI-filter divergence, the ADR-054 reconciliation, and the aggregate deadline `scan_pushed_heads` lacks (the container clamp bounds one child process, not one job).

Refs #4710, #5066, #5317, #5318.

rjmurillo:
# Pull Request

### What

Pre-push stops executing the whole pytest suite, caps are resized from measured in-hook time, and no `git_hook_policy` process can outlive a managed container. ADR-104 records the tier boundary.

### Why

A push from a remote container blocked long enough inside the pre-push hook that the container was reclaimed mid-push. The push died carrying no diagnostic, the session restarted, re-derived its state, and pushed again. The local run duplicated CI's `pytest.yml` partition matrix, so a copy of the remote gate was what stopped the push from reaching the gate it was copying.

Measured, twenty-seven consecutive real pushes: **140.5s to 165s**, median ~150s.

| component | before (recorded) | now (in-hook) |
|---|---|---|
| `python-tests` | 498.52s | **14.08s** collecting, **39.15s** on an import-graph subset |
| `pre-pr-validation` | 110.12s | ~105s |
| `security-scan` | not separated | ~22s |
| **total** | **679s** | **~152s** |

## Specification References

| **Issue** | Closes #5315 | pre-push runs the whole pytest suite locally and outlives container pushes |
| **Issue** | Refs #5317 | pre-pr-validation duplicates four fast-stage gates, and the obvious fix is unsound |
| **Issue** | Refs #5318 | the pre-push tail is unmeasured |
| **Issue** | Refs #4710 | reduce local hook and validation latency |
| **ADR** | `.agents/architecture/ADR-104-gate-tier-placement-and-budgets.md` | Gate Tier Placement |

These references do not close their linked items: #5317, #5318, #4710.

## Changes

- The whole-suite fallback collects rather than executes. Collection imports every test module, so **a broken import and a syntax error** still block the push. `AI_AGENTS_PYTEST_FULL_SUITE_LOCALLY=1` executes every partition locally whatever the import graph says, and rejects any other value rather than quietly doing less.
- `run_pytest` refuses an empty command list. A push must never pass by running zero tests.
- **The container bound is on the process, in two layers.** `_run_command` and `_run_command_bytes` clamp each child to 150s when a managed container is detected; `run_pytest` and `scan_pushed_heads` hold aggregate deadlines over their own children; and `main` arms a watchdog at 165s covering the whole process, including selection and setup, which no per-child deadline reaches. Workstations and CI are untouched. When the detector cannot be imported, neither the clamp nor the watchdog applies and both say so on stderr: a net that can block a legitimate push is worse than no net, but one that lapses in silence is worse than both.
- `.github/workflows/pytest.yml` now matches every tracked Markdown, JSON and text file by extension, plus five extensionless names. See below: an audit found 120 candidate test inputs the allowlist did not match.
- Caps resized from in-hook measurement, with inner budgets pulled under their outer caps so the inner timeout fires first (ADR-086 item 9). One cap that had no measurement behind it is back at main's value.
- ADR-104 plus its nine-round debate log.

## What review changed

Ten Copilot rounds ran against this branch after it left draft. Every one found a load-bearing claim that was wrong. None was caught by seven spec-validator runs, six ADR debate rounds, or four of my own review agents.

Across all ten the finding is one thing said ten ways: **a claim true of the case in front of the author, generalized to the case the gate actually runs in.** Each round moved it one level out.

**A defect class I claimed to catch does not exist here.** `pyproject.toml` sets `--import-mode=importlib`. A same-basename module collision is a collection error under pytest's default `prepend` mode and collects clean under `importlib`. Every probe behind the claim ran in a throwaway tree with no config, so all of them measured a pytest this repository never runs.

Re-probed under production configuration: broken import exits 2, syntax error exits 2, basename collision exits **0**. The contract is **two catches, three misses**, and all five are now executed rather than asserted about as strings. Fixtures copy `addopts` out of the real `pyproject.toml`, so the drift path that produced the false claim is gone.

**The bound, four rounds running.** The clamp covers one child. `run_pytest` emits up to four partition commands and took the 780s execution budget on the ordinary import-graph subset an everyday Python push takes. `scan_pushed_heads` loops over pushed refs and batches targets within a ref, each batch taking a fresh 150s. Then, with both aggregates in place, review found the work *between* children was still outside every deadline: `changed_from_git` shells out with no timeout and the import graph is built in-process, both before `run_pytest` starts its clock; `scan_pushed_heads` discovers paths, materializes a tree and probes semgrep's version before starting its own. Threading a deadline into each of those leaves the next one for the next reviewer, so the bound went around the process.

**The full-suite opt-in did less than its name, twice.** Hoisting `_validated_full_suite_opt_in` made an *invalid* value report on both paths and left a *valid* one acting on only one, so `=1` on a narrowable Python change ran whatever subset the selector chose, silently. And the variable was never stripped from the child environment, so a test invoking this policy inherited `=1` and took the full-suite path where it meant to exercise the selector.

**Exit 3 is overloaded three ways.** `_run_command` synthesizes it on timeout and on an `OSError` launching a child; pytest returns it for an internal error. `run_pytest` branched on the bare code, so a crash was announced as "pytest suite timed out" and sent the reader to the budget instead of the traceback the child had already printed.

**Seven guards written for this PR could not fail.** The collection-contract check read the entire ADR while calling its surface "ADR-104 rule 5". Its polarity guard searched for a bare term where every surface writes an article, so the one wording a demoted catch would arrive in was the one it could not see. The declared-budget module opened "as a ratchet that may only fall" with its baseline twenty lines above the assertion reading it. `CONTAINER_UNCLAMPED_JOBS` graded jobs by name. The deferred-gates control asserted an invented name was absent. The ratchet control's `or "rose from" in failure` half is in every failure message. And a fake written *this round* defaulted the captured deadline to `0.0`, so removing the production wiring recorded three identical sentinels and both assertions held against an unbounded job. Each is now derived from something the branch cannot edit in the same breath, or driven by a real mutation.

**The ratchet's enforcement point was stated backwards.** The record said the base-ref comparison runs on pre-push and skips in CI. `pytest.yml` checks out with `fetch-depth: 0`, so CI has the base ref and executes it; pre-push, on a change to `lefthook.yml` alone, cannot narrow the import graph from a YAML file and takes the collection stand-in, which imports the test without running it. Separately, `_config_at_ref` turned any `OSError` into "no base ref", so a git launch failure disabled the only baseline the branch cannot edit and reported a shallow checkout instead.

**One cap was cut without a measurement.** ADR-104 rule 7 says a cap is sized from a measured worst case. `workflow-local-run` went 30m to 10m and its child budget 1740s to 540s. A push carrying a workflow change did fire it, at 0.42s, and that is not the measurement the cut needed: this container has no actionlint, so the job reports DEGRADED before doing any work. Both values are back at main's, taking the declared workstation sum from 2850s to **3450s**. The base-ref ratchet still holds, because the base ref declares 4170s.

### The CI filter, audited rather than curated

Four rounds each found more filter roots by reading tests. The fifth asked for a reverse audit instead of a fifth batch, so one was run: parse every string literal in all 974 test modules, keep the ones that resolve to a tracked file, check each against the filter.

**120 candidate inputs were unmatched.** 92 Markdown, 14 JSON, 5 text, spanning every tree including `.agents/retrospective/**`, `.agents/sessions/**`, and `.agents/analysis/**`, which the filter's own comment names as the Markdown it deliberately skips. Two session JSONs are opened by exact path in `tests/test_validate_session_json.py`, so the tree the filter protected as inert was carrying real inputs.

Three extension globs plus five extensionless names close all 120; re-running the audit leaves one `__pycache__` artifact. The trade was the repo owner's call: a docs-only PR now runs the matrix it used to skip, so the failure direction flips from a gate silently not running to a run being wasted.

That 120 is a superset, not a census: the audit resolves string literals against the tree, so a test building `tmp_path / "STEP-0.5-METRICS.md"` matches the tracked file of that name. It is evidence the allowlist could not be finished by hand, not a count of dependencies.

Also from review: six citations pointed at #5316, a closed Renovate bump; the ADR recorded a retired 1740s budget and an unchanged 30m cap the branch had already cut; the debate log's round count was stale three times; a canonical-source citation named the wrong line and claimed a byte-exact quote that carried a four-space indent; a guard's docstring promised proof of a runtime read the check does not perform; four comments described the pre-clamp world; and both pytest flags were missing from the config catalog.

## Two kinds of claim, kept apart

**Enforced** means a test in this diff fails when the claim stops holding, and a control proves that test can fail. **Measured** means a number was observed and recorded, and nothing stops it drifting.

### Enforced

| Claim | Test | Control |
|---|---|---|
| The fallback collects rather than executes, with the collection budget not the suite budget | `test_pytest_import_selection.py::test_resolve_collects_on_full_verdict`, budget-switch test | opt-in flag path asserted separately |
| A broken import blocks the push; the same tree with it removed proceeds | `test_collection_probes.py::test_a_broken_import_makes_the_collection_stand_in_block_the_push` | the repair half is the control; runs the real command, mirrors production addopts |
| A syntax error blocks the push | `::test_the_other_probed_catch_also_blocks_the_push` | same repair-half structure |
| A basename collision does **not** block, under this repo's import mode | `::test_a_same_basename_collision_goes_uncaught_under_production_config` | removing the importlib flag from pyproject fails it |
| A missing fixture and two same-named test functions do **not** block | `::test_a_claimed_miss_really_does_collect_clean` | a syntax error appended to the same tree must block, proving collection reached it |
| An invalid opt-in value is reported on both paths | `test_pytest_import_selection.py::test_a_rejected_opt_in_value_is_reported_on_the_narrowed_path_too` | the fallback-path case is the paired assertion |
| A **valid** opt-in beats a narrowed selection | `::test_a_valid_opt_in_beats_a_narrowed_selection` | the invalid-value sibling is the paired case that stayed green while this one was broken |
| The opt-in does not reach the pytest child's environment | `test_pytest_parallelism_policy.py::test_run_pytest_consumes_worker_override_before_child_process` | asserted beside its two siblings, which were already stripped |
| A pytest internal error is not reported as a timeout | `test_pytest_import_selection.py::test_a_pytest_internal_error_is_not_reported_as_a_timeout` | reverting to the bare exit code fails it; the sibling covers a real timeout |
| `run_pytest` refuses an empty command list | `::test_run_pytest_refuses_an_empty_command_list` | n/a, direct assertion |
| A container bounds the whole pytest step on the opt-in path | `test_safe_push_pr_branch.py::test_a_container_bounds_the_whole_pytest_step_not_just_each_child` | dropping the aggregate clamp fails it at 779.99s against a 150s ceiling |
| The ordinary import-graph subset takes the 780s execution budget, and a container bounds it too | `::test_the_ordinary_subset_path_takes_the_execution_budget_too` and `::test_a_container_bounds_the_ordinary_subset_path_as_well` | the unclamped half is the control for the clamped one |
| The security scan carries one deadline across every ref and every batch | `test_security_scan_budget.py`, seven cases | removing the wiring and the between-refs check fails four of them |
| The scan's deadline is actually passed down, not defaulted | `::test_the_job_carries_a_deadline_at_all` | the capture fake has no default, so an absent argument raises rather than reading as zero |
| A watchdog bounds the whole process, and only in a container | `::test_the_watchdog_arms_only_in_a_container` | the workstation half is the control for the container half |
| An inner deadline always fires before the watchdog | `::test_the_process_ceiling_sits_above_the_subprocess_one` | n/a, direct ordering assertion |
| Both subprocess entry points clamp in a container | `test_lefthook_integration.py::test_both_subprocess_entry_points_clamp_in_a_container` | removing the byte-runner clamp fails that half and only that half; `::test_neither_entry_point_clamps_on_a_workstation` pairs it |
| The notice states every probed miss, and states catches as catches | `test_collection_contract_surfaces.py::test_the_notice_states_every_probed_miss_not_just_the_first` | rewriting the notice to disclaim the broken import fails it |
| Docstring, notice, and ADR-104 rule 5 name the same class set | `::test_every_in_repo_surface_states_the_whole_collection_contract` | `::test_the_contract_check_can_fail` strikes the class from each surface in turn |
| No surface disclaims a class it is supposed to catch, in either wording | `::test_no_surface_disclaims_a_class_it_is_supposed_to_catch` | inverting the ADR's own sentence, with the article it actually uses, fails it |
| The ADR surface is rule 5 alone, so the Alternatives table cannot mask a deletion | `::test_deleting_a_catch_from_rule_5_is_not_masked_by_the_rest_of_the_record` | reverting the narrowing fails it with the reviewer's own sentence |
| The four formerly-deferred gates are still in `_SEQUENCE` | `test_lefthook_prepush_tiering.py::TestTheDeferredGatesStillExist` | `::test_the_presence_check_notices_a_removed_gate` deletes a real gate and requires the guard to name it |
| No gate defers to a scheduler claim, no job sets the reverted flag | `::TestTheUnsoundSkipStaysReverted` | `::TestAnyFutureDeferralMustBeSound` |
| Every exit-swallowing pre-push job is allowlisted | `::TestNonBlockingJobsAreDeclaredNotDiscovered` | adding an always-true suffix to `bot-cascade-advisory` fails it |
| No single **child process** exceeds 150s in a container | `test_lefthook_container_bound.py::TestNoSingleChildOutlivesAContainer` | `::test_the_clamp_actually_binds` |
| The clamped roster matches what the config actually routes, both directions | `::test_the_roster_matches_what_the_config_actually_routes` | verified by mutation in each direction, each naming the job |
| The declared workstation sum does not rise against the base ref | `test_lefthook_declared_budget.py::test_declared_budget_does_not_rise_against_the_base_ref` | `::test_the_ratchet_catches_a_cap_raised_on_this_branch`, which now asserts both totals rather than a phrase every failure carries |
| The declared sum stays under its ceiling where no base ref is reachable | `::test_declared_budget_is_at_or_below_its_baseline` | `::test_a_raised_cap_breaks_the_budget` raises a cap to 99h |
| `workflow-local-run` keeps main's 30m cap until a measurement supports a cut | `test_lefthook_integration.py::test_configuration_bounds_every_job` | n/a, direct pin |
| Every `git_hook_policy` job keeps 30s headroom over its child budget | `::test_each_python_subprocess_budget_has_lefthook_headroom` | invariant over the whole roster |
| CI does not inherit the container clamp | `::test_ci_keeps_the_full_budget_even_inside_a_container_image` | paired with the container case |
| Skipping the clamp on an unimportable detector says so | `::test_an_unavailable_container_detector_says_so_before_going_unclamped` | removing the warning fails it |
| Every non-extension `pytest.yml` filter root is named by a test | `test_pytest_paths_filter_roots.py` | `::test_the_reader_check_can_fail` |
| The widened filter selects session Markdown, deliberately | `test_pytest_paths_filter_covers_episodes.py::test_the_filter_now_selects_tracked_session_markdown` | the inverse of what this test asserted before the audit |

### Measured, not enforced

- **"Finishes in seconds."** Twenty-seven real pushes, 140.5s to 165s. No test asserts wall-clock; a timing assertion is flaky across host speed and contention. What is enforced is the shape producing it.
- **The workstation declared sum, 57.5 minutes against ADR-104's 300s target.** The ratchet stops it rising against the base ref; nothing makes it fall (#5318).
- **The CI filter's reverse direction.** The audit that found the 120 candidates was run by hand and its result is a snapshot, not a gate. The forward guard (every named root is named by a test) is enforced; nothing fails when a new unmatched input appears.
- **`workflow-local-run` on the path its cap protects.** 0.42s measured in a container, where the job degrades to a warning. The `act` run a workstation performs has never been timed in-hook, which is why the cap is back at main's value rather than at a smaller one.
- **The watchdog firing.** Its arming, its ceiling, and its ordering against the inner deadlines are enforced. Nothing tests a real 165s hang, because a test that hangs for 165s is a test nobody runs.

## Type of Change

## Reviewer Expectations

**Review kind / scrutiny / urgency**: design review, high scrutiny, no rush

Judgement calls I would like a second opinion on:

1. **Is 165s the right whole-process ceiling, and is `os._exit` the right exit?** It clears every measured job by at least 4x. The abrupt exit loses a temporary directory; the alternative is losing the push and its diagnostic, which is the failure this PR exists to remove.
2. **Is `SEMGREP_TIMEOUT_SECONDS` the right aggregate for the security scan?** It is what one scan already had, so a one-ref workstation push is unchanged, but a multi-ref push now shares it instead of getting it per ref.
3. **Should the reverse audit become a test?** It takes ~4s over 974 modules and would make the filter self-closing. I left it out because the extension globs made the allowlist stop needing curation, which is the reason the guard would have existed.
4. **Should any of the six advisory reporters move out of pre-push?** ADR-104 rule 9 states the criterion; applying it would move or delete jobs.

## Testing

- [x] Tests added/updated

Full suite run locally: 28,395 passed, 77 skipped. Every control in the Enforced table was verified by mutation, not by inspection.

Two traps worth naming for anyone reproducing that number. The mutation harnesses refuse to run while their targets have uncommitted changes, and `build_all.py --check` counts a regenerated-but-uncommitted file as drift, so a full-suite run mid-edit produces failures that are preconditions rather than defects. And mypy rejects attribute assignment on a `ModuleType` while ruff's B010 autofix rewrites the `setattr` form straight back into it, so a fake module has to be built through its namespace to satisfy both.

### Security Review

**Files requiring security review:** `lefthook.yml`, `scripts/validation/git_hook_policy.py`

Both P1 findings addressed: the skills tree reaches CI's filter, and the flag whose bypass surface it flagged is gone with the reverted skip. The security scan's aggregate deadline and the process watchdog make the gate stricter, never weaker: each can refuse a scan, never skip one silently.

### Other Agent Reviews

- [x] Architect reviewed design changes
- [x] Copilot review addressed (ten rounds)

## Author Pre-flight

- [x] Builds locally (or N/A)
- [x] Pre-PR validation passed (as the `pre-pr-validation` pre-push job, ~105s, green)
- [x] Code follows project style guidelines
- [x] Self-review completed

## Known gaps

- `push-ref-policy` carries a 2m cap over `check_push_refs`, which runs many git children sequentially at 90s each. Bounded in a container by the watchdog, like every other `git_hook_policy` job; on a workstation the cap remains unproven against that aggregate.
- The base-ref budget ratchet skips where no base ref is reachable. Its normal enforcement point is CI, which checks out at full depth; pre-push cannot narrow the import graph from a YAML change and collects this test instead of running it.
- `workflow-local-run` is unmeasured on the path its cap protects. The cap is back at 30m, which is the safe direction, not evidence.
- The workstation declared sum is 57.5 minutes against ADR-104's 300s target (#5318).
- ADR-054's enforced 900s `security-scan` budget is three times ADR-104's target and is not reconciled (#5318). The new aggregate is 840s, under both.
- `pre-pr-validation` is ~70% of the hook and does not route through `git_hook_policy`, so the watchdog does not cover it; it carries its own 240s cap. The duplication behind that cost has a cheap fix that is unsound (#5317).
- ADR-104 and its debate log carry declared `taste-lint: ignore file-size` exceptions, matching four other ADRs: rules and rounds are cited by number, so splitting breaks the references.

## Related Issues

Closes #5315. Refs #5317, #5318, #4710, #5066.


rjmurillo:
Seven review rounds are now addressed in code. Summary of what changed since the last batch, grouped by the finding that drove it, so the resolved threads are checkable rather than just closed.

## Rule 8 is closed for pre-push

Two findings said the 150s clamp bounds one child process, not one job. Both jobs that run several subprocesses now hold their own aggregate deadline:

- `run_pytest` clamps its aggregate, so the whole pytest step is bounded at 150s in a container on every path, including the ordinary import-graph subset (which takes the 780s execution budget, not the collection one; that was the part the PR text had wrong).
- `scan_pushed_heads` now holds one deadline across every pushed ref **and** every target batch within a ref. Before, each batch took a fresh allowance, so a container push cost refs times batches times 150s. Budget is `SEMGREP_TIMEOUT_SECONDS`, unchanged for a one-ref workstation push and still under ADR-054's enforced 900s.

Verified by mutation: removing the wiring and the between-refs check fails four of the seven new tests in `tests/validation/test_security_scan_budget.py`.

The timeout-diagnostic finding falls out of the same change: `budget` is now the clamped value, so the message and the child's actual limit agree.

## The full-suite opt-in did less than its name

`AI_AGENTS_PYTEST_FULL_SUITE_LOCALLY=1` on a Python change the import graph could narrow ran whatever subset the selector chose, silently. Hoisting the validator fixed the *invalid*-value case on both paths and left the *valid* one broken on one. The flag now short-circuits selection and says so on stderr; the branch inside `_full_suite_stand_in` that became unreachable is deleted rather than left as decoration.

## Three guards that could not fail

- The collection-contract check read the whole ADR while calling its surface "ADR-104 rule 5". The Alternatives table's "A broken import" would have masked a deletion from rule 5. It now extracts the numbered rule 5 block, with a control that doctors rule 5, leaves the masking occurrence in place, and requires the check to name the ADR surface.
- `DECLARED_BUDGET_BASELINE_SECONDS` was described as a ratchet while sitting twenty lines from the assertion reading it. The ratchet now recomputes from `lefthook.yml` at the base ref; the constant is documented as a ceiling for environments with no base ref.
- `CONTAINER_UNCLAMPED_JOBS` graded jobs by name, so a rename onto a direct command joined the clamped side silently. `job_cost` reads the job's `run:` string now, and a new test requires the roster and the derived set to agree in both directions. No number moved (14 clamped, 20 not); what changed is whether they can stop matching quietly.

## The CI filter, audited instead of curated

Four rounds each found more roots by reading tests. A reverse audit was run: parse every string literal in all 974 test modules, keep the ones resolving to a tracked file, check each against the filter. **120 candidate inputs were unmatched**, 92 Markdown, 14 JSON, 5 text, spanning every tree including `.agents/retrospective/**`, `.agents/sessions/**`, and `.agents/analysis/**`, which the filter's own comment names as the Markdown it deliberately skips. Two session JSONs are opened by exact path in `tests/test_validate_session_json.py`.

Three extension globs plus five extensionless names close all 120. The trade was the repo owner's call: a docs-only PR now runs the matrix it used to skip, so the failure direction flips from a gate silently not running to a run being wasted.

One correction to that number: 120 is a superset, not a census. The audit resolves string literals against the tree, so a test building `tmp_path / "STEP-0.5-METRICS.md"` matches the tracked file of that name. It is evidence the allowlist could not be finished by hand, not a count of real dependencies.

## One cap went back up

`workflow-local-run` was cut from 30m to 10m (child budget 1740s to 540s) with no measurement, which ADR-104 rule 7, written on this branch, forbids. A push carrying a workflow change did fire it, at 0.42s, but this container has no `actionlint`, so the job reports DEGRADED before doing any work; that is not the path the cap protects. Both values are back at main's, taking the declared workstation sum from 2850s to 3450s. The base-ref ratchet still holds, because the base ref declares 4170s.

## Also

Two claimed collection misses (missing fixture, duplicate test-function names) had never been executed, only string-compared between surfaces that agree with each other. Both now run against real trees under production `addopts`, each with a syntax-error control proving collection reached the fixture. Four comments describing the pre-clamp world are retired. The stale `pre_pr_sequence.py:147` citation is now `_root_only` by name (line 162), with the signature at column 0 and byte-exact at 96 characters. Both pytest flags have config-catalog rows.

## Still open, and named as such

- `push-ref-policy` carries a 2m cap over `check_push_refs`, which runs many git children at 90s each. Same aggregate-versus-child shape as the two just closed; raised in round 8, still unverified.
- The base-ref budget ratchet skips where no base ref is reachable (a shallow CI checkout). The ceiling is the only guard there.
- `workflow-local-run` is unmeasured on the path its cap protects. Back at 30m is the safe direction, not evidence.

---

Issue #5241, opened 2026-08-21
Implement ADR-100: retire the PR size ceilings across all enforcement sites


rjmurillo:
ADR-100 is accepted. Its Decision names five items across four enforcement sites, with two ordering couplings that make the obvious decomposition wrong. This issue is the work list.

**Nothing here has landed.** `BLOCK_THRESHOLD = 20` and `MAIN_MERGE_BLOCK_THRESHOLD = 40` are still live in `scripts/validation/pr_commit_count.py:65,71` on `main`, and `_check_commit_limit` still blocks pre-push. Accepting the ADR changed no behavior.

## The two orderings that constrain everything

These are the reason this cannot simply be split into five parallel tasks.

1. **Both commit-ceiling sites demote together.** The ADR: *"Demote it to a report alongside the CI site, or the ceiling survives in the place it does the most damage."* The local pre-push gate is the one contributors actually hit, because it blocks the push rather than the merge. Retiring only the CI site retires nothing.
2. **Item 3 lands before or with item 4.** The ADR on item 4: *"Removing it before item 3 would tighten the gate rather than retire it."* Deleting `SKIP_SCOPE_CHECK` while the scope check still blocks leaves a blocking gate with no relief at all, which is strictly worse than today.

### 1. Both commit-ceiling enforcement sites stop blocking

| Site | File | Behavior |
|---|---|---|
| CI block | `scripts/ci/enforce_pr_validation.py:64-84` | reads `COMMIT_STATUS`, fetches labels, returns `LOGIC_ERROR` unless `commit-limit-bypass` present |
| Local pre-push | `scripts/validation/git_hook_policy.py:6122-6162` | `_check_commit_limit`, reached from `push-ref-policy` |

- Remove the CI blocking branch and its bypass-label handling. Keep the classifier's notice and warning output.
- `scripts/validation/pr_commit_count.py` already returns 0 on every classification path (`:493`, `:519`). It still exits 2 on argument or lookup error and 3 on `RuntimeError` (`:461`, `:480`, `:483`), so do not "simplify" it to always-0.
- Demote `_check_commit_limit` to a report in the same change. Note it counts differently: `git rev-list --count` includes merge commits, where the CI classifier counts authored non-merge only.
- Retire the `commit-limit-bypass` label and its human step with both sites.

[...]

Issue #5168, opened 2026-08-19
SessionStart context loader injects a stale, deprecated HANDOFF.md every session


rjmurillo:
## What

`.claude/hooks/SessionStart/invoke_context_loader.py` auto-loads `.agents/HANDOFF.md` (up to 4,000 chars) into context on every session start, alongside the latest retrospective.

## Why this is negative ROI

`.agents/HANDOFF.md` has carried this banner since 2025-12-22 and is unchanged since:

```
**Status**: 🟢 ACTIVE - Read-only reference (as of 2025-12-22)
...
**This file is now READ-ONLY.** Do not update during sessions.
```

Its own dashboard lists PRs #222, #212, #147, #89, #94, #95, #76, #93 as active — all from December 2025, long since resolved. ADR-014 (Distributed Handoff Architecture) superseded it with per-issue handoffs and Serena memory. The mandatory session-log/HANDOFF protocol that this hook's origin ADR-008 was designed to support was itself retired 2026-08-17 in PR #5135, whose own PR body calls it "negative-ROI workflow and session evidence machinery."

Net effect: every session pays ~1,000+ tokens to load an 8-month-stale dashboard that its own maintainers marked dead, with no compensating value. The retrospective half of the same hook is not in question — same-day retrospectives loaded during evaluation surfaced real, current findings.

## Fix

Stop loading `.agents/HANDOFF.md` in `invoke_context_loader.py`. Keep the latest-retrospective auto-load and the pending-skeleton reminder, which are the parts earning their token cost.

## Evidence

Found while auditing session-start/session-end hook ROI against this repo's own retrospectives and PR history (2026-08-19 session). Full trace: `.claude/hooks/SessionStart/invoke_context_loader.py:301-320` (pre-fix), `.agents/HANDOFF.md:1-15`, `e6c196d` (PR #5135), `.agents/architecture/ADR-008-protocol-automation-lifecycle-hooks.md`, `.agents/architecture/ADR-014-distributed-handoff-architecture.md`.

PR #5470, opened 2026-09-02
chore: retire the stale HANDOFF.md dashboard and its dead gates

rjmurillo:
### What

Delete `.agents/HANDOFF.md` and every live instruction that reads, keeps,
protects, or blocks on it.

### Why

The file was frozen read-only on 2025-12-22 and never updated since. Its
own dashboard still lists PRs #76 through #222 as pending. ADR-014
(distributed handoff architecture) superseded it, so this change executes
that ADR rather than making a new decision.

The token saving is small and worth stating honestly: 89 bytes of
always-on corpus, plus roughly 1,000 tokens per session for any agent
that obeyed `AGENTS.md`'s manual read instruction. The auto-load was
already removed by #5168. The real value is removing instructions that
contradict each other and a CI gate that guards nothing.

## Specification References

| **Issue** | Refs #5168 | removed the HANDOFF.md auto-load; this completes the retirement |
| **Issue** | Refs #5241 | PR size ceilings; this PR needed the documented scope bypass |

These references do not close their linked items.

## What was removed

- The file itself.
- `.github/workflows/validate-handoff-readonly.yml`, an entire CI
  workflow whose only job was to fail PRs that modify the file. Verified
  against live ruleset `11104075` that it was never a required check.
- The `handoff-protection` lefthook job and `check_handoff` in
  `git_hook_policy.py`.
- `check_handoff_modified` in `detect_hook_bypass.py`, which detected
  edits to a file that no longer exists.
- `universal.md` MUST NOT item 3, with every numbered citation of items
  4 through 7 repointed to 3 through 6 across scripts, skills, tests, and
  a Serena memory.
- The blocking gate in `implementer.shared.md` and its five mirrors that
  halted every implementer run when the file was missing. Three test
  files still asserted that gate; all eight assertions were failing on
  this branch and now assert the shipped contract.

## Testing

`uv run pytest tests/` reports 30,787 passed, 84 skipped, 0 failed.
`scripts/validation/pre_pr.py` reports "RESULT: All validations passed"
across its full 60-check run. `.agents/templates/HANDOFF.md`, a different
file and still the per-issue handoff template, is present and untouched.

## Two gates were bypassed, both with operator authorization

`SKIP_SCOPE_CHECK=1`: the branch changes 76 files, 50 after the generated
exclusion, against a 50-file hard limit. An atomic file removal cannot be
partially applied; a split leaves dangling references to a deleted file,
which is the exact state the gates reject.

`LEFTHOOK_EXCLUDE=security-scan`: `git_hook_policy.py:558` sets
`SEMGREP_TIMEOUT_SECONDS = 840`, then `_container_clamped` cuts it to 150s
in a managed container. Semgrep needs longer for this branch's 32
scannable files against 1,277 rules and failed closed at 149.65s,
reproduced twice with identical timing. The same scan runs on this PR in
CI, where the clamp does not apply, so the check is deferred rather than
skipped. Worth its own issue: the clamp's docstring says "a net that can
break the push it protects is worse than no net."

## Left alone, deliberately

Five ADRs and one spec still reference the file. Editing any `ADR-*.md`
fires the mandatory `adr-review` debate and falls under "Ask First:
Architecture." ADR-014 deserves a dated amendment; that is a separate,
deliberate action.

Session logs, archives, retrospectives, and analysis records keep their
references. They are dated accounts of what happened and read correctly
with the file gone.

Four governance docs and one test file were already over the 500-line
taste-lint ceiling before this change, so a one-line reference removal
was blocked by a pre-existing size violation. They carry scoped
`taste-lint: ignore file-size` markers with reasons rather than being
split here.


rjmurillo:
owner: local:pr-autofix
session: pr-autofix-20260902-a1
acquired_at: 2026-09-02T01:27:25Z
expires_at: 2026-09-02T01:42:25Z
base_sha: 60518a129a39df939c9f76378f090cd4526e9d20
claim_id: f42b8375792347d4a96c54f585702547

PR #5514, opened 2026-09-03
fix(tests): stop asserting deleted agent fixtures still exist

rjmurillo:
## Problem

`tests/eval/test_eval_suite_routing.py::test_shadowed_entrypoints_exist_in_the_tree` fails on current `main` (`90d7f7b17`):

```
FAILED tests/eval/test_eval_suite_routing.py::test_shadowed_entrypoints_exist_in_the_tree[.claude/agents/AGENTS.md] - AssertionError: .claude/agents/AGENTS.md is not in the tree
FAILED tests/eval/test_eval_suite_routing.py::test_shadowed_entrypoints_exist_in_the_tree[.claude/agents/CLAUDE.md] - AssertionError: .claude/agents/CLAUDE.md is not in the tree
```

Reproduced directly against a clean `origin/main` worktree, no PR involved:

```
$ uv run --frozen pytest tests/eval/test_eval_suite_routing.py -q -k test_shadowed_entrypoints_exist_in_the_tree
2 failed, 9 passed
```

## Root cause

PR #5495 deleted `.claude/agents/AGENTS.md` and `.claude/agents/CLAUDE.md` (frontmatter-less files the plugin loader was silently registering as dispatchable agents) but did not update `SHADOWED_ENTRYPOINTS` in `test_eval_suite_routing.py`, which still names both paths and asserts they're real files.

First noticed as a red `pytest (bulk-nested)` / `Run Python Tests` check on PR #5343, which merges `origin/main`. Confirmed it's inherited (not caused by that PR) by running the identical test against `origin/main` directly, unmodified.

## Fix

Split `SHADOWED_ENTRYPOINTS` into two lists:

- `SHADOWED_ENTRYPOINTS` (unchanged, still names both deleted paths): feeds `test_entrypoints_inside_prompt_and_skill_trees_are_not_shadowed`, which only calls `classify_path` on the string, no filesystem I/O. The routing assertion (an entrypoint filename wins over the `.claude/agents/` prefix rule) is still real coverage: if either name reappears under `.claude/agents/` by accident, this must still classify it as `entrypoints`, not `agents` or `other`, keeping it off the plugin loader's dispatchable-agent scan.
- `SHADOWED_ENTRYPOINTS_ON_DISK` (new): the same list minus the two #5495 deleted, feeds the negative control `test_shadowed_entrypoints_exist_in_the_tree`. Asserting the deleted paths are files would fail for a reason unrelated to routing, and reintroducing either as a real fixture here would put it back on the plugin loader's scan, the exact defect #5495 removed.

## Testing

- `uv run --frozen pytest tests/eval/test_eval_suite_routing.py -q`: 278 passed (was 280 collected, 2 failed, before the fix).
- Negative control: reverted only the negative-control test's parametrize target back to the original list, confirmed the exact same 2 failures reproduce, restored the fix, confirmed clean again.
- `uv run --frozen ruff check tests/eval/test_eval_suite_routing.py`: clean.
- `uv run --frozen python scripts/validation/pre_pr.py`: all validations passed.

Refs #5495

https://claude.ai/code/session_01JY5yjyTcPA9ND487HrQRYW

---
_Generated by [Claude Code](https://claude.ai/code/session_01JY5yjyTcPA9ND487HrQRYW)_

### DIFF
--- a/scripts/validation/check_git_hook_health.py
+++ b/scripts/validation/check_git_hook_health.py
@@ -54,6 +54,7 @@ An explicit non-repository is out of scope and exits 0. Missing git, timeouts,
 and unexpected command failures are verification failures and exit 3; the
 pre-PR adapter receives False for the same states.
 """
+# ruff: noqa: E402
 
 from __future__ import annotations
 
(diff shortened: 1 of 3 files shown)


rjmurillo:
CI red on the previous head (`7e26de824`): `pytest (mutation)` failed with `ModuleNotFoundError: No module named 'lefthook_inventory'` inside the mutation harness's isolated worktree copy.

Not this PR's regression: `check_git_hook_health.py`'s bare `from lefthook_inventory import (...)` only resolved by riding a `sys.path` side effect from another file, which breaks whenever `scripts.validation.check_git_hook_health` is imported as a package member with nothing having primed the path first (`tests/test_lefthook_integration.py` does this). Same root cause already diagnosed and fixed in PR #5512, and separately hit and ported into PR #5343.

Ported the fix into this branch now (`b2ac837ab`) rather than waiting for #5512 to merge, since it no-ops once main carries it.

---

Issue #5709, opened 2026-09-10
retro(policy): trigger-based retrospectives, reflect at session stop, retire pre-push gate


rjmurillo:
Part of epic #5698. This issue is self-contained; read only this issue.

## Review corrections (takes precedence over stale mechanism and causal language below)

- The `96` retrospective count is a historical filesystem snapshot, not proof that the pre-push gate caused low-value work. Preserve the date, path filter, and exclusions. The owner direction is a proposal to test, not independent evidence.
- The current `reflect` skill requires showing proposed learnings and awaiting explicit user approval before persistence. A Stop hook may nudge, but it must not force memory writes or trap a session in a block loop. “No marker” and “marker read failure” need distinct, tested behavior; fail-open means the hook cannot claim that reflect ran.
- Prove the actual Stop invocation contract and a session-scoped marker producer before changing `lefthook.yml` or adding a blocking hook. Test normal completion, repeated Stop events, interrupted sessions, absent marker, malformed marker, and read/write failure. A hook that blocks every session because no producer exists is a regression, not a nudge.
- Separate the reversible probe from removal of the retrospective gate. First measure whether trigger-based reflection changes retrospective volume, accepted-task outcomes, correction time, and user-visible friction. Remove the push gate only after the replacement path is live, observed, and does not deadlock or silently lose findings.
- The proposed “comment and close #5301” step is an external side effect. It requires current issue-state verification and owner authorization; it is not an automatic consequence of implementing this issue. Do not close a historical issue merely because a new design makes its premise obsolete.
- Do not add a new issue or control for every stale skeleton/reference discovered. Record findings in the implementation PR unless the owner selects them.

## Context (why this exists, for a reader with no history)

Retrospectives in this repo are push-driven today: a pre-push gate refuses to push unless a retrospective file dated today exists (or the session log carries retrospective evidence, or the session is trivial, or the escape hatch is set). That produces one retrospective-shaped artifact per push-worthy session regardless of whether anything notable happened. Measured on 2026-09-10: 96 files under `.agents/retrospective/` dated 2026-08 or 2026-09 (`ls .agents/retrospective/ | grep -E '^2026-(08|09)' | wc -l`).

[...]

Issue #5703, opened 2026-09-10
agents(retrospective): stop filing GitHub issues from retrospectives, keep findings in retro file


rjmurillo:
Part of epic #5698. This issue is self-contained; read only this issue.

## Review corrections (takes precedence over stale text below)

- Update all six retrospective agent surfaces, including hand-maintained `.github/agents/retrospective.agent.md`; it is not regenerated by `build/generate_agents.py` and currently contains the old routing.
- Remove the preceding `Backlog Candidates` table and every issue/backlog routing instruction, not only the two named creation blocks. Preserve unrelated learning-memory persistence.
- Emit one `Findings for the owner` table with `Item`, `Evidence (path:line)`, `Proposed action`, and `Class`. Determine `Class` from the frozen completion contract, not mechanically from P0/P1 severity. A `Blocker` names the falsified criterion and remains blocking; recording it does not permit declaring completion. Optional findings do not reopen completed work. An empty table is valid.
- Retrospectives recommend; the owner decides what becomes tracked work. This change removes an autonomous filing path; the existing backlog counts do not independently prove that retrospectives caused the entire backlog increase.
- Test all six agent surfaces, all skill mirrors, empty findings, all four classes, preserved learning persistence, and absence of issue/backlog routing. Require install parity as well as generator validation.

## Context (why this exists, for a reader with no history)

The retrospective workflow currently tells an agent to file a GitHub issue for
every P0/P1 Delta item it finds while closing out a retrospective, and to
write P2/P3 items to a separate backlog memory file. This instruction is
duplicated across two independent surfaces that both ship to agents:

1. The agent template family: `templates/agents/retrospective.shared.md`
   (canonical) plus two hand-maintained copies, `src/claude/retrospective.md`
   and `.claude/agents/retrospective.md`.
2. The retrospective skill: `.claude/skills/retrospective/SKILL.md`, its
   `references/frameworks.md` and `references/learning-template.md`, and the
   generator script `references/../scripts/run_retrospective.py` that renders
   an artifact matching that template byte for byte.

Both surfaces run the same protocol: classify a Delta item's priority, then
for P0/P1 call the `github` skill (`new_issue.py`) immediately, and for P2/P3
write a Serena memory file under `backlog/`. Neither step asks a person
first. The repository has run dozens of retrospectives in the last month and
a half, each one capable of filing multiple P0/P1 issues on its own judgment,
and the open-issue backlog grew by several hundred issues in the same window.
[...]

Issue #5684, opened 2026-09-10
check_agent_skill_discriminator hard-fails on every agent PR; its c1/c3 corpus was deleted by ADR-064


rjmurillo:
## Summary

`scripts/validation/check_agent_skill_discriminator.py` fails with exit 2 on every PR
that touches an agent definition. The guard at line 853 requires `.claude/commands/`
to exist, but PR #5640 (`72f0e63`) deleted that tree when it resumed the ADR-064
commands-to-skills migration. The gate has been unpassable for agent PRs since.

Observed on PR #5682 and PR #5681, both red on `Score changed agents`:

```
Commands directory not found: /home/runner/work/ai-agents/ai-agents/.claude/commands (cannot score c1/c3).
##[error]Process completed with exit code 2.
```

Reproduces on a clean `main` checkout:

```
$ python3 scripts/validation/check_agent_skill_discriminator.py --changed-files ".claude/agents/analyst.md"
Commands directory not found: /home/user/ai-agents/.claude/commands (cannot score c1/c3).
$ echo $?
2
```

## Why deleting the guard is the wrong fix

The guard is not incidental. Two of the discriminator's three CI-computable criteria
read the command corpus as their only input:

- c1: agent invoked from a slash command via `Task(subagent_type="<name>")`, searched
  across `.claude/commands/` and `templates/commands/`.
- c3: defined as N/A, scoring 0, whenever c1 is false.

Both command trees are gone. `_command_files()` already tolerates their absence and
returns an empty list, so with the guard removed the gate still runs, but c1 and c3
are structurally 0 for every agent forever.

Measured by calling `run_check()` directly against every agent file on `main`
(31 under `.claude/agents/`, 31 under `templates/agents/`, 62 total):

| Criterion | Agents scoring true |
|---|---|
| c1 | 0 / 62 |
| c2 | 27 / 62 |
| c3 | 0 / 62 |

Score distribution: `{0: 35, 1: 27}`. Maximum observed score is 1. The gate fails an
agent at score >= 2, so **no agent can fail it under any input**.

So the two obvious repairs are both wrong in opposite directions:

- Delete the guard -> the gate passes unconditionally and reports success while
  measuring nothing. That is the silent pass `ci-scripts.md` MUST-11 forbids.
- Keep the guard -> agent PRs stay permanently red with no author-side remedy.

## Options

Recording these for the decision, not choosing between them. Re-specifying c1 and c3
changes a discriminator locked by the Issue #2003 audit
[...]

Issue #5301, opened 2026-08-25
docs: clarify retrospective-policy's per-branch date-window requirement for worktree-driven pushes


rjmurillo:
## What

`scripts/validation/git_hook_policy.py:check_retrospective_evidence` (the `retrospective-policy` pre-push lefthook job) requires a retrospective dated today (or yesterday, across timezones) present in `.agents/retrospective/` on the branch being pushed. When a session drives a push from a different worktree/branch than the one that produced its own same-day retrospective (for example: a session working primarily on branch A, that also pushes fixes to a sibling branch B via an external worktree), branch B's working tree has no automatic way to satisfy this check unless a same-day retrospective file also exists there.

This is documented, expected behavior given `.claude/rules/retros.md`, not a defect in the gate. It is easy to miss because the failure mode looks identical to "no retrospective was written this session" even when one genuinely was, just on a different branch's tree.

## Where found

`.agents/retrospective/2026-08-25-pr5209-push-notification-false-completion.md`, Finding 1, remediation item 3 (originally left unowned per `.claude/rules/retros.md:17`'s requirement that every remediation action have an owner or a tracking issue; this issue is that owner).

## Proposed remediation

Something in this space, to be scoped by whoever picks this up:

- Document the per-branch date-window requirement explicitly in `.claude/rules/retros.md` or `.claude/rules/session-logs.md`, so a session driving multi-branch work knows in advance that each pushed branch needs its own same-day retrospective evidence, rather than discovering it via a failed push.
- Consider whether `check_retrospective_evidence` should also accept evidence from a *different* branch's same-day retrospective when the session can prove common authorship (for example, matching git author/committer identity within a bounded time window), though this weakens the gate's guarantee and needs its own scoping.
- At minimum, add a clearer error message pointing at this documented behavior when the gate fires, so the next session does not have to read `git_hook_policy.py`'s source to learn the exact date-window and per-branch conditions, as the originating session had to.

## References

- `scripts/validation/git_hook_policy.py:check_retrospective_evidence`, `_today_retrospective_exists`
- `.claude/rules/retros.md`
- `.agents/retrospective/2026-08-25-pr5209-push-notification-false-completion.md` (the retrospective this issue owns the remediation item from)

PR #5803, opened 2026-09-16
refactor(hooks): retire the per-day retrospective gate, keep the trigger-based ones

rjmurillo:
# Pull Request

### What

Deletes the `retrospective-policy` pre-push gate and the machinery only it used, then replaces every calendar-shaped retrospective trigger in the repo with the four event triggers from #5709.

### Why

The gate refused `git push` unless `.agents/retrospective/` held a file dated today or yesterday, or the branch's session log mentioned a retrospective. That made a retrospective a tax on pushing rather than a response to an incident. On a day when nothing notable happened, the artifact it forced recorded nothing and cost every push in the session.

The retrospective agent's trigger list said the same thing from the other side: "Agent produces output", "Task completes", "User provides feedback", "Session ends", "Milestone reached". Four of those five match nearly any turn.

Owner direction, this session: retrospectives are warranted circumstantially, not every day regardless of what happened.

## Specification References

| **Issue** | Refs #5709 | retro(policy): trigger-based retrospectives, reflect at session stop, retire pre-push gate |
| **Issue** | Refs #5301 | docs: clarify retrospective-policy's per-branch date-window requirement (now moot; see Noticed, not done) |
| **Issue** | Refs #5684 | check_agent_skill_discriminator hard-fails on every agent PR (a red check here; see Pre-existing CI failures) |

These references do not close their linked items: #5709 also asks for a `Stop` hook that nudges `reflect` at session end, which this PR does not build (see Deviations). #5301 needs owner authorization to close, per #5709's own review corrections. #5684 is a pre-existing bug this PR only surfaces.

## Changes

**Gate removal**

- `lefthook.yml`: `retrospective-policy` job deleted from the piped stdin group.
- `scripts/validation/git_hook_policy.py`: `check_retrospective_evidence`, `_session_has_retrospective_evidence`, `_today_retrospective_exists`, `_is_trivial_retrospective_session`, `_documentation_only`, `_recent_date_prefixes`, `RETROSPECTIVE_EVIDENCE_PATTERNS`, `DOCUMENTATION_PATTERNS`, `TRIVIAL_SESSION_SECONDS`, `_handle_retrospective`, the `retrospective` subparser row, and `SKIP_RETROSPECTIVE_GATE` in `run_pytest`'s stripped-env tuple. Confirmed no other caller for each before deleting; `_current_host_date_prefixes` stays because `_recent_session_candidates` still reads it.
- `_session_log_for_current_branch` stays. `check_adr_review_policy` and `check_branch_context` still call it.
- Tests: retrospective-gate tests removed from `tests/test_lefthook_integration.py`, `tests/validation/test_session_log_branch_aware.py` (the ADR-policy half of that file is untouched), `tests/ci/test_lefthook_prepush_fast_fail.py`, and the `#4194 retro-uses-branch-log` mutant plus its inert control from `tests/mutation/test_worktree_path_mutations.py`.

**Trigger lists**

The four triggers: a HIGH-confidence `reflect` finding, `main` going red, a reverted or force-fixed merge, an explicit user correction. The Post-PR Retrospective workflow is a fifth path and fires on its own at PR close.

- `templates/agents/partials/retrospective-claude-code-tools.mustache` (the partial the rendered agents actually read) and `templates/agents/retrospective.shared.md`, regenerated into `src/claude/agents/retrospective.md`, `src/copilot-cli/agents/retrospective.agent.md`, `src/vs-code-agents/retrospective.agent.md`, `.claude/agents/retrospective.md`, `.github/agents/retrospective.agent.md`.
- `templates/rules/retros.md`: new "When To Write One" section; mirrors regenerated into `.claude/rules/retros.md`, `src/claude/rules/retros.md`, `.github/instructions/retros.instructions.md`, `src/copilot-cli/instructions/retros.instructions.md`.
- `templates/skills/retrospective.SKILL.md.tmpl`: When-to-Use table now answers No to "a session ended and nothing above fired", and no longer claims a Stop hook produces auto-retro skeletons (that hook was deleted in #3349). Rendered into `.claude/skills/retrospective/SKILL.md`, `src/claude/skills/retrospective/SKILL.md`, `src/copilot-cli/skills/retrospective/SKILL.md`.
- `templates/skills/ai-agents-docs-of-record.SKILL.md.tmpl`: retrospective write-path column names the triggers instead of leading with the dead skeleton route. Rendered into `.claude/skills/ai-agents-docs-of-record/SKILL.md`, `src/claude/skills/ai-agents-docs-of-record/SKILL.md`, `src/copilot-cli/skills/ai-agents-docs-of-record/SKILL.md`.
- `CLAUDE.md`: the "Weekly retro" routing line was the last calendar cadence in the routing rules; it now reads `- Retro trigger -> invoke retrospective skill`.
- `.agents/governance/FAIL-OPEN-INVENTORY.md`: `SKIP_RETROSPECTIVE_GATE` moves UNDECIDED to RESOLVED by removal, matching the `SKIP_SCOPE_CHECK` precedent.
- `.serena/memories/git/git-merging-main-forfeits-the-docs-only-push-bypass.md`: banner marks the bypass half history. The file stays for its still-current half (a parallel lefthook group bills the slowest job even when a cheap gate refuses first, `gh pr update-branch`, the `HEAD^1` reset recipe and its merge-tree proof, the ruleset-vs-classic-protection 404).
- `docs/agent-catalog.md`: regenerated token count for the retrospective agent.

**Fallout the first push surfaced**

Three follow-up commits, each fixing something the full suite caught that the pre-push changed-scope selector did not:

- `tests/build_scripts/fixtures/agent_templates_b1/expected/claude/retrospective.md` and `tests/build_scripts/fixtures/agent_templates_b1/expected/copilot_source/retrospective.md`: byte-pinned renders, stale once the partial changed.
- `tests/build_scripts/fixtures/rule_templates_b2/expected/retros.md`: same, for the rule render.
- `CLAUDE.md` again: the first replacement line spelled out all four triggers at 206 bytes against the old line's 40, which broke the 3000 byte per-file ceiling (3154) and the 6600 byte workspace total (6648). The file sits at 2988 bytes on main, so there were 12 bytes of headroom. Cut to 47 bytes, landing at 2995.

Both fixture repins went through `skill_template_grammar.render` with the real partials directory, the same call path each test uses, so the pin still proves losslessness rather than restating the test's own output.

**Issue #5684: the discriminator's c1/c3 corpus**

- `scripts/validation/check_agent_skill_discriminator.py`: new `PIPELINE_SOURCES` table and `_pipeline_dirs`; `_command_files` reads every pipeline tree; the `main()` guard refuses on an empty corpus instead of on one named directory; docstring updated for what c1 and c3 now read.
- `tests/validation/test_agent_skill_discriminator_pipeline_sources.py`: six cases, in a new sibling file because the parent sits at 491 of the 500-line ceiling. All six fail against the unfixed script and pass against the fixed one; the 83 existing discriminator tests are unchanged and green.

## Acceptance criteria

- [x] `grep -n "retrospective-policy" lefthook.yml` returns nothing
- [x] `grep -n "check_retrospective_evidence" scripts/validation/git_hook_policy.py` returns nothing
- [x] `grep -n "SKIP_RETROSPECTIVE_GATE" scripts/validation/git_hook_policy.py` returns nothing
- [x] `grep -n "Weekly retro" CLAUDE.md` returns nothing
- [x] `grep -n "SKIP_RETROSPECTIVE_GATE" .claude/skills/ai-agents-config-catalog/SKILL.md` returns nothing (true before and after)
- [x] The retrospective agent's Trigger Conditions list the four triggers with no calendar or standalone "Session ends" entry, in the template, the partial, and all four rendered trees
- [x] A push on this branch, which carries no retrospective file dated today, completed with every pre-push gate green
- [x] No U+2014 or U+2013 in any touched file

## Type of Change

- [ ] Bug fix (non-breaking change fixing an issue)
- [x] Documentation update
- [x] Refactoring (no functional changes)

## Reviewer Expectations

**Review kind / scrutiny / urgency**: design review, high scrutiny, no rush. This retires an enforcement gate.

**Focus areas**:

1. Whether retiring the push gate before the replacement `reflect`-at-session-stop path exists is acceptable (see Deviations).
2. `scripts/validation/git_hook_policy.py`: whether any deleted helper had a caller I missed.
3. Whether the four triggers are the right four.

## Testing

Deliverables in this PR:

- [x] Tests added/updated (removed, with no dangling reference to a deleted symbol)

Evidence:

- `uv run --frozen pytest tests/ -q`: 33895 passed, 120 skipped, 3 failed at 0:21:51, the three being the two workspace-budget tests and their enforcer wrapper. All three are fixed by the CLAUDE.md size commit; `tests/test_workspace_limits.py tests/test_validate_workspace_budget.py` is 35 passed after it. Earlier full runs surfaced the two stale fixtures, one per run.
- `uv run --frozen pytest tests/mutation -q`: 100 passed (the harness needs a clean tree, so this ran after the first commit).
- `uv run --frozen python scripts/validation/pre_pr.py`: no FAIL. Two BLOCKED, both `reason=tool.absent`: actionlint and yamllint are not on PATH in this container, so neither examined a file. The `lefthook.yml` edit is therefore not yamllint-verified here; `tests/test_lefthook_integration.py` does parse and assert against the file.
- `git push`: every pre-push gate green on both pushes, including pre-pr-validation, security-scan, dash-prohibition, count-ratchets, and zero-collection-tests. The first push is itself the acceptance criterion that a branch with no retrospective dated today can push.

## Issue #5684, fixed here rather than worked around

`Score changed agents` was red on this PR and on every PR touching an agent file. `check_agent_skill_discriminator.py` scored c1 and c3 by reading `.claude/commands/` and `templates/commands/`, then exited 2 when the first was absent. ADR-064 deleted both trees and `pre_pr.py`'s "Commands Retired (ADR-064)" gate keeps them deleted, so the check refused before scoring anything. Reproduced on a pristine `origin/main` checkout: same message, exit 2.

The corpus now names skills first, which is where ADR-064 moved user-invocable pipelines. Only `SKILL.md` counts, so a skill's `references/*.md` cannot invent a caller from a documented sample invocation. An empty corpus still exits 2 rather than skipping c1/c3: c2 alone maxes out at 1 against a threshold of 2, so a degrading run could not fail whatever it was handed, which is the ADR-109 B1 install-parity failure mode and worse than the red check it would replace.

`Score changed agents` is green on the current head.

### What the repaired gate now sees, per ci-scripts.md MUST-13

Changed-files mode, which is what runs on a PR, scores this PR's agent 0/3 and exits 0.

Full-corpus `--all` mode, which runs only on the scheduled trigger, exits 1 naming five agents the gate could not report while it was refusing:

```
FAIL: the following agents are skill-shape candidates (score 2+) (7 of 62 scored):
  - architect (score 2/3: c1=Y c2=Y c3=n)        templates/agents/architect.shared.md
  - devops (score 2/3: c1=Y c2=Y c3=n)           .claude/agents/devops.md
  - milestone-planner (score 3/3: c1=Y c2=Y c3=Y) both copies
  - qa (score 2/3: c1=Y c2=n c3=Y)               templates/agents/qa.shared.md
  - task-decomposer (score 3/3: c1=Y c2=Y c3=Y)  both copies
```

Seven files, five agents. These are pre-existing conditions, not regressions from this diff, and each has the gate's documented escapes (refactor to a skill, `isolation_required: true` with a rationale, or a PR-body token). Two are a template-vs-plugin drift bug worth separating: `architect` and `qa` carry `isolation_required: true` in `.claude/agents/` but not in their `templates/agents/*.shared.md` source, so the escape hatch would be dropped the next time those agents regenerate.

Triaging the five is outside this PR's scope and is the owner's call. Two options, neither taken here: land as-is, which leaves the scheduled run red and any PR touching those seven files blocked until they are triaged; or record a `--update-baseline` file and wire `--baseline` into the scheduled run, which is what issue #4087 built the ratchet for and would grandfather the five while still failing new regressions.

**AI Spec Validation** reports `CRITICAL_FAIL` on both rows across four runs with the same body: "Copilot CLI infrastructure failure after 3 attempts (exit code 1). Check COPILOT_GITHUB_TOKEN scope, rate limits, or network connectivity." Four identical results is not transient, but the remedy is token scope or rate limits, which I cannot reach.

**Codex code review** and **Cursor Bugbot** both posted usage-limit notices instead of reviews. Neither is a finding.

### Security Review

> Required for: Authentication, authorization, CI/CD, git hooks, secrets, infrastructure

Not a separate agent pass: the security-relevant surface here is the deletion of a local advisory gate and its opt-out flag, and the `security-scan` and `security-suppression-policy` pre-push gates both passed on each pushed head. The gate never gated a security property; it gated whether a Markdown file existed.

- `lefthook.yml`
- `scripts/validation/git_hook_policy.py`

### Other Agent Reviews

- [ ] Critic validated implementation plan
- [ ] QA verified test coverage

## Author Pre-flight

- [x] Builds locally (`uv run python build/scripts/build_all.py`, `uv run python build/generate_agents.py`)
- [x] Pre-PR validation passed (`uv run python scripts/validation/pre_pr.py`)
- [x] Lint and formatting clean (`uv run --frozen ruff check scripts/ tests/`)
- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Comments added only where the why is non-obvious
- [x] Documentation updated

## Deviations from #5709

1. **No `Stop` hook.** #5709 step 1 adds a Stop-hook script that nudges `reflect`, and its review corrections say to land that, measure, and only then retire the push gate. This PR does the removal without it. The owner asked for the removal directly this session, and the Stop hook would add a per-session blocking check, which is the shape of the thing being removed. It remains open work under #5709.
2. **#5301 not closed.** #5709 step 9 says comment and close. Its review corrections say that is an external side effect needing current issue-state verification and owner authorization. Left open.
3. **#5703 and #5288 untouched**, as #5709 instructs.

## Noticed, not done

- The retrospective skill template still describes filling an auto-retro skeleton further down, around its artifact-path and completion-banner paragraphs. That is correct for the historical skeletons already on disk, so it is left as is; only the claim that a Stop hook *produces* them was wrong, and that is fixed.
- The skill-markdown vendor-portability checker reports the always-on skills doc at 2 refs against a baseline of 3, so a `--update-baseline` run would tighten the pin. Pre-existing and untouched by this PR.
- `merge-resolver` shows 20.3% similarity between the two agent trees in Agent Drift Detection. Pre-existing, advisory, unrelated to this change.
- This clone arrived shallow, which blocks the `push-ref-staleness` pre-push job. Fixed locally with `git fetch --unshallow origin`. Worth knowing if other cloud sessions hit the same wall.
- The pre-push `python-tests` job runs a changed-scope selector, not the full suite, so it stayed green across three separate breakages: two stale byte-pinned fixtures and a per-file byte ceiling. All three were caught only by a full local run. That is a real gap between what a contributor's push proves and what CI then rejects. Not filed as an issue, per #5709's instruction to record findings here instead.
- The per-file ceiling on the root instruction file had 12 bytes of headroom before this PR. Any future edit to it will hit the same wall.

## Related Issues

Refs #5709, Refs #5301, Refs #5684, Refs #3349

---

https://claude.ai/code/session_01R9Q27vZQdV6ZBtc4daJJPE

---


rjmurillo:
## CI status on `d83c899`

Four red checks. Two are mine and are fixed locally; two are not this PR's.

### Mine, fixed locally, pushing shortly

`pytest (bulk)`, `pytest (bulk-nested)`, `Run Python Tests`

Two byte-pinned lossless-render fixtures went stale because this PR edits the templates they pin:

- `tests/build_scripts/test_agent_templates_lossless.py` pins the rendered retrospective agent, and this PR edits `templates/agents/partials/retrospective-claude-code-tools.mustache`.
- `tests/build_scripts/test_rule_templates_lossless.py` pins the rendered `retros` rule, and this PR adds a section to `templates/rules/retros.md`.

Both regenerated through `skill_template_grammar.render` with the real partials directory, the same call path each test uses, so the pin still proves losslessness rather than restating the test's own output. Reproduced locally at this exact head before fixing, and `tests/build_scripts` is green after: 2323 passed, 2 skipped.

Worth noting for anyone who hits this: the pre-push `python-tests` job runs a changed-scope selector, not the full suite, so it was green while both fixtures were stale. Only a full local `pytest tests/` caught them.

### Not this PR's

**`Score changed agents`** is issue #5684. `check_agent_skill_discriminator.py` hard-exits 2 when `.claude/commands/` is absent, and ADR-064 deleted that directory. `pre_pr.py`'s "Commands Retired (ADR-064)" gate actively enforces that it stays gone, so the corpus the check needs for c1/c3 cannot come back. Neither the script nor `.github/workflows/agent-skill-discriminator-check.yml` is touched by this diff (`git diff origin/main --stat` over both is empty); this PR only trips it by changing an agent file, which is what the workflow keys on.

Reproduced on a pristine `origin/main` checkout:

```
$ CHANGED_FILES=".claude/agents/retrospective.md" python3 scripts/validation/check_agent_skill_discriminator.py
Commands directory not found: <root>/.claude/commands (cannot score c1/c3).
EXIT_ON_MAIN=2
```

No fix PR exists to port, so nothing was ported and this PR was not widened to fix it. Proposed patch for #5684, not applied here: the `commands_dir` guard in `main()` should degrade instead of exiting 2, skipping c1 and c3 and scoring the remaining criteria, with the skipped ones named in the output so a partial score cannot be read as a full one. Exiting 2 was right while a commands corpus existed; after ADR-064 there is nothing to wait for.

**AI Spec Validation** reports `CRITICAL_FAIL` on both rows with the same body: "Copilot CLI infrastructure failure after 3 attempts (exit code 1). Check COPILOT_GITHUB_TOKEN scope, rate limits, or network connectivity." That is the workflow's own tooling failing to reach Copilot, not a verdict about this diff. I have no means to re-run it.

`Validate PR` flagged one path in the description that the diff does not contain; the body has been corrected and the rest resolves once the fixture commits land.

---

Issue #5804, opened 2026-09-16
ai-review Copilot model selection is ungoverned and not the economical tier


rjmurillo:
## Problem

`COPILOT_MODEL` reaches Copilot CLI from exactly one place: the `copilot-model` input on `.github/actions/ai-review/action.yml`, exported at lines 205 and 251 into the diagnose and invoke steps. Four workflows call that action:

| Workflow | Trigger | AI calls per run |
|---|---|---|
| `ai-spec-validation.yml` | every PR `opened`/`synchronize`/`reopened`/`edited` plus `merge_group` | 2 (analyst, critic) |
| `pr-maintenance.yml` | hourly cron, matrix over up to 20 PRs | up to 2 per PR |
| `ai-metrics-analysis.yml` | weekly cron | 1 |
| `artifact-insight-scanner.yml` | weekly cron | 1 |

Three of the four inherited the action default `auto`. One pinned `claude-sonnet-4.5`.

Two defects follow:

1. **`claude-sonnet-4.5` is not a model Copilot CLI serves.** The current CLI model list is Claude Haiku 4.5, Claude Sonnet 5, Claude Opus 4.8, Claude Opus 5, plus the GPT/Gemini/MAI/Kimi entries. Copilot answers an unserved id with `not available; using "<x>" instead` and exits 0 (measured in `.agents/analysis/2026-08-12-adr-080-copilot-model-resolution.md`), so the scanner has been silently billing at the session default's rate, not the model the pin names. Nothing in the pipeline notices.

2. **`auto` is not the economical choice.** Since GitHub moved Copilot to usage-based billing on 2026-06-01, cost is per-token at published rates. `auto` discounts the chosen model by 10 percent but routes freely across tiers, including GPT-5.4 at $2.50 in / $15.00 out per 1M tokens. Claude Haiku 4.5 is $1.00 / $5.00, so `auto`'s worst case costs roughly 2.7x a Haiku pin on the two highest-volume paths.

## Governance gap

ADR-080's `check_model_pins.py` scans unit frontmatter only (`_UNIT_GLOBS`). It does not read `copilot-model` values in `.github/workflows` or the action default, which is why the retired pin sat there unflagged.

## Acceptance criteria

- [ ] The action default is a model Copilot CLI currently serves, at the cheapest Anthropic tier.
- [ ] No workflow override names an unserved model.
- [ ] A model substitution at runtime surfaces as a workflow warning instead of passing silently.
[...]

PR #5808, opened 2026-09-16
ci(ai-review): default to the cheapest model Copilot serves and make a silent substitution loud

rjmurillo:
# Pull Request

### What

`COPILOT_MODEL` reaches Copilot CLI from one place: the `copilot-model` input on `.github/actions/ai-review/action.yml`. This changes that default from `auto` to `gpt-5.6-luna`, gives the driver a matching floor so an unset value cannot send `--model ""`, drops a `claude-sonnet-4.5` override that Copilot CLI no longer serves, warns when Copilot substitutes a model at runtime, and adds tests that fail when any of those regress.

### Why

Four workflows call the action. The two that dominate spend are `ai-spec-validation.yml` (every PR `opened`/`synchronize`/`reopened`/`edited` plus `merge_group`, two AI calls per run) and `pr-maintenance.yml` (hourly cron, matrix over up to 20 PRs, up to two calls per PR). Since Copilot moved to usage-based billing on 2026-06-01, cost is per-token at published model rates, so the selected model's rate is the whole bill.

Rates per 1M tokens, from GitHub's published pricing:

| Model | In / Out | vs the new default |
|---|---|---|
| GPT-5.6 Luna | $0.20 / $1.20 | baseline |
| MAI-Code-1.1-Flash | $0.20 / $1.20 | equal |
| Gemini 3.6 Flash | $0.75 / $3.75 | 3.75x / 3.1x pricier |
| Claude Haiku 4.5 | $1.00 / $5.00 | 5.0x / 4.2x pricier |
| GPT-5.4 (`auto`'s ceiling) | $2.50 / $15.00 | 12.5x pricier |

Three defects were live:

1. `artifact-insight-scanner.yml` pinned `claude-sonnet-4.5`. Copilot CLI no longer serves that id. Copilot answers an unserved id with `not available; using "<x>" instead` and exits 0 (measured in `.agents/analysis/2026-08-12-adr-080-copilot-model-resolution.md`), so that pin was buying the session default, not the model it named, and nothing reported it.
2. `auto` discounts the chosen model 10 percent but routes freely across tiers, up to GPT-5.4. Against the new default that is roughly 12x on the two high-volume paths.
3. The driver that actually builds the `copilot --agent … --model … --prompt …` argv had no floor on the env var. `parse_config` read `env.get("COPILOT_MODEL", "")`, so an unset or explicitly-empty value sent `--model ""` to the CLI. The line below it already floors its own input (`env.get("CONTEXT_MODE") or "summary"`). A composite action input passed explicitly-empty overrides the input's own default, so the YAML default was never a floor.

### Why the cheap tier is safe to take

The earlier draft of this PR stopped at the cheapest Anthropic model on a same-family argument. That argument does not survive inspection: `auto` already routes to GPT-5.4 against these same agent files, so a non-Anthropic model has been driving them without reported breakage. The open question was the small-and-fast tier, not the vendor.

Output-contract risk is bounded downstream rather than by model choice. `parse_ai_review_output.py` reads output with no `VERDICT:` line as `UNKNOWN`, `UNKNOWN` is in `BLOCKING_VERDICTS`, and the check goes red. A model that cannot hold the contract fails loudly.

What that does not catch is a well-formed verdict from a shallow review. That is the accepted trade, and it is recorded in both the action input description and the module comment so the next reader sees the cost, not just the saving.

## Specification References

| **Issue** | Refs #5804 | ai-review Copilot model selection is ungoverned and not the economical tier |

These references do not close their linked items: #5804 also covers the retry-on-quota follow-up noted below.

## Changes

- `.github/actions/ai-review/action.yml`: `copilot-model` default `auto` to `gpt-5.6-luna`; description now carries the rates, the `auto` trade-off, the currently-served id list, the fallback behavior, and the driver floor it mirrors.
- `.github/workflows/artifact-insight-scanner.yml`: drop the `copilot-model: claude-sonnet-4.5` override so the scanner inherits the new default.
- `scripts/ci/_copilot_model.py` (new): `DEFAULT_COPILOT_MODEL`, plus `MODEL_FALLBACK_PATTERN` and `report_model_fallback`, which emit a `::warning::` naming the requested and the served model. The underscore prefix keeps it out of the wiring gate's script glob, which requires a workflow call site.
- `scripts/ci/invoke_copilot_cli.py`: floor the model to `DEFAULT_COPILOT_MODEL`, and report a substitution right after the CLI streams are captured. `is_infrastructure_failure` already matches `not available`, but only when the output has no `VERDICT:` line, and a fallback run does have one, so the substitution passed silently.
- `scripts/ci/diagnose_copilot_cli.py`: same floor, plus the repo-root `sys.path` bootstrap the invoke driver already had so the import resolves when the action runs it as a script.
- `tests/ci/test_ai_review_model_economy.py` (new): served-model rate table, assertions that the default is the cheapest served model and cheaper than the cheapest Anthropic one, every workflow override, and a guard tying the action default to the driver constant.
- `tests/ci/test_copilot_model.py` (new): model-floor and substitution-detection unit coverage.
- `tests/ci/test_invoke_copilot_cli.py` and `tests/ci/test_diagnose_copilot_cli.py`: integration coverage that neither driver can emit an empty `--model`.

## Acceptance criteria

- [x] The action default is the cheapest model Copilot CLI currently serves.
- [x] No workflow override names an unserved model.
- [x] An unset or empty `COPILOT_MODEL` cannot reach the CLI as an empty `--model` value.
- [x] A model substitution at runtime surfaces as a workflow warning instead of passing silently.
- [x] A test fails when the default, the driver floor, or any override regresses.

## Type of Change

## Reviewer Expectations

**Review kind / scrutiny / urgency**: targeted, standard scrutiny. Cost-driven, so not urgent to merge but wasteful to sit on.

**Focus areas**:

- Review depth at the cheap tier. The format contract is enforced (missing verdict blocks), but a shallow well-formed `PASS` is not detectable by tooling. If review quality visibly drops, `gemini-3.6-flash` at $0.75 / $3.75 or `claude-haiku-4.5` at $1.00 / $5.00 are the next rungs, both still far under `auto`.
- Is the cheap tier enough for the `pr-comment-responder` step in `pr-maintenance.yml`, which runs with `execute-script` and a 10-minute timeout? It is the most agentic of the four call sites and the most likely to want a per-caller override.
- The served-model rate table in the new economy test is hand-maintained from GitHub's docs and will drift when Copilot's lineup changes. It fails closed (an unrecognized id fails the test), so drift blocks rather than passes silently, but it is still a table someone has to update.
- Trade-off taken: a versioned pin is the class ADR-080 warns about, and the action's old description named the `claude-opus-4.5` retirement that broke required checks as the reason for `auto`. The new tests plus the runtime warning are the compensating controls.

## Testing

Deliverables in this PR:

- [x] Tests added/updated

Commands run:

- `uv run pytest tests/ci/ -q`: 3067 passed, 14 skipped, exit 0.
- `uv run python scripts/validation/pre_pr.py`: 158 PASS, 0 FAIL, exit 0.
- `uv run ruff check` and `ruff format --check` on the changed files: clean.
- Negative controls, each run and observed to fail before restoring: reverting the default to `claude-haiku-4.5` (served, 5x pricier) fails 3 assertions; reverting to `auto` fails 6; restoring the `claude-sonnet-4.5` override fails the override assertion; diverging the driver constant from the action default fails 2; restoring the empty-string fallback fails 3 assertions in the invoke driver and 1 in the diagnostics driver.

Not verified locally: `validate_workflow_yaml` and `validate_yaml_style` reported `BLOCKED reason=tool.absent` in the pre-PR run because actionlint and yamllint are not installed in that container. Both workflow files parse under `yaml.safe_load`, but the linters themselves did not run there; the CI YAML Lint job is the check for those, and it passed.

### Security Review

The diff touches CI configuration, so noting the surface explicitly: no permissions, tokens, or secret handling changed. `report_model_fallback` prints only the model names matched out of already-redacted streams (`redact_secrets` runs before it).

- `.github/actions/ai-review/action.yml`
- `.github/workflows/artifact-insight-scanner.yml`

## Author Pre-flight

- [x] Builds locally (or N/A)
- [x] Self-review completed
- [x] Documentation updated (the action input description carries the rationale)

Two things noted on the path taken rather than hidden:

- The taste-lint 500-line threshold tripped twice during this change, first when the fallback helper was inlined into the invoke driver and again when the floor tests were appended to its test file. Both were resolved by extracting to the new module and test file rather than by raising the baseline.
- The AI reviewers on this repository are currently out of budget, so spec validation returns an infrastructure `CRITICAL_FAIL` that does not block merge. That predates this branch (see the comment thread below and PR #5801) and is the condition this change is meant to slow down. A separate follow-up is worth filing: `invoke_copilot_cli.py` retries an infrastructure failure three times, which multiplies a quota failure by three instead of short-circuiting it the way `is_permanent_auth_failure` short-circuits a rejected token.

## Related Issues

Refs #5804

---

https://claude.ai/code/session_017qiq2xAF1TWfwGxd8ymvXt


rjmurillo:
## Spec validation's `CRITICAL_FAIL` on this PR is pre-existing, and it is the problem this PR is about

The `AI-SPEC-VALIDATION` comment above reports `CRITICAL_FAIL` on both signals:

```
MESSAGE: Copilot CLI infrastructure failure after 3 attempts (exit code 1).
Check COPILOT_GITHUB_TOKEN scope, rate limits, or network connectivity.
```

This is not caused by the model change in this PR. PR #5801 merged before this branch existed, ran on the old `auto` default, and produced the byte-identical message ([comment](https://github.com/rjmurillo/ai-agents/pull/5801#issuecomment-5691781739), run 35053757468). Two other reviewers on that same PR say why:

- `chatgpt-codex-connector[bot]`: "Codex usage limits have been reached for code reviews. ... Credits must be used."
- `cursor[bot]`: "Bugbot couldn't run - usage limit reached."

So the AI reviewers are out of budget, and Copilot's exit 1 is the same wall. The job does not block merge (`Spec validation skipped due to infrastructure failure. Not blocking merge.`), which is why this has gone unnoticed: **the repository has been paying for AI review calls and receiving no review.**

Two consequences worth recording, neither fixed here:

1. **This strengthens the change rather than undermining it.** The default moves to `gpt-5.6-luna` at $0.20 in / $1.20 out per 1M tokens, against `auto`'s free routing up to `gpt-5.4` at $2.50 / $15.00. On the two high-volume paths (`ai-spec-validation` per PR synchronize, `pr-maintenance` hourly across up to 20 PRs) that is roughly a 12x reduction in the rate being charged.

2. **The retry loop multiplies a quota failure.** `invoke_copilot_cli.py` retries an infrastructure failure up to 3 times. When the cause is exhausted quota rather than a transient fault, every review spends three calls to learn the same thing. Worth a follow-up: treat a quota or spend-limit response as permanent, the way `is_permanent_auth_failure` already treats a rejected token, instead of retrying it. Filed under #5804 rather than widened into this PR.

No action taken on this PR for the failure itself: it is not this PR's, it does not block merge, and the fix is a budget decision rather than a code change.

---

PR #5812, opened 2026-09-16
fix(adr): one frontmatter contract for the three parsers that disagreed

rjmurillo:
# Pull Request

### What

Fixes the defect issue #5275 reports, and carries the buy/build/extend
evaluation that chose the approach. The three ADR frontmatter parsers now reach
the same verdict on every fence shape measured.

### Why

A closing fence padded with one trailing space parsed cleanly in the lifecycle
gate, ran all seven of its checks against real values, and crashed the index
build on the same file. A corpus could pass governance and break CI, on a change
no author would notice.

## Specification References

| Type | Reference | Description |
|------|-----------|-------------|
| **Issue** | Fixes #5275 | Three ADR frontmatter parsers disagree on closing-fence strictness |

`Fixes` is supported by the diff: acceptance criterion 1 by the shared contract
and the plugin-side mirror, criterion 2 by the new agreement test, criterion 3
by leaving the shared helper's other callers untouched (see Scope below).

## Changes

- `scripts/validation/frontmatter_contract.py`: the shared contract, delegating
  fence detection and YAML loading to `python-frontmatter`. No regex, no fence
  arithmetic.
- `build/scripts/generate_adr_index.py`: migrated onto it; local boundary regex
  and duplicate-key loader removed; the shipped README recipe rebuilt on the
  library; a wrong ADR-073 citation corrected.
- `scripts/validation/check_adr_lifecycle.py`: migrated onto it.
- `.claude/skills/adr-review/scripts/detect_adr_changes.py` and its two plugin
  mirrors: boundary mirrored from the library, because this tree cannot import it.
- `tests/validation/test_frontmatter_contract.py`: 39 cases, 100% coverage.
- `tests/validation/test_frontmatter_parser_agreement.py`: the cross-parser gate.
- `tests/skills/adr-review/test_detect_adr_changes_frontmatter_parity.py`: pins
  the mirror to the library shape by shape.
- `.agents/analysis/frontmatter-parser-build-vs-buy.md` and its three probes:
  the evaluation and its reproducible measurements.

## Result

Measured with the committed probe, before and after:

| | Before | After |
|---|---|---|
| Fence shapes where the three parsers disagree | **6 of 10** | **0 of 10** |
| `--- ` one trailing space | lifecycle passed, index **crashed** | all three accept |
| `----` four dashes | plugin rejected, other two accepted | all three accept |
| `--- nope` trailing text | lifecycle **accepted** | all three reject |
| ADR index | 109 records | 109 records, byte-identical |
[...]

Issue #5275, opened 2026-08-22
Three ADR frontmatter parsers disagree on closing-fence strictness


rjmurillo:
## What

Three independent parsers extract an ADR's YAML frontmatter block, and each accepts a different set of closing-fence shapes:

| Parser | File | Closing-fence rule | Strictness |
|---|---|---|---|
| `_parse_yaml_frontmatter` | `scripts/validation/yaml_utils.py:26` | `text.find("\n---", 3)` — a substring search, matches any line starting with `\n---` regardless of what follows | Loosest: accepts `--- anything` |
| `_split_frontmatter` | `scripts/validation/check_adr_lifecycle.py` (`_split_frontmatter`) | Deliberately mirrors `yaml_utils.py` verbatim (its own docstring quotes the identical `text.find("\n---", 3)` line and cites it as canonical) | Same as above |
| `_split_frontmatter` | `.claude/skills/adr-review/scripts/detect_adr_changes.py:225-239` | `line.strip() == "---"` — a line-based comparison, accepts trailing whitespace but rejects trailing text | Medium: accepts `--- ` (padded), rejects `--- anything` |
| `_FRONTMATTER_RE` | `build/scripts/generate_adr_index.py:78` | `^---\r?\n(...)\r?\n---\r?\n(...)$` — requires the closing `---` to be followed immediately by `\r?\n`, nothing else | Strictest: rejects even `--- ` (padded) |

## Evidence (verified by direct execution)

```python
from scripts.validation.yaml_utils import _parse_yaml_frontmatter
import re

text = "---\nid: ADR-999\nstatus: accepted\n--- \nBody text here.\n"  # one trailing space on the closing fence

_parse_yaml_frontmatter(text)
# => {'id': 'ADR-999', 'status': 'accepted'}   -- PARSES SUCCESSFULLY

_FRONTMATTER_RE = re.compile(r"^---\r?\n([\s\S]*?)\r?\n---\r?\n([\s\S]*)$")
_FRONTMATTER_RE.match(text)
# => None   -- generate_adr_index.py's parse_frontmatter() then hits:
#      raise AdrIndexError(f"{path.name} opens with '---' but has no closing "
#                           "'---' fence; the frontmatter block is unterminated")
```

A single trailing space on an ADR's closing frontmatter line:
- **Passes** `check_adr_lifecycle.py`'s gate cleanly (frontmatter parses, all seven lifecycle checks run against real values).
[...]

PR #5813, opened 2026-09-16
feat(metrics): measure real lefthook gate latency for epic #5456 release gate 3

rjmurillo:
# Pull Request

### What

Adds `scripts/metrics/gate_latency.py`, a measurement-only sampler that runs a real lefthook hook N times, parses lefthook's own per-job summary, and emits per-job and whole-hook latency as JSON and markdown. Plus REQ-027, an execution plan, and the committed measurements.

### Why

Epic #5456 release gate 3 asks whether local-gate p95 regressed. The figure it reads today is the declared sum of `lefthook.yml` `timeout:` fields (pre-push 3,330s), not a measurement, because this repository has no sampler. The disposition ledger records gate 3 as `Unchecked` for exactly that reason, and it is the only unchecked gate on that epic whose blocker is absent tooling rather than an open issue or paid authentication.

Issue #5318 items 1 and 2 already own this measurement ask, so this amends an existing owner rather than creating a new one, per the epic's Execution section.

## Specification References

| Type | Reference | Description |
|------|-----------|-------------|
| **Issue** | Refs #5456 | epic(v0.7.0): subtract the control plane and prove the smaller system |
| **Issue** | Refs #5318 | perf(hooks): the pre-push tail is unmeasured |
| **Spec** | `.agents/specs/requirements/REQ-027-gate-latency-sampler.md` | Requirement, AC-01 to AC-16 |
| **Spec** | `.agents/plans/active/5456-gate-3-latency-measurement.md` | Execution plan |

These references do not close their linked items: both issues carry work beyond this change.

## Changes

- 35 files, +4,202 / -2.
- Nine source modules under `scripts/metrics/`: `gate_latency.py` (CLI surface), `gate_latency_sampler.py` (measurement), `gate_latency_inputs.py` (validation), plus `lefthook_summary.py`, `gate_latency_stats.py`, `gate_latency_probe.py`, `gate_latency_models.py`, `gate_latency_classes.py`, `gate_latency_io.py`.
- Thirteen test modules under `tests/metrics/`, **214 tests passing**, split so each mirrors one source module or one concern.
- `.agents/metrics/gate-latency-v0.7.0/`: six committed measurement records, plus the summary at `.agents/metrics/gate-latency-v0.7.0.md`.
- Gate 3 and gate 4 evidence in `.agents/metrics/control-plane-dispositions-v0.7.0.md`.

## The headline measurement

| Hook | Declared budget | Worst observed | Gap |
|---|---|---|---|
[...]

PR #5814, opened 2026-09-16
refactor(validation): move the shared frontmatter helper and its mirror onto the contract

rjmurillo:
# Pull Request

### What

Moves `yaml_utils._parse_yaml_frontmatter` and the hand-rolled copy in
`spec_contradiction.py` onto the shared frontmatter contract that #5812 added.
Two parsers retired; four call sites inherit the change without being touched.

### Why

#5812 fixed the four parsers issue #5275 named. `yaml_utils` is the one the other
three had **copied their boundary arithmetic from**, and its module docstring
invites reuse by "other validators", so leaving it hand-rolled left the source of
the divergence in place for the next validator to clone.

## Specification References

| **Issue** | Refs #5275 | Three ADR frontmatter parsers disagree on closing-fence strictness |

These references do not close their linked items: #5275 was closed by #5812. This
is the follow-on the evaluation's rollout step 5 called for.

## Changes

- `scripts/validation/yaml_utils.py`: delegates to the contract, keeps its
  `dict | None` signature, no longer imports `yaml` or does fence arithmetic.
- `scripts/validation/spec_contradiction.py`: takes the shared fence and nothing
  else; its string-based value handling is unchanged.
- `tests/validation_pre_pr/test_frontmatter.py`: 10 cases pinning what changed
  and what must not.
- `tests/validation/test_spec_contradiction.py`: 5 cases, same split.

## Evidence for acceptance criterion 3

#5275's third criterion asks for no behaviour change to `yaml_utils`'s other
callers without their own review. That is measured here, not assumed:

**Across 4,074 markdown files** under `.agents`, `.claude`, `src`, `templates`,
`docs`, `.github`, `scripts`, `build` and `tests`, the old and new
implementations return **identical results on every file**.

Duplicate keys stay last-one-wins rather than adopting the contract's rejection.
The ADR gates asked for that rule; `validate_design_review.py` and
`validate_copilot_agent_frontmatter.py` did not.

## One deliberate change, one caught regression

**Deliberate:** `--- trailing text` no longer closes a block, the same tightening
#5812 applied elsewhere. A forged close could otherwise hide what followed it
from every caller of this helper.

**Caught before commit:** the first version returned the contract's `{}` for an
*empty* frontmatter block, where `yaml.safe_load("")` returns `None` and the old
`isinstance(result, dict)` test rejected it. That flipped
`check_adr_lifecycle`'s reason for such a record from "frontmatter block is
empty" to a missing-field complaint.

Worth noting how it surfaced: **no file in the corpus has an empty frontmatter
block**, so the 4,074-file scan could not see it. Only
`test_empty_frontmatter_block_names_itself` in the existing suite caught it. Only
`VALID` now returns a mapping, and a new test pins the case.

## Why `spec_contradiction` shares only the fence

Its value handling is deliberately string-based: it feeds a string comparison,
and the contract loads YAML, which would coerce `true` to a bool and `3` to an
int and silently change what counts as a contradiction. A test asserts every
value stays a `str`.

Its docstring also claimed parity with `yaml_utils` on being "top-level keys
only, no external YAML dependency". Both halves were false, since that helper
calls `yaml.safe_load`. Corrected rather than left, per
`canonical-source-mirror.md`: a wrong citation is worse than none.

## Acceptance criteria

- [x] Both parsers delegate their fence to the shared contract
- [x] `yaml_utils` keeps its `dict | None` signature so no call site changes
- [x] Behaviour parity measured across the corpus, not assumed
- [x] New tests fail against the pre-migration helper, so they are not vacuous
- [x] Duplicate-key and string-value semantics preserved where callers depend on them

## Type of Change

- [ ] Bug fix (non-breaking change fixing an issue)
- [ ] Documentation update
- [x] Refactoring (no functional changes)

## Reviewer Expectations

**Review kind / scrutiny / urgency**: full review, standard scrutiny, no rush.

**Focus areas**: the empty-block case is the one worth checking, since it is
where the first attempt was wrong and the corpus could not tell me. Second is
whether keeping duplicate keys last-one-wins here is the right call, or whether
those two validators should adopt the rejection in their own change.

## Testing

Deliverables in this PR:

- [x] Tests added/updated

**4,985 passed, 12 skipped** across `tests/validation` and
`tests/validation_pre_pr`. The lifecycle gate reports 0 violations across 109 ADR
records and the ADR index still matches.

### Security Review

### Other Agent Reviews

- [ ] Critic validated implementation plan
- [ ] QA verified test coverage

## Author Pre-flight

- [x] Builds locally (or N/A)
- [x] Pre-PR validation passed (full pre-push sequence)

Ratchets unchanged: taste 558, ruff 0, type-ignore 35, subprocess encoding 232,
memory index 377 of 378. `spec_contradiction.py` crossed the 500-line taste cap
during the work and was brought back to exactly 500 by condensing its own new
docstring and import, not by adding a suppression.

## Related Issues

Refs #5275

---

https://claude.ai/code/session_0192sUBNe6xkjFWqk5AKp3d1

---
_Generated by [Claude Code](https://claude.ai/code/session_0192sUBNe6xkjFWqk5AKp3d1)_


rjmurillo:
## The spec-validation FAIL above is infrastructure, not a finding on this diff

Same condition already documented on #5812 ([comment](https://github.com/rjmurillo/ai-agents/pull/5812#issuecomment-5694061885)); recording it here so a reviewer of this PR is not misled by the red verdict.

Both rows read `Copilot CLI infrastructure failure after 3 attempts (exit code 1)`. That is the analysis step failing to run, not a verdict that the implementation diverges from a spec. It is repo-wide: #5808, #5812 and #5813 hit the identical failure on unrelated diffs.

Verified on this PR's head `ebc515dc9`:

- The `Spec-to-Implementation Validation` workflow run itself concluded **`success`** ([run 35101551302](https://github.com/rjmurillo/ai-agents/actions/runs/35101551302)). The Copilot steps carry `continue-on-error: true`, so the comment's `FAIL` does not gate the PR.
- `Validate Spec Coverage` is green.
- **0 failed check runs.**

No fix ported, because none exists to port: `COPILOT_GITHUB_TOKEN` is a repository secret and its scope, rate limit, and reachability are maintainer-side. Nothing in this PR's scope clears it, and widening the PR to chase it would be wrong. Four PRs are now affected, which is worth its own issue.

---

